PR #1, opened 2019-09-18
[ADD] Flux: First draft basic flux (signal - intent - action)

Zinston:
Implements a POC of a basic flux using the signal -> intent -> action
flow as follows:

1. User triggers an `Event`.
2. `EventNormalizer` takes the event and formats a `Signal`.
3. `EventManager` takes the `Signal`, induces the user's intention and
   triggers the corresponding `'intent' Action`, through the
   `Dispatcher`.
4. `VDocument` and the `JWPlugins` that declared `ActionHandlers` to the
   `Dispatcher` for this `'intent' Action` generate `Action` sequences
   (`'composite' Actions` composed of `'primitive' Actions`) to follow
   up on that `'intent' Action`, through their respective
   `ActionHandlers` (callbacks).
5. `Dispatcher` does the same again with the generated `Actions` until
   the work is done.

A `JWPlugin` declares `ActionHandlers` in its constructor by defining an
`intents` object (to react on `'intent' Actions`), an `actions` object
(to react on `'primitive'` and `'composite' Actions` -- to be redefined)
and a `commands` object.
`intents` and `actions` take as keys the names
of the `Actions` they react to and as values a string mapping to a key
in their `commands` object (in the future we want these commands to be
callable cross-plugins).
`commands` takes as keys the names of the commands (to be used by
`intents` and `actions`) and as values the `ActionHandlers` (methods
that take an `Action` and return an `Action`).

`DevTools`' `ActionComponent` reacts to everything so it can be logged.

Since `owl` introduced a breaking change the meantime, I also had to make a fix for that. It's contained in a separate commit.
Noticing some extra (minor) issues in `DevTools`, I fixed those too, in another commit.
Finally I integrated the brand new `Dispatcher.registry` into `DevTools` for debugging (in another commit).


src/core/types/Flux.ts
```diff
@@ -0,0 +1,34 @@
...
+    origin: string;
+}
+
+export type ActionPayload = {};
+
+export interface Action {
+    name?: string;
+    type: 'intent' | 'primitive' | 'composite';
+    subActions?: Action[];
+    value?: ActionPayload;
```
Zinston:
```(javascript)
payload?: ActionPayload;
```

src/core/utils/CorePlugin.ts
```diff
@@ -0,0 +1,30 @@
+import { JWPlugin } from '../JWPlugin';
+import { VDocument } from '../stores/VDocument';
+import { VNode } from '../stores/VNode';
+import { Action } from '../types/Flux';
+
+export class CorePlugin extends JWPlugin {
+    vDocument: VDocument;
```
dmo-odoo: We had a discussion yesterday as to where to put the document, and I think no plugin should have direct access to the document, which is our actual state. We need to discuss about when we start creating the commitState step.

src/core/utils/CorePlugin.ts
```diff
@@ -0,0 +1,30 @@
...
+    };
+    commands = {
+        onRemoveIntent: this.removeSide,
+    };
+    actions = {
+    };
+    constructor(dispatcher, vDocument) {
+        super(dispatcher);
+        this.vDocument = vDocument;
+        this._root = this.vDocument._root;
```
dmo-odoo: Why ?

src/core/utils/EventManager.ts
```diff
@@ -1,7 +1,8 @@
 import { EventNormalizer } from './EventNormalizer';
+import { Signal, Action } from '../types/Flux';
 
 export interface EventManagerOptions {
-    dispatch?: Function;
+    dispatch?: (action: Action) => void;
```
dmo-odoo: Maybe the Dispatcher should ```export type DispatchFunction = (action: Action) => void``` to make it explicit this cannot be any function but *has* to be the one actually dispatching ?

src/core/utils/EventManager.ts
```diff
@@ -14,15 +15,37 @@ export class EventManager {
...
+
+    /**
+     * Induce the user's intention from the received signal, based on the user's
+     * configuration and context.
+     * TODO: this is just a stub
+     *
+     * @param {Signal} signal
+     * @returns {Action}
+     */
+    _induceIntent(signal: Signal): Action {
```
dmo-odoo: I'm not a fan of the word "induce", but I'm at a loss here, honestly.

src/core/utils/EventManager.ts
```diff
@@ -14,15 +15,37 @@ export class EventManager {
...
+        switch (signal.type) {
+            case 'remove':
+            // return a 'remove' action of type 'intent', with the right payload
```
dmo-odoo: You meant ```TODO:``` right ?

src/core/types/Flux.ts
```diff
@@ -0,0 +1,32 @@
+export interface Signal {
+    type: string; // ex.: 'keydown'
+    params: object; // ex.: {ctrlKey: true}
+    elements?: Set<HTMLElement>; // (?) see EventNormalizer
```
dmo-odoo: TODO: fix comment

src/core/types/Flux.ts
```diff
@@ -0,0 +1,32 @@
...
+    origin: string;
```
dmo-odoo: Is this relevant ?

src/core/types/Flux.ts
```diff
@@ -0,0 +1,32 @@
+export interface Signal {
```
dmo-odoo: I'm not a fan of the word ```Signal``` even though we used it a lot when discussing this signal/intent stuff. I think ```Event``` would be way easier to grasp for newcomers to the project, especially since they are generated by the ```EventManager```, but it's an already existing type.
Basically, we have ```DOMEvent ~> EventNormalizer => NormalizedEvent ~> EventManager => Intent ~> Dispatcher```. I don't like ```NormalizedEvent``` either since the fact that they are normalized is just an implementation detail. Couldn't the ```EventNormalizer``` use the same api as the DOMEvent ?

src/core/utils/EventNormalizer.ts
```diff
@@ -1,3 +1,5 @@
+import { Signal } from '../types/Flux';
```
dmo-odoo: I'm sorry but I really dislike the term ```Flux```. At Odoo we have always used ```Flow``` instead. Not saying you are wrong, just that I'm not used to it. That being said, I wouldn't really like calling the file ```Flow``` either. Personally, I would rename it ```Actions``` and make it a `d.ts` file so that it is automatically imported everywhere in the project.

src/core/utils/EventNormalizer.ts
```diff
@@ -61,12 +63,12 @@ export class EventNormalizer {
     _observer: MutationObserver;
     _selectAllOriginElement: DOMElement; // original selection/target before updating selection
     _mousedownInEditable: MouseEvent; // original mousedown event when starting selection in editable zone
-    _triggerEvent: Function; // callback to trigger on events
+    _sendSignal: (signal: Signal) => void; // callback to trigger on events
```
dmo-odoo: This renaming is funny considering I don't like the term `Signal` and want to think about whether the Normalizer could trigger events in a DOMEvent-compatible format... ;-)

src/core/utils/EventNormalizer.ts
```diff
@@ -294,30 +296,73 @@ export class EventNormalizer {
...
+                        {
+                            value: '<br/>',
+                        },
+                        elements,
+                    ),
+                );
             } else {
-                this._triggerEvent('insert', data, elements);
+                this._sendSignal(
+                    this._formatSignal(
```
dmo-odoo: If you don't name the dispatcher function `_sendSignal`, you can implement `_sendSignal` such that it formats then send the signal, thus saving you one function call everytime... ;-)

src/core/utils/EventNormalizer.ts
```diff
@@ -636,15 +681,15 @@ export class EventNormalizer {
...
-    _isAtVisibleEdge(element: HTMLElement, side: string): boolean {
+    _isAtVisibleEdge(element: HTMLElement, side: 'start' | 'end'): boolean {
         if (!this.editable.contains(element)) return false;
         // Start from the top and do a depth-first search trying to find a
         // visible node that would be in editable and beyond the given element.
         let node = this.editable;
-        const child = side === 'left' ? 'firstChild' : 'lastChild';
-        const sibling = side === 'left' ? 'nextSibling' : 'previousSibling';
+        const child = side === 'start' ? 'firstChild' : 'lastChild';
+        const sibling = side === 'start' ? 'nextSibling' : 'previousSibling';
```
dmo-odoo: 👀

src/plugins/Awesome.ts
```diff
@@ -8,15 +8,18 @@ interface AwesomeConfig extends JWPluginConfig {
...
     init(): Action {
         super.init();
         return {
-            type: ActionType.INSERT,
-            value: 42,
+            type: 'primitive',
+            name: 'insert',
+            payload: {
+                value: 42,
+            },
```
dmo-odoo: Maybe we should remove this function (and/or this entire plugin) ? It makes no sense anyway. That being said, this PR would probably not be the right place to do so.

src/plugins/DevTools/DevTools.ts
```diff
@@ -3,9 +3,6 @@ import { JWOwlUIPlugin } from '../../ui/JWOwlUIPlugin';
 import devtoolsTemplates from './DevTools.xml';
 
 export class DevTools extends JWOwlUIPlugin {
-    constructor() {
-        super();
-        this.componentsRegistry.push(DevToolsComponent);
-        this.templates = devtoolsTemplates;
-    }
+    componentsRegistry = this.componentsRegistry.concat(DevToolsComponent);
+    templates = devtoolsTemplates;
```
dmo-odoo: This should not be in this commit. It won't not truly fit in any other commit however, since its some basic refactoring, so I suggest you to put it in the commit where you add the registry inspector in DevTools, at least the commit will target DevTools specifically so it will make more sense IMO.

src/ui/OwlUI.ts
```diff
@@ -47,28 +53,56 @@ export class OwlUI {
...
+     * Mount all of a plugin's Components to Owl and register its components'
+     * intents, actions and commands (TODO: this does not work for
+     * sub-components since those are mounted by their parent and we do not have
+     * access to their instantiation).
      *
      * @param {JWOwlUIPlugin} pluginInstance
      * @returns {Promise<void>}
      */
-    async _mountPluginComponents(pluginInstance: JWOwlUIPlugin): Promise<void> {
+    async _handleComponents(pluginInstance: JWOwlUIPlugin): Promise<void> {
```
dmo-odoo: Same remark about instances being the default case. I'd rename the parameter to `plugin`.

src/ui/OwlUI.ts
```diff
@@ -47,28 +53,56 @@ export class OwlUI {
...
-        const components = pluginInstance.componentsRegistry.slice();
-        components.forEach((_Component: typeof Component) => {
-            this._mount(_Component, env);
-        });
+        const Components = pluginInstance.componentsRegistry.slice();
+        for (let i = 0; i < Components.length; i++) {
+            const component = new Components[i](env);
+            await this._mount(component);
+            this._register(component, pluginInstance);
+        }
```
dmo-odoo: Why do you need the slice ? Why not the following:
```
plugin.componentsRegistry.forEach(Component => {
    const component = new Component(env);
    await this._mount(component);
    this._register(component, plugin);
});
```

src/ui/OwlUI.ts
```diff
@@ -47,28 +53,56 @@ export class OwlUI {
...
+                            ' "' +
+                            name +
+                            '". TODO: resolve conflict (chain commands if they exist).',
+                    );
+                } else {
+                    pluginInstance[type][name] = component[type][name];
+                }
+            });
+        });
+    }
```
dmo-odoo: For the moment, I would move this inside the DevTools plugin. I think it is the very special case of the DevTools which wants to be a UI plugin AND respond to intents/actions/etc in the same time. I would not offer this possibility to other UI plugins in the core system (yet.)

src/core/types/Flux.ts
```diff
@@ -0,0 +1,32 @@
...
+    [key: string]: string;
+}
+
+export interface PluginActions {
+    [key: string]: string;
+}
+
+export interface PluginCommands {
+    [key: string]: ActionHandler;
+}
```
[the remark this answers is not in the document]
Zinston: 👍

src/core/utils/EventManager.ts
Zinston: Yes i do

src/core/utils/EventManager.ts
Zinston: `toIntent` ? `signalToIntent` ? `matchIntent` ?

src/plugins/Awesome.ts
Zinston: I'd remove the plugin.

src/ui/OwlUI.ts
Zinston: I don't anymore. Also there is no `await` in the loop so I can put a `forEach` again.

src/ui/OwlUI.ts
Zinston: I guess you're right.

src/core/dispatcher/Dispatcher.ts
```diff
@@ -45,4 +106,72 @@ export class Dispatcher {
...
+     * @param {'intent'|'action'} type
+     * @param {string} name
+     * @returns {DispatcherRegistryRecord}
+     */
+    _getRecord(type: 'intent' | 'action', name: string): DispatcherRegistryRecord {
+        const record = this.registry.find((record: DispatcherRegistryRecord): boolean => {
+            return record.type === type && record.name === name;
+        });
+        return record;
+    }
```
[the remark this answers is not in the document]
Zinston: The solution with `export type DispatcherRegistry = Record<ActionIdentifier, Record<HandlerToken, ActionHandler>>;` does not entirely solve that problem: we still need to iterate over the records to look at each subrecord (`Record<HandlerToken, ActionHandler>`).

src/core/types/Flux.ts
Zinston: Removed Signal ;-)

src/core/utils/EventNormalizer.ts
Zinston: :-D

src/ui/OwlUI.ts
Zinston: Damn I think I can't. The component has to be mounted and that is done in OwlUI.

Zinston: OK I got it to work. Only trade-off: now a JWOwlUIPlugin has to manually call an init() method in its constructor. That is in order to mount its own specific components. I can't automize that in the parent class because that would imply initializing the Components list before calling super(), which is not supported by typescript. But I feel that's a very minor trade-off, and that's allowed me to bypass it in DevTools so I could register the handlers _before_ mounting the components.

Zinston:
@dmo-odoo Applied the comments. I managed to take the registering of handlers and commands out of OwlUI and into DevTools. Only trade-off is every ui plugin now has to call an init method manually to mount its components.

src/core/utils/EventNormalizer.ts
```diff
@@ -294,30 +294,83 @@ export class EventNormalizer {
...
+                        {
+                            value: '<br/>',
+                        },
+                        elements,
+                    ),
+                );
             } else {
-                this._triggerEvent('insert', data, elements);
+                this._triggerEvent(
+                    this._formatCustomEvent(
```
dmo-odoo: Why these ones were not changed to formatAndTrigger ?

src/core/utils/EventNormalizer.ts
```diff
@@ -294,30 +294,83 @@ export class EventNormalizer {
...
+    }
+    /**
+     * Format a custom event and trigger it.
+     *
+     * @param {string} type
+     * @param {object} [params]
+     * @param {Set<HTMLElement>} [elements]
+     */
+    _formatAndTriggerEvent(type: string, params = {}, elements?: Set<HTMLElement>): void {
+        this._triggerEvent(this._formatCustomEvent(type, params, elements));
```
dmo-odoo: At this stage, I think I would just rename `_triggerEvent` into `_eventCallback`, then rename `_formatAndTriggerEvent` into `_triggerEvent`, and finally move the code of `_formatCustomEvent` inside `_triggerEvent`. This would be a private function of the normalizer where it triggers a Custom Event on the callback, the fact that it has to format it first is an implementation detail IMO.

src/core/utils/EventNormalizer.ts
```diff
@@ -555,8 +608,10 @@ export class EventNormalizer {
             origin: 'composition',
         };
 
-        this._triggerEvent('setRange', insertionRange, []);
-        this._triggerEvent('insert', insertedText, elements);
+        this._triggerEvent(
+            this._formatCustomEvent('setRange', { value: insertionRange }, new Set()),
```
dmo-odoo: Search/Replace miss

src/core/utils/EventNormalizer.ts
```diff
@@ -902,10 +957,12 @@ export class EventNormalizer {
             (!this._compiledEvent || this._compiledEvent.key === 'a') &&
             this._isSelectAll(this._getRange())
         ) {
-            this._triggerEvent('selectAll', {
-                origin: this._compiledEvent ? 'keypress' : 'pointer',
-                target: this._selectAllOriginElement,
-            });
+            this._triggerEvent(
+                this._formatCustomEvent('selectAll', {
```
dmo-odoo: miss (Though if you rename the functions like I suggested, you won't be able to miss it anymore since _formatCustomEvent would not exist anymore ;-))

src/plugins/DevTools/DevTools.ts
```diff
@@ -4,9 +4,25 @@ import devtoolsTemplates from './DevTools.xml';
...
+
+    /**
+     * Mount and register all components.
+     */
+    _mountAndRegisterComponents(): void {
+        const components = this._instantiateComponents(this.Components);
+        components.forEach((component: OwlUIComponent<{}, {}>): void => {
+            this.env.editor.dispatcher.register(component.handlers, component.commands);
+        });
+        this._mountComponents(components);
```
dmo-odoo: I'm sorry I don't remember: Why can't all JWOwlUIPlugin behave like this ? In the sense of "Why can't this be the default behavior ?"

src/plugins/DevTools/components/DevToolsComponent.ts
```diff
@@ -17,6 +18,15 @@ export class DevToolsComponent extends OwlUIComponent<{}, DevToolsState> {
...
         height: 300,
+        actions: [], // Stack of all actions performed since init
+    };
+    handlers: PluginHandlers = {
+        intents: {
+            '*': 'addAction',
+        },
+    };
+    commands = {
+        addAction: this.addAction.bind(this),
```
dmo-odoo: I'm thinking about something. You remember when we talked about namespacing the commands ? What about something like `DevTools:addAction` ? Then you could hook anything in the same way. You want to hook an "intent" (or a "command" if we rename it), you hook `addAction`. You want to hook the specific addAction command of the DevTools ? You hook `DevTools:addAction`. You want to hook the plugin Whatever which is already hooking `DevTools:addAction` ? You hook `Whatever:addAction`. You want to hook the `wrap` primitive ? You hook `Document:wrap`. That's it. One mechanism. (Just an idea for later.)

src/core/utils/EventNormalizer.ts
```diff
@@ -918,8 +918,8 @@ export class EventNormalizer {
             const target = this._mousedownInEditable.target as Element;
             this._mousedownInEditable = null;
             if (ev.target instanceof Element) {
-                let range: Range;
-                if (target === ev.target) {
+                let range: Range = this._getRange();
+                if (!target.contains(range.startContainer) && target === ev.target) {
```
dmo-odoo: This particular part of the EventNormalizer is lacking documentation, but it's not your fault, it's simply something I notice now.

src/core/JWEditor.ts
```diff
@@ -53,6 +53,18 @@ export class JWEditor {
...
+        // CorePlugin is a special mandatory plugin. It's the plugin that will
+        // handle matching between the actions the DOM. We want it to be a
+        // plugin because it needs to behave in the same way as a plugin,
+        // listening to Actions and reacting on them _within the dispatching
+        // loop_, thereby giving a chance to other plugins to react as well.
+        // Given its specificity, it requires that we pass it the vDocument
+        // instance so we instantiate it here to pass that through the
+        // constructor.
+        const corePlugin = new CorePlugin(this.dispatcher, this.vDocument);
+        this._addPluginInstance(corePlugin);
```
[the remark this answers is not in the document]
Zinston: Yes. This is strictly temporary, so we can already have a flux. It should go away when Christophe introduces his system. I'll add a comment to clarify.

src/core/dispatcher/Dispatcher.ts
```diff
@@ -1,48 +1,147 @@
-import { Action } from '../actions/Action';
-
 export type HandlerToken = string;
-export type ActionHandler = (action: Action) => void;
+export type Handlers = Record<HandlerToken, ActionHandler>;
+export type DispatcherRegistry = Record<ActionIdentifier, Handlers>;
```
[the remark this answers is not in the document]
Zinston: Well... `DispatcherRegistry` is used later in `ActionsComponent`, and it contains `Handlers`. I could remove them and then use `Record<ActionIdentifier, Record<HandlerToken, ActionHandler>>` instead in `ActionsComponent` but that seems confusing to me.

src/core/dispatcher/Dispatcher.ts
```diff
@@ -1,48 +1,147 @@
...
+     * the given action type, unless `onlyIncludeSpecific` is true.
+     * '*' is a magic key used to listen to action of a given type.
+     *
+     * @param {ActionIdentifier} actionIdentifier
+     * @param {boolean} [onlyIncludeSpecific] if true, do not include '*' records
+     * @returns {Handlers}
+     */
+    _getHandlers(actionIdentifier: ActionIdentifier, onlyIncludeSpecific = false): Handlers {
+        const record = Object.assign({}, this.registry[actionIdentifier]) || {};
+        if (!onlyIncludeSpecific) {
```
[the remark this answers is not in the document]
Zinston: I figured we should have a way to retrieve only the handlers that are specific to the given action identifier, simply because it's not exactly intuitive that `_getHandlers` would actually include wildcard handlers (even though in most cases that is in fact what we want). You're absolutely right about the name though.

src/core/dispatcher/Dispatcher.ts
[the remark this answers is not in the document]
Zinston: Sounds good. That's actually the type `Handlers` so I simply added a `handlers` key next to `registry` and in `_register` I add the new handler to both the `registry` and the `handlers` maps. Now I can indeed simply do `this.handlers[handlerToken]` to retrieve a specific handler.

src/plugins/DevTools/DevTools.ts
Zinston: I would like it to be. But for that any extension of `JWOwlUIPlugin` should declare its `Components` _before_ calling `super` (so that `JWOwlUIPlugin` can mount the components). Typescript doesn't allow anything to happen before `super`.
Now the specific case of _registering_ the components in the dispatcher is very specific to `DevTools`: we should not make it the default that UI plugins listen to actions - there is no reason for them to do that. `DevTools` is kind of a weird beast on that level: it's a UI plugin that also behaves a bit like a regular plugin...

src/plugins/DevTools/components/DevToolsComponent.ts
Zinston: I like that idea very much. To explore!

src/plugins/DevTools/DevTools.xml
```diff
@@ -189,59 +191,170 @@
...
+        </table>
+    </t>
+
+    <!-- ACTIONS.Payload[key] -->
+    <t t-name="payloadValue">
+        <t t-if="value and value.nodeName">
+            &lt;<t t-esc="value.nodeName.toLowerCase()"/>&gt;
+        </t>
+        <t t-elif="value == true">true</t>
+        <t t-elif="!value and value !== false and value !== 0 and value !== null">undefined</t>
```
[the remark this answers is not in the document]
Zinston: `typeof` doesn't work in owl: it doesn't compile properly and I get the following error: `ReferenceError: typeof_75 is not defined`. I guess it believes I tried to use a variable named `typeof`...?

src/plugins/DevTools/components/ActionsComponent.ts
```diff
@@ -1,36 +1,106 @@
...
 import { OwlUIComponent } from '../../../ui/JWOwlUIPlugin';
+import { HandlerToken, DispatcherRegistryRecord } from '../../../core/dispatcher/Dispatcher';
 
 interface ActionsState {
-    selectedIndex: number;
+    actions: Action[]; // Stack of all actions performed since init
+    openTab: 'selected' | 'registry';
+    registry: DispatcherRegistryRecord[];
+    selectedActionIndex: number;
+    selectedHandlerToken: HandlerToken | null;
```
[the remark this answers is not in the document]
Zinston: I guess you're right. But isn't it more explicit to specify that it can be `null` though? So we know we should test for that? I think I'd rather add ` | null` next to `selectedActionIndex: number`.

src/plugins/DevTools/components/ActionsComponent.ts
```diff
@@ -1,36 +1,106 @@
...
+            : null;
+        this.selectRegistryRecord(correspondingRecordID);
+    }
+    /**
+     * Select the action handler corresponding to the given token.
+     *
+     * @param {HandlerToken|null} handlerToken
+     */
+    selectHandler(handlerToken: HandlerToken | null): void {
+        this.state.selectedHandlerToken = handlerToken;
```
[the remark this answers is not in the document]
Zinston: We need it so it can be called in the xml template. Now where I use `this.selectHandler(null)`, nothing keeps me from using `this.state.selectedHandlerToken = null` but personally I'd rather use the method in case later we decide that something else that just `this.state.selectedHandlerToken = null` needs to be done when selecting a handler. I'd rather not scatter the code.

src/core/dispatcher/Dispatcher.ts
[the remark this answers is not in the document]
dmo-odoo: Yeah you're right, I think I didn't realize one was referencing the other when I wrote my comment.

src/plugins/DevTools/DevTools.xml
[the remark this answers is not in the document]
dmo-odoo: Why should null be treated differently ?

src/plugins/DevTools/components/ActionsComponent.ts
```diff
@@ -1,36 +1,106 @@
...
 
 export class ActionsComponent extends OwlUIComponent {
     state: ActionsState = {
-        selectedIndex: 0, // Index of the selected action in the stack
+        actions: [], // Stack of all actions performed since init
+        openTab: 'selected',
+        registry: this.env.editor.dispatcher.registry,
+        selectedActionIndex: null, // Index of the selected action in the stack
+        selectedHandlerToken: null,
+        selectedRegistryRecordIndex: null, // Index of the selected registry record
```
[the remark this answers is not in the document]
Zinston: I thought the 'index' ones required a comment to clarify what they referred to, but that handler token spoke for itself as it's a type... But ok I'll add a comment ;-)

src/plugins/DevTools/components/ActionsComponent.ts
```diff
@@ -1,36 +1,106 @@
...
+    selectHandler(handlerToken: HandlerToken | null): void {
+        this.state.selectedHandlerToken = handlerToken;
+    }
+    /**
+     * Select the registry record at given index
+     *
+     * @param {number|null} index
+     */
+    selectRegistryRecord(index: number | null): void {
+        this.state.selectedRegistryRecordIndex = index;
```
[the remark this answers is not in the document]
Zinston: A registry record has for key an `ActionIdentifier`, and it contains a set of handlers (`Record<HandlerToken, ActionHandler>`). So... it's a registry record. On the other hand it's not an index anymore but an `ActionIdentifier` so the name does have to change!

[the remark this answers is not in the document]
Zinston: `selectedActionIdentifier`

src/plugins/DevTools/DevTools.ts
```diff
@@ -4,9 +4,25 @@ import devtoolsTemplates from './DevTools.xml';
...
+    //--------------------------------------------------------------------------
+    // Private
+    //--------------------------------------------------------------------------
+
+    /**
+     * Mount and register all components.
+     */
+    _mountAndRegisterComponents(): void {
+        const components = this._instantiateComponents(this.Components);
+        components.forEach((component: OwlUIComponent<{}, {}>): void => {
```
dmo-odoo: I don't think you have to specify types in forEach. I personally thinks it makes the code very hard to read. Is `components` an array of `OwlUIComponent` or an array of functions with `(component: OwlUIComponent): void` as signature ? I know the answer but just by reading the code ? Took me a while to figure what was going on.

src/plugins/DevTools/components/ActionsComponent.ts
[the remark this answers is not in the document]
dmo-odoo: If we choose to keep the method then yes, we indeed should continue to use it.

src/core/actions/ActionGenerator.ts
```diff
@@ -5,37 +5,97 @@ export class ActionGenerator {
...
+            origin: origin,
+        };
+        if (payload) {
+            action.payload = payload;
+        }
+        if (position) {
+            action.position = position;
+        }
+        return action;
+    }
```
dmo-odoo: When I see all these functions, I wonder why we don't simply use a `actionParams` parameter, which follow the interface of these function arguments, and smply add the missing `id` property via `make`. No more `if`s. You want a payload ? You set it. You didn't set a payload ? We won't set a payload. If the payload is an object, you'll have to make the call multiline anyway, and it will look quite strange with every parameter listed in the function call. Plus, you want to set `position` but not `payload` ? You're screwed, you'll have to pass a useless value like null or undefined for `payload`... :(

PR #2, opened 2019-09-18
[IMP] Use karma for tests

Goaman:
This commit introduce a test infrastructure based on karma.
"expample.test.ts" should be removed when adding the first proper test.


Zinston:
> "expample.test.ts" should be removed when adding the first proper test.

You're the "expample"!

package.json
```diff
@@ -24,23 +24,38 @@
   },
   "homepage": "https://github.com/odoo/jabberwock#readme",
   "devDependencies": {
-    "@types/jest": "^24.0.18",
+    "@types/chai": "^4.2.1",
+    "@types/dom-inputevent": "^1.0.5",
```
dmo-odoo: Interesting. I created an InputEvent interface in our project because it was missing from TypeScript. I guess it can be deleted if we now depend from inputevent ?

package.json
```diff
@@ -24,23 +24,38 @@
...
-    "jest": "^23.6.0",
+    "karma": "^4.3.0",
+    "karma-chai": "^0.1.0",
+    "karma-chrome-launcher": "^3.1.0",
+    "karma-firefox-launcher": "^1.2.0",
+    "karma-mocha": "^1.3.0",
+    "karma-phantomjs-launcher": "^1.0.4",
+    "karma-sinon": "^1.0.5",
+    "karma-webpack": "^4.0.2",
+    "lodash": "^4.17.15",
```
dmo-odoo: Why ?

package.json
Goaman: yep I think so

package.json
Goaman: I forgot to remove it, it should not be there

PR #5, opened 2019-09-23
[ADD] DevTools: log Handlers and VNodes in console

Zinston:


Zinston:
@dmo-odoo ping :-)

src/core/stores/VNode.ts
```diff
@@ -132,4 +144,51 @@ export class VNode {
...
+    }
+    /**
+     * Recalculate the indexes of this node's children.
+     */
+    _updateIndexes(): VNode {
+        this.children.forEach((child: VNode, i: number) => {
+            child.index = i;
+        });
+        return this;
+    }
```
dmo-odoo: remove

src/plugins/DevTools/DevTools.xml
```diff
@@ -97,6 +97,7 @@
             <t t-if="props.vNode.value">:
                 "<t t-esc="props.vNode.value"/>"
             </t>
+            <button class="logger" t-on-click="logVNode(props.vNode)" style="display: inline; padding-left: 1em;">&gt;_</button>
```
dmo-odoo: padding => margin

dmo-odoo: Why inline style ?

Zinston:
@dmo-odoo Done applying review.

src/plugins/DevTools/DevTools.css
```diff
@@ -55,6 +55,25 @@ devtools-panel .selectable-line.selected {
...
+    background-color: transparent;
+    color: #666666;
+    padding: 0;
+    margin: 0;
+    display: none;
+}
+
+devtools-panel .about > button.logger {
+    display: inline;
+    padding-left: 1em;
```
dmo-odoo: Didn't we discuss to use margin here rather than padding? Not a big deal though.

Zinston:
@dmo-odoo Rebased on master, ready.

src/plugins/DevTools/components/InfoComponent.ts
```diff
@@ -27,6 +27,15 @@ export class InfoComponent extends OwlUIComponent<{}> {
...
         return this._repr(this.state.range.start);
     }
+    /**
+     * Log a `VNode` to the console.
+     *
+     * @param {VNode} vNode
+     */
+    logVNode(vNode: VNode): void {
+        // Get from vDocument so we don't log a proxy
+        console.log(this.env.editor.vDocument.find(vNode.id));
```
dmo-odoo: I realize I never actually *wrote* a comment summarizing the discussion we had about proxification, thus resulting in this PR looking like it is abandonned.

In short, since what you truly want to do is "unproxify" Owl's proxy, I would remove everything this PR adds to `VDocument` then put the code of `_findByID` in OwlUI and call it something like `unproxifyVNode` or `deproxifyVNode` depending on which one is more correct in english.

Or better still ! Do we truly need owl to *observe* the VNodes ? I mean, we know when they change since it triggers a refresh, and I don't think we ever do `node.something = ...` in the DevTools, so owl would not see the changes triggered by the editor itself anyway, so do we really need to put the VDocument in the observed state ?

packages/plugin-devtools/assets/DevTools.xml
```diff
@@ -122,6 +122,7 @@
             <t t-if="props.vNode.char">:
                 "<t t-esc="props.vNode.char"/>"
             </t>
+            <button class="logger" t-on-click="logVNode(props.vNode)">&gt;_</button>
```
dmo-odoo: @aab-odoo Didn't you tell me that it was possible to use `console.log` in owl templates ?

dmo-odoo:
I'm marking this PR as blocked while waiting for aab's input.

packages/plugin-devtools/assets/DevTools.xml
aab-odoo: @dmo-odoo It's not. It will produce `this.console.log('blabla');`

PR #7, opened 2019-10-08
[POC] Renderer: render DOM and Range

Zinston:
This implements a `setRange` flux:
1. User clicks in DOM
2. `EventNormalizer` interprets the click and sends a `CustomEvent`
3. `EventManager` interprets the `CustomEvent` and triggers an `Intent` (`'setRange'`, generated by the `ActionGenerator`) via the `Dispatcher`
4. `CorePlugin` picks up the `Intent` and matches the DOM nodes to their respective `VNodes`, moves the `Range VNodes` accordingly, and triggers a 'render' `Intent` via the `Dispatcher`.
5. After giving a chance to other plugins to react on it, `CorePlugin` picks up the action and triggers a rendering in the `Renderer`.
6. `Renderer` rerenders the tree and sets the range according to the position of the `Range VNodes`

TODO
* update the store AFTER the whole flux: requires a mechanism to recognize the end of the flux
* handle range on `BR` nodes properly. Currently the `Renderer` sets the range logically where it belongs, but that is not how the DOM works: it prefers following a set of rules inspired by Kamoulox to handle the range on consecutive `BR` nodes.

### DIFF
--- a/src/core/utils/CorePlugin.ts
+++ b/src/core/utils/CorePlugin.ts
@@ -1,15 +1,25 @@
 import { JWPlugin } from '../JWPlugin';
 import { VDocument } from '../stores/VDocument';
+import { VNode } from '../stores/VNode';
+
+interface VPosition {
+    ref: VNode;
+    side: 'before' | 'after';
+}
 
 export class CorePlugin extends JWPlugin {
     vDocument: VDocument;
     handlers = {
         intents: {
             remove: 'onRemoveIntent', // names are just to show relationships here
+            render: 'render',
+            setRange: 'navigate',
         },
     };
     commands = {
+        navigate: this.navigate.bind(this),
         onRemoveIntent: this.removeSide,
+        render: this.render.bind(this),
     };
     constructor(dispatcher, vDocument) {
         super(dispatcher);
@@ -20,7 +30,96 @@ export class CorePlugin extends JWPlugin {
     // Public
     //--------------------------------------------------------------------------
 
-    removeSide(intent: Action): void {
+    removeSide(intent: Intent): void {
         console.log('REMOVE SIDE:' + intent);
+    }
+    /**
+     * Navigate to a given Range (in the payload of the Intent).
+     *
+     * @param intent
+     */
+    navigate(intent: Intent): void {
+        this._setRange(intent.payload['value']);
+    }
+    /**
+     * Render the `vDocument`.
+     */
+    render(): void {
+        this.vDocument.renderer.render(this.vDocument.editable);
+    }
+
+    //--------------------------------------------------------------------------
+    // Private
+    //--------------------------------------------------------------------------
+
+    /**
+     * Return a position in the `VDocument` as a tuple `[VNode, 'before' | 'after']`
+     * (ie. the position is before or after a given `VNode`). The position is
+     * always given on the leaf.
+     *
+     * @param container
+     * @param offset
+     */
+    _getPosition(container: DOMElement | Node, offset: number): VPosition {
+        const containers = this.vDocument.domMap.get(container);
+        const isTextNode = container.nodeType === 3;
+        if (isTextNode) {
+            return this._positionToLeaf({
+                ref: containers[offset] || containers[containers.length - 1],
+                side: offset < containers.length ? 'before' : 'after',
+            });
+        }
+        return this._positionToLeaf({ ref: containers[offset], side: 'before' });
+    }
+    /**
+     * Return a tuple with the range nodes in ltr order (rtl if `reverse` is
+     * true) and updates their `order` property accordingly.
+     *
+     * @param reverse
+     */
+    _orderedRangeNodes(reverse = false): [VNode, VNode] {
+        const rangeNodes = this.vDocument.rangeNodes;
+        const res: [VNode, VNode] = [rangeNodes.start, rangeNodes.end];
+        if (reverse) {
+            res.reverse();
+        }
+        res.forEach((rangeNode: VNode, i: 0 | 1) => {
+            rangeNode.order = i;
+        });
+        return res;
+    }
+    /**
+     * Move the position to its deepest first descendent.
+     *
+     * @param position
+     */
+    _positionToLeaf(position: VPosition): VPosition {
+        while (position.ref.children.length) {
+            position.ref = position.ref.firstChild;
+            position.side = 'before';
+        }
+        return position;
+    }
+    /**
+     * Move the `Range VNodes` to the position in `VDocument` corresponding to
+     * the given DOM Range and trigger a rendering.
+     *
+     * @param range
+     */
+    _setRange(range: Range): void {
+        const startPos: VPosition = this._getPosition(range.startContainer, range.startOffset);
+        const endPos: VPosition = this._getPosition(range.endContainer, range.endOffset);
+        const [first, last] = this._orderedRangeNodes(range['direction'] === 'rtl');
+
+        // Move the range nodes. If the start has to be moved *after* its
+        // reference node, the end should be moved first.
+        if (startPos.side === 'after') {
+            endPos.ref[endPos.side](last);
+            startPos.ref[startPos.side](first);
+        } else {
+            startPos.ref[startPos.side](first);
+            endPos.ref[endPos.side](last);
+        }
+        this.dispatcher.dispatch(this.intent({ name: 'render' }));
     }
 }
--- a/src/core/utils/Parser.ts
+++ b/src/core/utils/Parser.ts
@@ -56,6 +56,56 @@ export class Parser {
         }
     }
     /**
+     * Return true if the given node is immediately preceding (`side` === 'end')
+     * or following (`side` === 'start') a segment break, to see if its edge
+     * space must be removed.
+     *
+     * @param {Element} node
+     * @param {'start'|'end'} side
+     * @returns {boolean}
+     */
+    static _isAgainstSegmentBreak(node: Element, side: 'start' | 'end'): boolean {
+        const siblingSide = side === 'start' ? 'previousSibling' : 'nextSibling';
+        const sibling = node && (node[siblingSide] as DOMElement);
+        return (sibling && this._isBlock(sibling)) || this._isEdgeOfBlock(node, side);
+    }
+    /**
+     * Return true if the node is a block.
+     * Is considered a block, every node that is displayed as a block, and `BR`
+     * elements.
+     *
+     * @param {DOMElement} node
+     * @returns {boolean}
+     */
+    static _isBlock(node: DOMElement): boolean {
+        const display = window.getComputedStyle(node, null).display;
+        return display === 'block' || node.tagName === 'BR';
+    }
+    /**
+     * Return true if the node is at the given edge of a block.
+     *
+     * @param {Element | Node} node
+     * @param {'start'|'end'} side
+     */
+    static _isEdgeOfBlock(node: Element | Node, side: 'start' | 'end'): boolean {
+        const ancestorsUpToBlock = [];
+
+        // Move up to the first block ancestor
+        let ancestor = node;
+        const isTextNode = ancestor.nodeType === 3;
+        while (ancestor && (isTextNode || !this._isBlock(ancestor as DOMElement))) {
+            ancestorsUpToBlock.push(ancestor);
+            ancestor = ancestor.parentElement;
+        }
+
+        // Return true if no ancestor up to the first block ancestor has a
+        // sibling on the specified side
+        const siblingSide = side === 'start' ? 'previousSibling' : 'nextSibling';
+        return ancestorsUpToBlock.every(ancestor => {
+            return !ancestor[siblingSide];
+        });
+    }
+    /**
      * Parse an (non-text, non-format) element.
      *
      * @param {Element} node the node to parse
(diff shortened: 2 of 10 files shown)


src/core/utils/Parser.ts
```diff
@@ -55,6 +55,56 @@ export class Parser {
...
+    static _isEdgeOfBlock(node: Element | Node, side: 'start' | 'end'): boolean {
+        const ancestorsUpToBlock = [];
+
+        // Move up to the first block ancestor
+        let ancestor = node;
+        const isTextNode = ancestor.nodeType === 3;
+        while (ancestor && (isTextNode || !this._isBlock(ancestor as DOMElement))) {
+            ancestorsUpToBlock.push(ancestor);
+            ancestor = ancestor.parentElement;
+        }
```
Zinston: Should be:
```javascript
let isTextNode = ancestor.nodeType === 3;
while (ancestor && (isTextNode || !this._isBlock(ancestor as DOMElement))) {
    ancestorsUpToBlock.push(ancestor);
    ancestor = ancestor.parentElement;
    isTextNode = ancestor.nodeType === 3;
}
```

src/core/utils/CorePlugin.ts
```diff
@@ -20,7 +30,96 @@ export class CorePlugin extends JWPlugin {
...
+            rangeNode.order = i;
+        });
+        return res;
+    }
+    /**
+     * Move the position to its deepest first descendent.
+     *
+     * @param position
+     */
+    _positionToLeaf(position: VPosition): VPosition {
```
dmo-odoo: What about `_deepestPosition`, `_makePositionDeepest`, or any other idea that expresses the concept of deepest rather than leaf ? (Deepest is more obvious for me for some reason, but maybe you disagree ?)

src/core/utils/CorePlugin.ts
```diff
@@ -20,7 +30,96 @@ export class CorePlugin extends JWPlugin {
...
+
+    /**
+     * Return a position in the `VDocument` as a tuple `[VNode, 'before' | 'after']`
+     * (ie. the position is before or after a given `VNode`). The position is
+     * always given on the leaf.
+     *
+     * @param container
+     * @param offset
+     */
+    _getPosition(container: DOMElement | Node, offset: number): VPosition {
```
dmo-odoo: Why is a plugin playing with DOM elements ? IMO, only the parser and renderer should have references to DOM elements, all the other parts of the editor should completely ignore them and only work on our virtual document.

src/core/utils/CorePlugin.ts
```diff
@@ -20,7 +30,96 @@ export class CorePlugin extends JWPlugin {
...
+        // Move the range nodes. If the start has to be moved *after* its
+        // reference node, the end should be moved first.
+        if (startPos.side === 'after') {
+            endPos.ref[endPos.side](last);
+            startPos.ref[startPos.side](first);
+        } else {
+            startPos.ref[startPos.side](first);
+            endPos.ref[endPos.side](last);
+        }
+        this.dispatcher.dispatch(this.intent({ name: 'render' }));
```
dmo-odoo: I don't think plugins should be able to trigger a re-render (at least not until we face a true need for it.) IMO, the dispatcher calls the handlers for an intent, when the handlers have been called (potentially recursively) and they all got their chance at modifying the virtual document, the dispatcher can then trigger the commitState and the rendering process.

src/core/utils/CorePlugin.ts
```diff
@@ -20,7 +30,96 @@ export class CorePlugin extends JWPlugin {
...
+     * @param reverse
+     */
+    _orderedRangeNodes(reverse = false): [VNode, VNode] {
+        const rangeNodes = this.vDocument.rangeNodes;
+        const res: [VNode, VNode] = [rangeNodes.start, rangeNodes.end];
+        if (reverse) {
+            res.reverse();
+        }
+        res.forEach((rangeNode: VNode, i: 0 | 1) => {
+            rangeNode.order = i;
```
dmo-odoo: I once attented a conference where the speaker was arguing against using single letters (among others) for variable names as it did not give any indication of what the variable was supposed to represent. One of the guy in the room asked him about the `i` variable in loops. The speaker responded something along the line of "Are you so lazy that you can't call the variable  `index` or, even better, `whateverIndex` depending on your context, since that *is* what that variable is ?" and it rang very true to me. Basically, the only place where I still accept single-letter variables is in very short and obvious constructions like: `keys.forEach(k => k.trim())`. So TL;DR: rename `i` into either `index` or `nodeIndex`... ;-)

src/core/utils/CorePlugin.ts
```diff
@@ -20,7 +30,96 @@ export class CorePlugin extends JWPlugin {
...
+
+        // Move the range nodes. If the start has to be moved *after* its
+        // reference node, the end should be moved first.
+        if (startPos.side === 'after') {
+            endPos.ref[endPos.side](last);
+            startPos.ref[startPos.side](first);
+        } else {
+            startPos.ref[startPos.side](first);
+            endPos.ref[endPos.side](last);
+        }
```
dmo-odoo: I won't lie: I don't understand anything of what is going on in this function, except that we're supposedly moving the range... :/

src/core/utils/Renderer.ts
```diff
@@ -15,6 +22,14 @@ export class Renderer {
     // list of char nodes to aggregate:
     _currentCharNodes: VNode[] = [];
     _lastCreatedNode: DOMElement | Node;
+    _nextEndOffset = 0;
+    _nextStartOffset = 0;
+    _rangeToSet: RangeToSet;
+    _waitedForFormat = false;
+    _delayingRange = {
+        start: false,
+        end: false,
```
dmo-odoo: In the renderer implementation I suggested on the other commit, I think the range nodes, when encountered, should just be stored in a variable and treated separately when the rest of the nodes have been rendered. I'm not reviewing the rendering part of this commit since I expect much will change because of the review of earlier commits (and also because that PR is huge ! :p)

src/plugins/DevTools/components/DevToolsComponent.ts
```diff
@@ -51,6 +49,20 @@ export class DevToolsComponent extends OwlUIComponent<{}, DevToolsState> {
...
+     */
+    onRender(): void {
+        this.render();
+        const inspectorComponent = this.refs.InspectorComponent as InspectorComponent;
+        if (inspectorComponent) {
+            const rangeNodes = this.env.editor.vDocument.rangeNodes;
+            inspectorComponent._selectNode(rangeNodes.start);
+            // In order for owl-framework to unfold the parents of start *and*
+            // of end, we need to select both, with a tick in between.
+            setTimeout(() => inspectorComponent._selectNode(rangeNodes.end), 0);
```
dmo-odoo: In the function that computes whether a node should be folded or not, why not check if it contains the range instead ?

src/core/utils/Renderer.ts
```diff
@@ -0,0 +1,197 @@
...
+        if (this.vDocument.domMap.has(element)) {
+            const matches = this.vDocument.domMap.get(element);
+            if (!matches.some((match: VNode): boolean => match.id === vNode.id)) {
+                matches.push(vNode);
+            }
+        } else {
+            this.vDocument.domMap.set(element, [vNode]);
+        }
+    }
+}
```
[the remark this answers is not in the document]
Zinston: Just for readability purposes, this is a copy paste of your code but with syntax highlighting.

```typescript
// File: BasicDOMRenderer.ts
const getRenderingTagName = function(node: VNode) {
    switch (node.type) {
        case 'CHAR':
            return;
        case 'HEADING1':
        case 'HEADING2':
        case 'HEADING3':
        case 'HEADING4':
        case 'HEADING5':
        case 'HEADING6':
            return 'H' + node.type[7];
        case 'LINE_BREAK':
            return 'BR';
        case 'PARAGRAPH':
            return 'P';
    }
}

export const BasicDOMRenderer = {
    render: function(node: VNode): Element {
        const tagName = getRenderingTagName(node.type);
        return document.createElement(tagName);
    }
}

// File: VNode.ts
import {BasicDOMRenderer} from 'BasicDOMRenderer';
export type VNodeRenderingMethod = (node: VNode) => Element;
VNode {
    // More complex node will use more complex rendering methods from more complex renderers.
    render: VNodeRenderingMethod = BasicDOMRenderer.render.bind(this);

    nextSibling: VNode {
        // To implement.
    }

    getParent: VNode {
        // To implement.
    }
}

// File: Renderer.ts
_renderNode(node: VNode, parent: Element | DocumentFragment): void {
    if (node.type === 'CHAR') {
        // If the node has a format, render the format nodes first.
        Object.keys(node.format).forEach(type => {
            if (node.format[type]) {
                const formatNode = document.createElement(formatToTag(type));
                parent.appendChild(formatNode);
                // Update the parent so the text is inside the format node.
                parent = formatNode;
            }
        })

        // Consecutive compatible char nodes are rendered as a single text node.
        let text = node.value;
        let next = node.nextSibling();
        while (next && next.type === 'CHAR' && this._isSameFormat(node, next)) {
            text += next.value;
            node = next;
            next = node.nextSibling();
        }
        parent.appendChild(document.createTextNode(text));

    } else {
        // Create the element matching this node and append it.
        parent.appendChild(node.render());
    }

    this._renderNextNode(node, parent);
}

_renderNextNode(node: VNode, parent: Element | DocumentFragment): void {
    if(node.children.length) {
        // Render the first child with the current node as parent, if any.
        this._renderNode(node.children[0], node);
    } else if (node.nextSibling()) {
        // Render the siblings of the current node with the same parent, if any.
        this._renderNode(nextSibling, parent);
    } else {
        // Render the next ancestor sibling in the ancestor tree, if any.
        let ancestor = node.getParent();
        // Climb back the ancestor tree to the first parent having a sibling.
        while (ancestor && !ancestor.nextSibling()) {
            ancestor = ancestor.getParent();
        }
        // At this point, the found ancestor has a sibling. If no ancestor
        // having a sibling could be found, the tree has been fully rendered.
        if (ancestor) {
            this._renderNode(ancestor.nextSibling(), ancestor.getParent());
        }
    }
}

_isSameFormat(a: VNode, b: VNode) {
    return Object.values(a.format) === Object.values(b.format);
}
```

src/core/utils/Parser.ts
```diff
@@ -55,6 +55,56 @@ export class Parser {
...
+    /**
+     * Return true if the given node is immediately preceding (`side` === 'end')
+     * or following (`side` === 'start') a segment break, to see if its edge
+     * space must be removed.
+     *
+     * @param {Element} node
+     * @param {'start'|'end'} side
+     * @returns {boolean}
+     */
+    static _isAgainstSegmentBreak(node: Element, side: 'start' | 'end'): boolean {
```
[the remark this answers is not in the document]
Zinston: Added to docstring:
A segment break is a sort of line break, not considering automatic breaks that are function of the screen size. In this context, a segment is what you see when you triple click in text in the browser.
Eg: `<div><p>◆one◆</p>◆two◆<br>◆three◆</div>` where ◆ = segment breaks.

src/core/utils/Parser.ts
```diff
@@ -55,6 +55,56 @@ export class Parser {
...
+        const siblingSide = side === 'start' ? 'previousSibling' : 'nextSibling';
+        const sibling = node && (node[siblingSide] as DOMElement);
+        return (sibling && this._isBlock(sibling)) || this._isEdgeOfBlock(node, side);
```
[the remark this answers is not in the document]
Zinston: Absolutely right, though we technically need CSS to determine whether a node is a block box or not.

src/core/utils/Parser.ts
Zinston: Wel... Then `_isBlock` would be testing more than it should: you can _not_ be a block and still be something else than a text node.

src/core/utils/Parser.ts
```diff
@@ -128,11 +178,78 @@ export class Parser {
...
         const parsedNodes: VNode[] = [];
-        for (let i = 0; i < node.textContent.length; i++) {
-            const char: string = node.textContent.charAt(i);
-            const parsedNode: VNode = new VNode(VNodeType.CHAR, node.tagName, char, format);
+        const text: string = this._removeFormatSpace(node);
+        for (let i = 0; i < text.length; i++) {
+            const char: string = text.charAt(i);
+            const parsedNode: VNode = new VNode(
+                VNodeType.CHAR,
+                node.tagName || '#text',
```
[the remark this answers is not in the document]
Zinston: `nodeName` should do the trick. It's purely for representation purposes and indeed it's not really needed.

src/core/utils/Parser.ts
```diff
@@ -128,11 +178,78 @@ export class Parser {
...
+                char,
+                format,
+            );
             parsedNodes.push(parsedNode);
         }
         return parsedNodes;
     }
+    /**
+     * Return a string with the value of a text node stripped of its format space,
```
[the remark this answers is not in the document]
Zinston: Absolutely.

src/core/utils/Parser.ts
```diff
@@ -128,11 +178,78 @@ export class Parser {
...
+        const consecutiveSpace = /  */g;
+
+        // (Comments refer to the w3 link provided above.)
+        // Phase I: Collapsing and Transformation
+        let newText = text
+            // 1. All spaces and tabs immediately preceding or following a
+            //    segment break are removed.
+            .replace(spaceBeforeNewline, '$2')
+            .replace(spaceAfterNewline, '$1')
+            // 2. Segment breaks are transformed for rendering according to the
```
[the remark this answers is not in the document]
Zinston: This is a copy paste from w3.

src/core/utils/Parser.ts
```diff
@@ -128,11 +178,78 @@ export class Parser {
...
+            .replace(newlines, ' ')
+            // 3. Every tab is converted to a space (U+0020).
+            .replace(tabs, ' ')
+            // 4. Any space immediately following another collapsible space —
+            //    even one outside the boundary of the inline containing that
+            //    space, provided both spaces are within the same inline
+            //    formatting context—is collapsed to have zero advance width.
+            //    (It is invisible, but retains its soft wrap opportunity, if
+            //    any.)
+            .replace(consecutiveSpace, ' ');
```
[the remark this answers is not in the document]
Zinston: Idem. The whole whitespace thing is a mess I know.

src/core/utils/Parser.ts
```diff
@@ -128,11 +178,78 @@ export class Parser {
...
+
+        // Phase II: Trimming and Positioning
+        // 1. A sequence of collapsible spaces at the beginning of a line
+        //    (ignoring any intervening inline box boundaries) is removed.
+        if (this._isAgainstSegmentBreak(node, 'start')) {
+            const startSpace = /^ */g;
+            newText = newText.replace(startSpace, '');
```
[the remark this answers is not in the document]
Zinston: Newline as in \n in a text, beginning of a line as in beginning of block, or after <br>, or after a block.

src/core/utils/Parser.ts
```diff
@@ -128,11 +178,78 @@ export class Parser {
...
+        }
+        // 2. If the tab size is zero, tabs are not rendered. Otherwise, each
+        //    tab is rendered as a horizontal shift that lines up the start edge
+        //    of the next glyph with the next tab stop. If this distance is less
+        //    than 0.5ch, then the subsequent tab stop is used instead. Tab
+        //    stops occur at points that are multiples of the tab size from the
+        //    block’s starting content edge. The tab size is given by the
+        //    tab-size property.
+        // TODO
```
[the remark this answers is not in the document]
Zinston: Right?

src/core/utils/Parser.ts
```diff
@@ -128,11 +178,78 @@ export class Parser {
...
+        // 3. A sequence at the end of a line (ignoring any intervening inline
+        //    box boundaries) of collapsible spaces (U+0020) and/or ideographic
+        //    spaces (U+3000) whose white-space value collapses spaces is
+        //    removed.
+        if (this._isAgainstSegmentBreak(node, 'end')) {
+            const endSpace = /[ \u3000]*$/g;
+            newText = newText.replace(endSpace, '');
+        }
```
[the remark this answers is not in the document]
Zinston: Idem.

Zinston:
This PR was split into https://github.com/odoo-dev/jabberwock/pull/12 and https://github.com/odoo-dev/jabberwock/pull/13

src/core/utils/CorePlugin.ts
```diff
@@ -20,7 +30,96 @@ export class CorePlugin extends JWPlugin {
...
+    }
+
+    //--------------------------------------------------------------------------
+    // Private
+    //--------------------------------------------------------------------------
+
+    /**
+     * Return a position in the `VDocument` as a tuple `[VNode, 'before' | 'after']`
+     * (ie. the position is before or after a given `VNode`). The position is
+     * always given on the leaf.
```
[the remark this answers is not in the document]
Zinston: Yes. The tree's leaf on that particular branch.

src/core/utils/CorePlugin.ts
Zinston: As discussed, we will indeed ensure that this is done by the `Renderer`. I had done it here because at the time what we had agreed upon was to have this `CorePlugin` handle the connection between the DOM and the `VDocument`. But that was always meant to be temporary, until the memory structure is introduced and the whole world falls on its head.

src/core/utils/CorePlugin.ts
Zinston: I like leaf but I don't mind changing the name, I'm not particularly attached to it so if it's clearer to you that way I'll gladly comply ;-)

src/core/utils/CorePlugin.ts
Zinston: 1. Getting the position where the range nodes should be (before or after which node)
2. Putting it there by calling `before()` or `after()` on the reference node and passing it the correct range node

src/core/utils/CorePlugin.ts
Zinston: This should indeed not remain once we have a mechanism for "committing" the changes to the `VDocument`, which depends on CHM's memory system. But indeed I can move this to the dispatcher, at the end of the loop, though right now we have no mechanism to ensure that the flow is over and we don't need to run the loop again.

```typescript
// In `Dispatcher`:
if (Object.keys(handlers).length && action.name !== 'render') {
    this.dispatch(ActionGenerator.intent({ name: 'render' }));
} else {
        console.warn(`No plugin is listening to the ${action.type} "${action.name}".`);
}
```

PR #12, opened 2019-10-16
[POC] Renderer: first version of a basic renderer

Zinston:
Also includes some related improvements.

This is the first part of a split of https://github.com/odoo-dev/jabberwock/pull/7 into two PRs.

### DIFF
--- a/src/core/stores/VNode.ts
+++ b/src/core/stores/VNode.ts
@@ -1,3 +1,5 @@
+import { BasicHtmlRenderingEngine, RenderingEngine } from '../utils/BasicHtmlRenderingEngine';
+
 export enum VNodeType {
     ROOT = 'ROOT',
     PARAGRAPH = 'PARAGRAPH',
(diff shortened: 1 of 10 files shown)


src/core/stores/VNode.ts
```diff
@@ -62,94 +62,234 @@ export class VNode {
...
+    //--------------------------------------------------------------------------
+
+    /**
+     * Insert a VNode at the given index within this VNode's children.
+     * Return self.
+     *
+     * @param child
+     * @param index
+     */
+    _insertAtIndex(child: VNode, index: number): VNode {
```
Zinston: This should start with:
```typescript
if (child.parent) {
    child.parent.removeChild(child);
}
```
This way if we insert an element that already is in the `vDocument`, it's moved, not duplicated.

src/core/stores/VNode.ts
```diff
@@ -62,94 +62,241 @@ export class VNode {
...
+    childAfter(child: VNode): VNode {
+        return this.nthChild(this.indexOf(child) + 1);
+    }
+    /**
+     * Return the child at given index.
+     *
+     * @param index
+     */
+    nthChild(index: number): VNode {
+        return (index >= 0 && this.children[index]) || undefined;
```
dmo-odoo: If index < 0, this.children[index] will be undefined anyway.

PR #24, opened 2019-10-28
Master rangenode cleaner age

dmo-odoo:
As of 28 October 2019, this is the latest version of #13 (and therefore of #7).


dmo-odoo:
Most commits from this PR now live at https://github.com/odoo-dev/jabberwock/pull/28.

The commits that were not moved in that PR are the following:
- [FIX] Renderer: append placeholder br to empty elements
- [FIX] Renderer: render a right edge LINE_BREAK
- [IMP] DevTools: select node in dom on dblClick in DevTools
- fix renderer: always target leaf

The renderer fixes were not moved because I rewrote the renderer to a point where I have no idea whether they apply or not. I don't know how to trigger the case they are supposed to fix, and I don't know why these commits are supposed to fix it, so I am unable to reapply them in my implementation of the rendering of the range nodes.

The DevTools commit was not moved because the feature does not seem to work anymore with latest owl due to issue https://github.com/odoo/owl/issues/420 in which ged shamefuly copy-pasted what I told him in french on Discuss... :p

I'm not closing this PR since there are still some commits that @Zinston might want to recover.

PR #29, opened 2019-10-30
[IMP] test: specify the kind of test to run

Goaman:


package.json
```diff
@@ -7,7 +7,9 @@
     "dev": "webpack-dev-server",
     "build": "webpack",
     "postinstall": "npm link owl-framework",
-    "test": "karma start"
+    "test-all": "karma start",
+    "test-perf": "karma start --include-files test/**/*.perf.test.ts",
+    "test": "karma start --exclude-files test/**/*.perf.test.ts"
```
dmo-odoo: I would have done `npm run test` and `npm run perf`, but then I admit there would be no good way to name the "test-all" version. Is this really useful to have a "all" version ?

Performance tests are very different than behavior tests IMO and I'm not sure we really need a way to run both of them in one command. The mean reason why I dislike performance tests (and so the reason why I wanted to exlude them from `npm run test` in the first place) is because their performance is so very very much tied to the performance of your machine that I consider them irrelevant from the code testing process. I understand why we need them, but I don't see myself running them very often, especially on my Macbook Air.

Moreover, what is the point of testing the performance if the behavior is wrong ? Example: the ORM of Odoo could dramatically perform better if it just stopped testing a lot of cases, so the performance would be "woah!" but a part of the test suite would not work anymore, so the performance tests would be completely meaningless in such case. 

IMO, I would run `npm run test` then, only if the tests passed successfully and I am confident that my computer performances are ok (like: I'm not at 5% battery) then I'd run `npm run perf`. If we have a `npm run test-all` then it tends to mean you are doing it wrong if you don't use this one all the time since the others "don't test everything" since they are not "test-all", so you'd have to run `npm run test-all` all the time, which means we're back to the initial state that I wanted to avoid. 

I would definitely keep them as two separated test suite: 'test' and 'perf'. What do you think ?

PR #33, opened 2019-11-06
Miscellaneous fixes and improvements

Zinston:
This commit contains the missing commits from https://github.com/odoo-dev/jabberwock/pull/24, as well as a commit to hide the range nodes when traversing the VDocument, and a couple more fixes that were required along the way.

### DIFF
--- a/src/core/JWEditor.ts
+++ b/src/core/JWEditor.ts
@@ -26,11 +26,17 @@ export class JWEditor {
 
     constructor(editable?: HTMLElement) {
         this.el = document.createElement('jw-editor');
+        // Semantic elements are inline by default.
+        // We need to guarantee it's a block so it can contain other blocks.
+        this.el.style.display = 'block';
         this.dispatcher = new Dispatcher(this);
         this.pluginsRegistry = [];
 
         if (!editable) {
             editable = document.createElement('jw-editable');
+            // Semantic elements are inline by default.
+            // We need to guarantee it's a block so it can contain other blocks.
+            editable.style.display = 'block';
         }
         this._originalEditable = editable;
 
--- a/src/core/stores/VDocument.ts
+++ b/src/core/stores/VDocument.ts
@@ -1,5 +1,5 @@
 import { VNode, VNodeType } from './VNode';
-import { VRange, Direction } from './VRange';
+import { VRange } from './VRange';
 
 export let withRange = false;
 
@@ -53,9 +53,8 @@ export class VDocument {
         VDocument.withRange(() => {
             const nodes = this.range.selectedNodes;
             if (!nodes.length) return;
-            this.range.collapse(this.range.direction === Direction.FORWARD ? 'start' : 'end'); // Reset the direction of the range.
-            let reference =
-                this.range.direction === Direction.FORWARD ? this.range.end : this.range.start;
+            this.range.collapse(this.range.start); // Reset the direction of the range.
+            let reference = this.range.end;
             nodes.forEach(vNode => {
                 // If the node has children, merge it with the container of the
                 // range. Children of the merged node that should be truncated
--- a/src/core/stores/VDocument.ts
+++ b/src/core/stores/VDocument.ts
@@ -1,6 +1,8 @@
 import { VNode, VNodeType } from './VNode';
 import { VRange } from './VRange';
 
+export let withRange = false;
+
 export class VDocument {
     root: VNode;
     range = new VRange();
@@ -51,34 +53,54 @@ export class VDocument {
      */
     truncate(nodes: VNode[]): void {
         if (!nodes.length) return;
-        // Determine where and how to reattach the orphaned child nodes.
-        const firstNode = nodes[0];
-        let reference = firstNode.previousSibling();
-        let methodName = 'after';
-        if (!reference) {
-            reference = firstNode.nextSibling();
-            methodName = 'before';
-        }
-        if (!reference) {
-            reference = firstNode.parent;
-            methodName = 'append';
-        }
-        nodes.forEach(vNode => {
-            // If the node has children, merge it with the container of the
-            // range. Children of the merged node that should be truncated as
-            // well will be deleted in the following iterations since they
-            // appear in `nodes`. The children array must be cloned in order to
-            // modify it while iterating.
-            // TODO: test whether the node can be merged with the container.
-            if (vNode.children.length) {
-                vNode.children.slice().forEach(child => {
-                    reference[methodName](child);
-                    reference = child;
-                });
+        VDocument.withRange(() => {
+            // Determine where and how to reattach the orphaned child nodes.
+            const firstNode = nodes[0];
+            let reference = firstNode.previousSibling();
+            let methodName = 'after';
+            if (!reference) {
+                reference = firstNode.nextSibling();
+                methodName = 'before';
             }
-            // Then remove.
-            vNode.remove();
+            if (!reference) {
+                reference = firstNode.parent;
+                methodName = 'append';
+            }
+            nodes.forEach(vNode => {
+                // If the node has children, merge it with the container of the
+                // range. Children of the merged node that should be truncated as
+                // well will be deleted in the following iterations since they
+                // appear in `nodes`. The children array must be cloned in order to
+                // modify it while iterating.
+                // TODO: test whether the node can be merged with the container.
+                if (vNode.children.length) {
+                    vNode.children.slice().forEach(child => {
+                        reference[methodName](child);
+                        reference = child;
+                    });
+                }
+                // Then remove.
+                vNode.remove();
+            });
+            this.range.collapse(); // Reset the direction of the range.
         });
-        this.range.collapse(); // Reset the direction of the range.
+    }
+
+    //--------------------------------------------------------------------------
+    // Context
+    //--------------------------------------------------------------------------
+
+    /**
+     * Call a callback on this VNode without ignoring the range nodes.
+     *
+     * @param callback
+     */
+    static withRange<T>(callback: () => T): T {
+        // Record the previous value to allow for nested calls to `withRange`.
+        const previousValue = withRange;
+        withRange = true;
+        const result = callback();
+        withRange = previousValue;
+        return result;
     }
 }
--- a/src/core/stores/VDocument.ts
+++ b/src/core/stores/VDocument.ts
@@ -1,5 +1,5 @@
 import { VNode, VNodeType } from './VNode';
-import { VRange } from './VRange';
+import { VRange, Direction } from './VRange';
 
 export let withRange = false;
 
@@ -23,7 +23,7 @@ export class VDocument {
     insert(node: VNode): void {
         // Remove the contents of the selection if needed.
         if (!this.range.isCollapsed()) {
-            this.truncate(this.range.selectedNodes);
+            this.deleteSelection();
         }
         this.range.start.before(node);
     }
@@ -35,7 +35,7 @@ export class VDocument {
     insertText(text: string): void {
         // Remove the contents of the selection if needed.
         if (!this.range.isCollapsed()) {
-            this.truncate(this.range.selectedNodes);
+            this.deleteSelection();
         }
         // Split the text into CHAR nodes and insert them at the range.
         const characters = text.split('').reverse();
@@ -46,43 +46,32 @@ export class VDocument {
         });
     }
     /**
-     * Truncate the tree by removing the given nodes and merging their orphaned
-     * children into the parent of the first removed node.
-     *
-     * @param nodes
+     * Truncate the tree by removing the selected nodes and merging their
+     * orphaned children into the parent of the first removed node.
      */
-    truncate(nodes: VNode[]): void {
-        if (!nodes.length) return;
+    deleteSelection(): void {
         VDocument.withRange(() => {
-            // Determine where and how to reattach the orphaned child nodes.
-            const firstNode = nodes[0];
-            let reference = firstNode.previousSibling();
-            let methodName = 'after';
-            if (!reference) {
-                reference = firstNode.nextSibling();
-                methodName = 'before';
-            }
-            if (!reference) {
-                reference = firstNode.parent;
-                methodName = 'append';
-            }
+            const nodes = this.range.selectedNodes;
+            if (!nodes.length) return;
+            this.range.collapse(this.range.direction === Direction.FORWARD ? 'start' : 'end'); // Reset the direction of the range.
+            let reference =
+                this.range.direction === Direction.FORWARD ? this.range.end : this.range.start;
             nodes.forEach(vNode => {
                 // If the node has children, merge it with the container of the
-                // range. Children of the merged node that should be truncated as
-                // well will be deleted in the following iterations since they
-                // appear in `nodes`. The children array must be cloned in order to
-                // modify it while iterating.
+                // range. Children of the merged node that should be truncated
+                // as well will be deleted in the following iterations since
+                // they appear in `nodes`. The children array must be cloned in
+                // order to modify it while iterating.
                 // TODO: test whether the node can be merged with the container.
-                if (vNode.children.length) {
+                if (vNode.hasChildren()) {
                     vNode.children.slice().forEach(child => {
-                        reference[methodName](child);
+                        reference.after(child);
                         reference = child;
                     });
                 }
-                // Then remove.
-                vNode.remove();
             });
-            this.range.collapse(); // Reset the direction of the range.
+            // Then remove.
+            nodes.forEach(vNode => vNode.remove());
         });
     }
 
--- a/src/core/stores/VNode.ts
+++ b/src/core/stores/VNode.ts
@@ -408,6 +408,30 @@ export class VNode {
         }
     }
     /**
+     * Return the previous leaf in a depth-first pre-order traversal of the
+     * tree that satisfies the given predicate. If no predicate is given return
+     * the previous leaf in a depth-first pre-order traversal of the tree.
+     *
+     * @param [predicate]
+     */
+    previousLeaf(predicate?: Predicate): VNode {
+        return this.previous((node: VNode): boolean => {
+            return !node.hasChildren() && (!predicate || predicate(node));
+        });
+    }
+    /**
+     * Return the next leaf in a depth-first pre-order traversal of the tree
+     * that satisfies the given predicate. If no predicate is given return the
+     * next leaf in a depth-first pre-order traversal of the tree.
+     *
+     * @param [predicate]
+     */
+    nextLeaf(predicate?: Predicate): VNode {
+        return this.next((node: VNode): boolean => {
+            return !node.hasChildren() && (!predicate || predicate(node));
+        });
+    }
+    /**
      * Walk the document tree starting from the current node (included) by
      * calling the `next` iterator until the returned node satisfies the given
      * predicate or is falsy.
--- a/src/core/stores/VNode.ts
+++ b/src/core/stores/VNode.ts
@@ -83,7 +83,7 @@ export class VNode {
     totalLength(__current = 0): number {
         __current += this.length;
         this.children.forEach((child: VNode): void => {
-            if (child.children.length) {
+            if (child.hasChildren()) {
                 __current = child.totalLength(__current);
             }
         });
@@ -118,6 +118,12 @@ export class VNode {
             __current = child.text(__current);
         });
         return __current;
+    }
+    /**
+     * Return true if this VNode has children.
+     */
+    hasChildren(): boolean {
+        return this.children.length > 0;
     }
     /**
      * Return true if this VNode has a format property set to true.
@@ -267,7 +273,7 @@ export class VNode {
      */
     firstLeaf(predicate?: Predicate): VNode {
         const isValidLeaf = (node: VNode): boolean => {
-            return !node.children.length && (!predicate || predicate(node));
+            return !node.hasChildren() && (!predicate || predicate(node));
         };
         if (isValidLeaf(this)) {
             return this;
@@ -283,7 +289,7 @@ export class VNode {
      */
     lastLeaf(predicate?: Predicate): VNode {
         const isValidLeaf = (node: VNode): boolean => {
-            return !node.children.length && (!predicate || predicate(node));
+            return !node.hasChildren() && (!predicate || predicate(node));
         };
         if (isValidLeaf(this)) {
             return this;
@@ -313,7 +319,7 @@ export class VNode {
      */
     lastDescendant(predicate?: Predicate): VNode {
         let lastDescendant = this.lastChild();
-        while (lastDescendant && lastDescendant.children.length) {
+        while (lastDescendant && lastDescendant.hasChildren()) {
             lastDescendant = lastDescendant.lastChild();
         }
         if (lastDescendant && predicate) {
--- a/src/core/stores/VNode.ts
+++ b/src/core/stores/VNode.ts
@@ -1,4 +1,5 @@
 import { BasicHtmlRenderingEngine, RenderingEngine } from '../utils/BasicHtmlRenderingEngine';
+import { withRange, VDocument } from './VDocument';
 
 export type Predicate = (node: VNode) => boolean;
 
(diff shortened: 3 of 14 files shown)


Zinston:
I just realized two issues:
- commit https://github.com/odoo-dev/jabberwock/pull/33/commits/f8c455145c42124fc0ae95334bc285b1b63c9df7 is broken
- commit https://github.com/odoo-dev/jabberwock/pull/33/commits/ecd1f7730720cae3518ba1849f4743a78020574b may render a trailing line break properly but it fails to account for the range (impossible to set the range on the trailing br)

src/core/stores/VNode.ts
```diff
@@ -453,6 +453,32 @@ export class VNode {
             return next;
         }
     }
+    /**
+     * Return the previous leaf in a depth-first pre-order traversal of the
+     * tree that satisfies the given predicate. If no predicate is given return
+     * the previous leaf in a depth-first pre-order traversal of the tree.
+     * A leaf is a childless node.
```
dmo-odoo: If the guy does not know what a tree "leaf" is in programming, then it's highly unlikely he would have understood "depth-first pre-order traversal" or "predicate" anyway, just sayin'... ;-) (same for `nextLeaf`)

src/core/stores/VNode.ts
```diff
@@ -453,6 +453,32 @@ export class VNode {
...
+     * Return the next leaf in a depth-first pre-order traversal of the tree
+     * that satisfies the given predicate. If no predicate is given return the
+     * next leaf in a depth-first pre-order traversal of the tree.
+     * A leaf is a childless node.
+     *
+     * @param [predicate]
+     */
+    nextLeaf(predicate?: Predicate): VNode {
+        return this.next((node: VNode): boolean => {
+            return (!predicate || predicate(node)) && !node.children.length;
```
dmo-odoo: I would test the children first, as I expect most nodes are not leaves, so checking "being a leaf" first should be more efficient in a world where people might have crafted very complicated predicates.

src/core/utils/CorePlugin.ts
```diff
@@ -39,19 +40,33 @@ export class CorePlugin extends JWPlugin {
...
     }
+    /**
+     * Remove in the backward direction (backspace key expected behavior).
+     */
     removeBackward(): void {
-        // TODO: this is a stub
         const range = this.editor.vDocument.range;
         if (range.isCollapsed()) {
-            range.setStart(range.start.previous());
+            const previous = VDocument.withRange(() => range.start.previousLeaf());
```
dmo-odoo: Why would you want that `previousLeaf` may return a range node ? (same for `removeForward`)

src/core/utils/Renderer.ts
```diff
@@ -128,6 +130,13 @@ export class Renderer {
         if (!context.currentVNode.children.length && !context.currentVNode.properties.atomic) {
             this._insertFiller(context.currentVNode, element);
         }
+
+        // If a LINE_BREAK has no next sibling, it must be rendered as two BRs
+        // in order for it to be visible.
+        const nextSibling = context.currentVNode.nextSibling();
+        if (!__isSecondBR && context.currentVNode.type === VNodeType.LINE_BREAK && !nextSibling) {
+            return this._renderElement(context, true);
+        }
```
dmo-odoo: What about the BasicHTMLRenderingEngine of the BR node doing that instead ? Makes more sense to me.

src/plugins/DevTools/components/TreeComponent.ts
```diff
@@ -69,6 +71,18 @@ export class TreeComponent extends OwlUIComponent<NodeProps> {
...
     }
+    onDblClickNode(): void {
+        const location: VRangeDescription = {
+            start: this.props.vNode,
+            direction: Direction.FORWARD,
+        };
+        this.env.editor.dispatcher.dispatch(
+            ActionGenerator.intent({
+                name: 'setRange',
+                payload: { vRangeLocation: location },
```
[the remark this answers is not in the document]
Zinston: image

src/core/utils/Renderer.ts
```diff
@@ -200,7 +200,9 @@ export class Renderer {
             position = RelativePosition.INSIDE;
         }
         // The location is a tuple [reference, offset] implemented by an array.
-        const location = VDocumentMap.toDomLocation(reference);
+        const location = VDocumentMap.toDomLocation(
+            position === RelativePosition.BEFORE ? reference.firstLeaf() : reference.lastLeaf(),
+        );
```
Zinston: Move this further in the PR and use nextLeaf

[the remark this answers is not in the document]
Zinston: Not really a FIX, I should have put IMP I guess.

src/core/stores/VDocument.ts
```diff
@@ -51,34 +55,53 @@ export class VDocument {
...
-            // well will be deleted in the following iterations since they
-            // appear in `nodes`. The children array must be cloned in order to
-            // modify it while iterating.
-            // TODO: test whether the node can be merged with the container.
-            if (vNode.children.length) {
-                vNode.children.slice().forEach(child => {
-                    reference[methodName](child);
-                    reference = child;
-                });
+        VDocument.withRange(() => {
```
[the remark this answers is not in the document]
Zinston: We want to move the range nodes with the rest so it reranges automatically.

src/core/stores/VDocument.ts
```diff
@@ -51,34 +55,53 @@ export class VDocument {
...
+    // Context (technical)
+    //--------------------------------------------------------------------------
+
+    /**
+     * Call a method on this VNode without ignoring the range nodes.
+     *
+     * @param method
+     */
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    static withRange(method: () => any): any {
```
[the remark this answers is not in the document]
Zinston: Right, thanks, I forgot about that ◔_◔

src/core/stores/VNode.ts
```diff
@@ -18,7 +19,9 @@ export enum VNodeType {
 }
 export const RangeTypes = [VNodeType.RANGE_START, VNodeType.RANGE_END];
 export const TextTypes = [VNodeType.CHAR, VNodeType.RANGE_START, VNodeType.RANGE_END];
-export const ElementTypes = Object.keys(VNodeType).filter(t => !t.startsWith('RANGE'));
+export const ElementTypes = Object.keys(VNodeType).filter(
+    (t: VNodeType) => !RangeTypes.includes(t),
+);
```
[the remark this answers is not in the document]
Zinston: Aha! Finally!

src/core/stores/VNode.ts
```diff
@@ -516,15 +556,25 @@ export class VNode {
...
+    /**
+     * Return true if the context implies showing the range nodes instead of
+     * hiding them when traversing the document (default is false).
+     *
+     * @see VDocument.withRange
+     * @see VDocument.context
+     */
+    get _withRange(): boolean {
+        return context.withRange;
+    }
```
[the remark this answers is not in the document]
Zinston: I figured we might need other context information in the future and that this would fit into that but that might have been overzealous.

src/core/stores/VNode.ts
```diff
@@ -91,18 +103,22 @@ export class VNode {
     }
     /**
      * Return the index of this VNode within its parent.
+     * Note: this method always takes account of range nodes.
```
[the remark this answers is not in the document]
Zinston: _children can be set, children can't. How would we handle all the splices etc on children?

src/plugins/DevTools/DevTools.xml
```diff
@@ -81,7 +81,7 @@
     <t t-name="treeChildren">
         <div class="children">
             <TreeComponent t-if="!state.folded"
-                t-foreach="props.vNode.children" t-as="child"
+                t-foreach="props.vNode._children" t-as="child"
```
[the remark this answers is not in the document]
Zinston: I had made a public counterpart for this specific purpose and you told me it was better to get rid of it, oh you little! 
image

src/core/utils/Renderer.ts
Zinston: You mean like this?

```typescript
export const BasicHtmlRenderingEngine = {
    /**
     * Render the given node to HTML.
     *
     * @param node
     */
    render: function(node: VNode): HTMLElement[] {
        const tagName = getRenderingTagName(node.type);
        const element = document.createElement(tagName);
        if (node.type === VNodeType.LINE_BREAK && !node.nextSibling()) {
            return [element, element.cloneNode()];
        }
        return element;
    },
};
```
and modify `_renderElement` to handle lists of elements?

src/core/stores/VNode.ts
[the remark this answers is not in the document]
Zinston: Done. Also removed `TextTypes` and `ElementTypes`.

src/core/stores/VRange.ts
```diff
@@ -167,14 +170,16 @@ export class VRange {
...
-        if (position === RelativePosition.AFTER) {
-            reference.after(this.start);
-        } else {
-            reference.before(this.start);
-        }
-        this._direction = null; // Invalidate range direction cache.
-        return this;
+        return VDocument.withRange(() => {
+            reference = reference.firstLeaf();
+            if (position === RelativePosition.AFTER && reference !== this.end) {
```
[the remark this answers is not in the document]
Zinston: We can't set the start range as next sibling of the end range: that would be a backward collapsed range, which is not a thing in our abstraction (collapsed is always forward). I guess this is a fix but I think the problem only became visible with this commit.

src/core/utils/Renderer.ts
Zinston: I ended up doing that, and the `_insertFiller` thing. I put the part where I return a `DocumentFragment` in a separate commit so it can be used by the other two commits.

Zinston:
@dmo-odoo I applied the changes that I could, and answered the rest. There's a new commit in there, to return `DocumentFragment`s in `BasicHtmlRenderingEngine`'s `render` method, so that the `br` wizardry could be handled there instead of in the `Renderer`. I also put commit messages on the fixes so it's clearer. That includes the ones you didn't review.

src/core/stores/VNode.ts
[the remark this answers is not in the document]
dmo-odoo: What about we don't and the guy just has to use the methods ? Maybe children might be an iterator to prevent accessing the array and be more efficient when walking the entire dom (as opposed to cloning/filtering the array like right now), then we would need an `hasChildren` method like the `hasChildNodes` of the dom. What do you think ? We would lose the ability to do `node.children.length`, but the API would be entirely consistent with respect to whether range nodes should be taken into account or not, we would not clone the children array anymore, and we would ensure people can't play with the children array itself directly.

src/core/stores/VDocument.ts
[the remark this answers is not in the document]
dmo-odoo: Then I guess we only need it for when we move children ?

src/core/stores/VRange.ts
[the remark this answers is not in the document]
dmo-odoo: But this wouldn't happen if we don't do `withRange`, would it ? We really need to talk about the uses of `withRange`. I thought we would only use it when we specifically wanted to encounter the range, but here we are using it (for a reason I do not understand but it doesn't necessarily mean it's wrong) then, if we encounter it, we have a special if to protect us, which is specifically what I wanted to avoid in the first place.

src/plugins/DevTools/DevTools.xml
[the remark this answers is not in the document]
dmo-odoo: We need to find a way to allow the owl templates to use `withRange` in some fashion or a similar option.

src/core/stores/VNode.ts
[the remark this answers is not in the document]
Zinston: Agreed. But I meant the splices in the (private) methods of VNode itself.

src/core/stores/VDocument.ts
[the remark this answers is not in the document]
Zinston: This is now handled in https://github.com/odoo-dev/jabberwock/pull/33/commits/2d10ecf5145f1d3b5e3deedd2eba4574d6e1e70d where `truncate` becomes `deleteSelection`.

src/core/utils/CorePlugin.ts
Zinston: Handled in https://github.com/odoo-dev/jabberwock/pull/33/commits/27895a6e7515bc45f4920c99d68660e1ec104643#diff-a011953d0e7ac92c7b6d27420365d566R48

src/core/stores/VNode.ts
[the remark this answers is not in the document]
Zinston: @dmo-odoo We need to look into this on Tuesday.

src/plugins/DevTools/DevTools.xml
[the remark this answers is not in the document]
Zinston: It's possible if we put a helper function in the owl component.

Zinston:
@dmo-odoo I think this is as far as I'll get with rebasing. I could keep going but this might be overkill so if you think I should please just tell me, otherwise here's a reviewable/mergeable version.

src/core/utils/CorePlugin.ts
```diff
@@ -43,17 +43,27 @@ export class CorePlugin extends JWPlugin {
         // TODO: this is a stub
         const range = this.editor.vDocument.range;
         if (range.isCollapsed()) {
-            range.setStart(range.start.previous());
+            const previous = range.start.previousLeaf();
+            if (previous && range.start.index > 0) {
+                range.setStart(previous, RelativePosition.BEFORE);
+            } else if (previous) {
+                range.setStart(previous, RelativePosition.AFTER);
+            }
```
dmo-odoo: I think I would have done:
```
let previous: VNode;
let position: RelativePosition;
if (previousSibling) {
    previous = previousSibling.firstLeaf();
    position = RelativePosition.BEFORE;
} else {
    previous = range.start.previousLeaf();
    position = RelativePosition.AFTER;
}
```
Not sure that's better though... I think we have a problem here. I think your version is what we want because it is shorter and more generic, in the sense that `previousLeaf` already does the sibling check I rewrote here. However, it requires to check the `range.start.index`, which is not obvious if you don't precisely know what `previousLeaf` does. Maybe we need something like `previousAnchor` or `previousLocation` which would return a tuple `[VNode, RelativePosition]` ? Because whichever of the two above solution we choose, we are inevitably duplicating the `previous` logic in some way.

src/core/utils/CorePlugin.ts
```diff
@@ -43,17 +43,27 @@ export class CorePlugin extends JWPlugin {
...
         // TODO: this is a stub
         const range = this.editor.vDocument.range;
         if (range.isCollapsed()) {
-            range.setEnd(range.end.next());
+            const next = range.end.nextLeaf();
+            if (next && next === range.end.nextSibling()) {
+                range.setEnd(next, RelativePosition.AFTER);
+            } else if (next) {
+                range.setEnd(next, RelativePosition.BEFORE);
+            }
```
dmo-odoo: Same issue here, obviously.

src/core/stores/VRange.ts
```diff
@@ -171,17 +258,19 @@ export class VRange {
...
-            if (position === RelativePosition.AFTER && reference !== this.end) {
-                reference.after(this.start);
-            } else {
-                reference.before(this.start);
-            }
-            this._direction = null; // Invalidate range direction cache.
-            return this;
-        });
+    _setTail(reference: VNode, position = RelativePosition.BEFORE): VRange {
+        reference = reference.firstLeaf();
```
dmo-odoo: It's funny because this commit removes the withRange that was added by the previous commit, so I guess it was not truly needed ?

src/core/stores/VRange.ts
```diff
@@ -171,17 +258,19 @@ export class VRange {
...
+        if (!reference.hasChildren() && !reference.properties.atomic) {
```
dmo-odoo: Wait, the concept of atomic did not exist before, is this a hidden fix ?

src/core/utils/BasicHtmlRenderingEngine.ts
```diff
@@ -50,6 +50,12 @@ export const BasicHtmlRenderingEngine = {
         if (!node.hasChildren() && !node.properties.atomic) {
             renderedElement.appendChild(document.createElement('BR'));
         }
+
+        // If a LINE_BREAK has no next sibling, it must be rendered as two BRs
+        // in order for it to be visible.
+        if (node.type === VNodeType.LINE_BREAK && !node.nextSibling()) {
+            fragment.appendChild(document.createElement(tagName));
+        }
```
dmo-odoo: Here is another place where we should have different classes/renderers, such that this only exist in the renderer of a `LINE_BREAK` node such that the type check is not needed.

src/core/utils/Renderer.ts
```diff
@@ -106,6 +106,8 @@ export class Renderer {
      * Render the element matching the current vNode and append it.
      *
      * @param context
+     * @param __isSecondBR true if we are rendering a second BR of a pair that
```
dmo-odoo: nope

src/core/JWEditor.ts
```diff
@@ -31,6 +31,10 @@ export class JWEditor {
 
         if (!editable) {
             editable = document.createElement('jw-editable');
+            editable.style.display = 'block';
```
Zinston: This is for the other CSS commit of this PR

src/core/stores/VNode.ts
```diff
@@ -407,6 +407,30 @@ export class VNode {
             return next;
         }
     }
+    /**
```
Zinston: Fix commit title

src/core/stores/VDocument.ts
```diff
@@ -51,34 +53,54 @@ export class VDocument {
...
+    //--------------------------------------------------------------------------
+
+    /**
+     * Call a method on this VNode without ignoring the range nodes.
+     *
+     * @param method
+     */
+    static withRange<T>(method: () => T): T {
+        // Record the previous value to allow for nested calls to `withRange`.
+        const previousValue = !!withRange;
```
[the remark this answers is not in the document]
Zinston: Unnecessary indeed.

src/core/stores/VNode.ts
```diff
@@ -1,4 +1,5 @@
 import { BasicHtmlRenderingEngine, RenderingEngine } from '../utils/BasicHtmlRenderingEngine';
+import { withRange, VDocument } from './VDocument';
```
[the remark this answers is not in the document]
Zinston: While it does sound good, that would potentially make each VNode considerably heavier for the memory system. To investigate.

src/core/stores/VNode.ts
```diff
@@ -69,6 +68,15 @@ export class VNode {
...
 
+    /**
+     * Return the VNode's children.
+     */
+    get children(): VNode[] {
+        if (withRange) {
+            return this._children;
+        }
+        return this._children.filter(child => !child.isRange());
+    }
```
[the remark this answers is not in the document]
Zinston: Let's put a pin on this for a future PR.

src/core/stores/VRange.ts
Zinston: This should be in a commit to set the range in empty non-atomic nodes.

src/core/utils/Renderer.ts
```diff
@@ -113,8 +113,13 @@ export class Renderer {
         elements.forEach((element: Node): void => {
             context.parentNode.appendChild(element);
             VDocumentMap.set(context.currentVNode, element);
+            element.childNodes.forEach(child => VDocumentMap.set(context.currentVNode, child));
+        });
+        return Object.assign(context, {
+            vNode: context.currentVNode,
+            parentElement: context.parentNode,
+            lastRendered: fragment.lastChild,
```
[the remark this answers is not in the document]
Zinston: Mais...!!!?

src/core/JWEditor.ts
Zinston: Squashed both into one commit with a new CSS file.

src/core/stores/VNodes/CharNode.ts
```diff
@@ -0,0 +1,10 @@
+import { VNode, VNodeType, FormatType } from '../VNode';
```
Zinston: @dmo-odoo Here's a first working draft. What do you think of the direction it's taking?

src/core/utils/Renderer.ts
dmo-odoo: image

src/core/utils/Renderer.ts
```diff
@@ -112,8 +112,12 @@ export class Renderer {
         Array.from(fragment.childNodes).forEach((element: Node): void => {
             context.parentNode.appendChild(element);
             VDocumentMap.set(context.currentVNode, element);
+            element.childNodes.forEach(child => VDocumentMap.set(context.currentVNode, child));
+        });
+        return Object.assign(context, {
+            vNode: context.currentVNode,
+            parentElement: context.parentNode,
```
dmo-odoo: image

src/core/stores/VNodes/RangeNode.ts
```diff
@@ -0,0 +1,16 @@
+import { VNode, VNodeType } from '../VNode';
+
+export class RangeNode extends VNode {
+    properties = {
+        atomic: true,
+    };
+    constructor(tailOrHead: 'tail' | 'head') {
```
dmo-odoo: `side` ?

dmo-odoo: `edge` ?

dmo-odoo: Or `type` and the expected value is either `VNodeType.Range_TAIL` or `VNodeType.RANGE_HEAD` to avoid creating a new string bridge to these values.

src/core/stores/VNodes/SimpleElementNode.ts
```diff
@@ -0,0 +1,7 @@
+import { VNode, VNodeType } from '../VNode';
+
+export class SimpleElementNode extends VNode {
```
dmo-odoo: At this time, why not simply `ElementNode` ?

src/core/stores/VNode.ts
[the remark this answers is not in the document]
Zinston: @Gorash

src/core/stores/VNodes/CharNode.ts
```diff
@@ -0,0 +1,10 @@
...
+
+export class CharNode extends VNode {
+    properties = {
+        atomic: true,
+    };
+    constructor(char: string, format: FormatType) {
+        super(VNodeType.CHAR, '#text', char, format);
```
dmo-odoo: I suggest renaming the `value` property of `VNode` to `name` such that we could merge `originalTag` and `value` since `value` is not used for elements and `originalTag` is this useless `#text` for text nodes. The render function of element nodes would know that `name` is the `nodeName` while the render function of char nodes would know that `name`, in the context of a char node, is the character itself. Maybe this `name` property could then go into the `properties` dict ?

src/core/stores/VRange.ts
```diff
@@ -19,8 +20,8 @@ export interface VRangeDescription {
 }
 
 export class VRange {
-    readonly _tail = new VNode(VNodeType.RANGE_TAIL);
-    readonly _head = new VNode(VNodeType.RANGE_HEAD);
+    readonly _tail = new RangeNode('tail');
+    readonly _head = new RangeNode('head');
```
dmo-odoo: This conforts me in saying that we should use the VNodeType here.

src/core/utils/Parser.ts
```diff
@@ -91,12 +96,8 @@ function parseElementNode(currentContext: ParsingContext): ParsingContext {
...
         });
     } else {
-        const parsedNode: VNode = new VNode(
-            getNodeType(node),
-            node.nodeName,
-            undefined,
-            context.format[0],
-        );
+        const parsedNode: VNode = VNodeFromTag(node.nodeName);
+        Object.assign(parsedNode.format, context.format[0]);
```
dmo-odoo: This format assignation only makes sense for line break, right ?

PR #34, opened 2019-11-06
[ADD] Core: handle basic enter

Zinston:


Zinston:
@dmo-odoo Updated.

dmo-odoo:
@Zinston Looks like it's not rebased on master after our merges of last wednesday ;-)

Zinston:
> @Zinston Looks like it's not rebased on master after our merges of last wednesday ;-)

Now it is ;-)

src/core/utils/CorePlugin.ts
```diff
@@ -34,17 +35,29 @@ export class CorePlugin extends JWPlugin {
...
+     * Handle the (shift+) enter key.
+     *
+     * @param intent
+     */
+    enter(intent: KeydownIntent): void {
+        if (intent.payload.shiftKey) {
+            this.editor.vDocument.insert(new VNode(VNodeType.LINE_BREAK));
+        } else {
+            this.editor.vDocument.enter();
+        }
```
dmo-odoo: No. Enter and Shift+Enter trigger are two different intents and so two different commands. I think Shift+Enter is `insertLineBreak`, so no need for a "keyboard intent" which makes no sense in our abstraction IMO.

src/core/utils/CorePlugin.ts
```diff
@@ -34,17 +35,29 @@ export class CorePlugin extends JWPlugin {
...
      * @param intent
      */
     insert(intent: InsertIntent): void {
-        // TODO: check the intent to insert other things than text.
-        this.editor.vDocument.insertText(intent.payload.value);
+        if (typeof intent.payload.value === 'string') {
+            this.editor.vDocument.insertText(intent.payload.value);
+        } else {
+            this.editor.vDocument.insert(intent.payload.value);
+        }
```
dmo-odoo: I would rename the current `insert` into `insertText` and then have `insert(node)` and `insertText(string)`

src/core/utils/EventManager.ts
```diff
@@ -82,12 +82,18 @@ export class EventManager {
     _matchIntent(customEvent: CustomEvent): Intent {
         // TODO: this value is an implicit any!
         let payload = customEvent.detail;
-        let name;
+        let name = customEvent.type;
```
dmo-odoo: Then what is the point of https://github.com/odoo-dev/jabberwock/pull/34/commits/ae601e0948b7beeb34b3ef726552c7cefdea6bcf#diff-cb93f8d40d99945285b71140ed2bb049R133 ?

src/core/stores/VDocument.ts
```diff
@@ -25,6 +25,29 @@ export class VDocument {
     // Public
     //--------------------------------------------------------------------------
 
+    /**
+     * Handle the general expected behavior (a block split) of the enter key.
+     */
+    enter(): void {
```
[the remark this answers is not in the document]
Zinston: Shall we call it `splitParagraph` and have the insert BR intent be triggered by the EventManager?

src/core/stores/VDocument.ts
```diff
@@ -25,6 +25,29 @@ export class VDocument {
...
+            this.deleteSelection();
+        }
+        // Collect all the siblings after the first range node.
+        const nextSiblings: VNode[] = [this.range.start];
+        VDocument.withRange(() => {
+            if (this.range.start.nextSibling()) {
+                this.range.start.next(next => {
+                    nextSiblings.push(next);
+                    return !next.nextSibling();
+                });
```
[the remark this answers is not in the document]
Zinston: That seems like a good idea to add that method, but then we would want it for `next` as well... How would you call that one?

src/core/stores/VDocument.ts
```diff
@@ -25,6 +25,29 @@ export class VDocument {
...
+            }
+        });
+        // Duplicate the parent and put the siblings in it.
+        const duplicatedParent = this.range.start.parent.duplicate();
+        this.range.start.parent.after(duplicatedParent);
```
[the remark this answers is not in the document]
Zinston: Uuuuh no. This line puts the copied parent after its original. The line after that appends the siblings to it. This is in no way setting the range, it's just taking the range's parent and putting something after it.

src/core/stores/VNode.ts
```diff
@@ -78,6 +78,12 @@ export class VNode {
     render<T>(to = 'html'): T {
         return this.renderingEngines[to].render(this) as T;
     }
+    /**
+     * Return a new VNode that is a duplicate of this node.
+     */
+    duplicate(): VNode {
+        return new VNode(this.type, this.originalTag, this.value, this.format);
```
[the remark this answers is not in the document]
Zinston: Wanna call it `shallowCopy` or something?

src/core/utils/CorePlugin.ts
Zinston: Ok, makes sense. Should be handled in EventManager then?

src/core/utils/EventManager.ts
Zinston: Sorry, that's a rebase mistake.

src/core/stores/VDocument.ts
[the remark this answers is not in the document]
dmo-odoo: French people have it easy, they call it `nexts` ! 👍 🇫🇷 👀 

I thought about it this morning, and I'm not sure we would want an equivalent for `next`. The problem with `next` is that its only bound is the edge of the tree. If you would call `node.nexts(predicate)`, you'll get all the node of the *entire* tree that are after the given node and satisfy the predicate. Would it be useful ? I mean, you would have no idea where these nodes come from, just that they are somewhere in the VDOM after the given node... 

With `nextSiblings` however, you now you are bound by the parent, so you know all nodes returned by the function automatically share some cognitive context.

src/core/stores/VDocument.ts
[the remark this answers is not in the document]
dmo-odoo: Right, I read it too fast and missed the `.parent` part, my bad !

src/core/stores/VNode.ts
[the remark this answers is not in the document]
dmo-odoo: Shall we implement a `split` method on `VNode` ? We used to have it while programming on the whiteboard. Maybe it's time to implement the `VNodeSpec` I was talking about months ago. It would be strongly dependent on the PR where you transform VNode types into different classes though so maybe we don't want to wait for it ? If we don't want to wait for it, we can keep this duplicate method, and add a TODO where it is called:
`TODO: replace this `duplicate` method by `split` on VNode`

src/core/stores/VDocument.ts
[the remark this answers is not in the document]
dmo-odoo: Yes, it's how I envisioned it. The EventManager checks whether shift was pressed as well and triggers the correct intent (and thus the correct command) accordingly. Maybe we could even simplify this process later with the keymap if we create a keymap which binds `Enter` to `splitParagraph` and `SHIFT+ENTER` to `insertLineBreak`.

I'm still not 100% convinced of the name though. Is the intent of the user really to split something ? I'm not sure... I think the user just wants to insert a carriage return. It's an implement detail of the tree structure of HTML that "inserting a carriage return" into a node ends up actually splitting it. It even makes sense if you think about it as a typwriting machine (like all control characters and key terminology were inherited from):

You would have mechanically activated some formatting mode in your typewriter, let's call it "H1". You write some text, then you pull on the "carriage return and line feed" lever, you are at the beginning (carriage return) of the next line (line feed) but you didn't toggle the "H1" mode by mechanically pressing its key, so you are continuing to write an "H1". It's just a detail of HTML that it actually split the H1 at its end in this case and that you end up with two H1.

However, this analogy has an issue. The old debate of CR vs LF. Technically, CR gets you at the start of the current line, and LF gets you on the next line, so we would need "CRLF" to have the proper result. So `insertCarriageReturn` is *technically* wrong, but so are the OS platforms using only `CR` or `LF` instead of `CRLF`.

To somewhat stay close to the CRLF terminology while not sticking too much to it, maybe we could do `insertReturn` ? So we would have `insertReturn` and `insertLineBreak` but then at that point you'd be surprised that "inserting a return" ends up splitting the node, and I just went full circle at this point.

Our challenge for today is to find a name that does not include `split`, because I don't think that splitting is the intent of the end user, but still hints the developer that the node might will be split.

Right now, I have the following ideas: `insertParagraphBreak`, `breakParagraph`, `break`. I like the word "break" because it conveys the "split" idea while not forcing it. Since there is "line break", you understand that "paragraph break" is just another typing control mechanism. It just breaks at a different level, which is what we want IMO. However, `break` is too similar to "line break" so I think we have to keep the word "paragraph" to avoid any confusion. Of my early three propositions, only two remains:
- `insertParagraphBreak`:
    - PROS: obvious counterpart of `insertLineBreak. Exactly the intent of the user.
    - CONS: "insert" dilutes the link between "break" and "split". You again end up splitting something by inserting something into it. Isn't that weird ? At first glance I'd say yes. At second glance, I'm more "well, I inserted a *break*, I should not really be surprised that I actually *broke* it, should I ?"
- `breakParagraph`:
    - PROS: more direct, I'm obviously breaking this paragraph.
    - CONS: But again, is this what I wanted to do ? If I'm at the end of a paragraph, do I want to "break" it ? If I'm in a cell of a table, I press enter, does it "break" something ? I'm scared of actually *breaking*, in the sense of "destroying", some stuff by calling this command in my code.

Looking at the `inputType` attribute of `InputEvent`, they have `insertLineBreak` which is "insert a line break" and `insertParagraph` which is "insert a paragraph break". Notice how `insertParagraph` is missing the word "break" but they still had to use it in the description because it makes no sense without it ? I definitely chose my challenger for today and it is `insertParagraphBreak` ! 🥇

src/core/utils/CorePlugin.ts
dmo-odoo: Yes, see my above comment about the keymap.

src/core/utils/EventManager.ts
dmo-odoo: 👍

src/core/stores/VDocument.ts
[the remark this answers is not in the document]
Zinston: I'd also be in favor of either `insertParagraphBreak` or something to do with the carriage return like `insertCarriageReturn` or `returnCarriage` or `crlf`.

Zinston:
@dmo-odoo Ready. There were three more fixes I wanted to move here from master-coretests-age, but there were apparently a lot of things to discuss regarding those (https://github.com/odoo-dev/jabberwock/pull/40/commits/038aa5785a2f96bcfbf007f3b469f90707554cce, https://github.com/odoo-dev/jabberwock/pull/40/commits/8ef1f7372a5249e17e81c0a8a1761b04cd6f4a21 and https://github.com/odoo-dev/jabberwock/pull/40/commits/c70d6541794a2fb3e26cdb80f2b01cf748393db5).

src/core/stores/VNode.ts
```diff
@@ -483,6 +483,34 @@ export class VNode {
...
+     *
+     * @param [predicate]
+     * @param [__current] recursive collector
+     */
+    previousSiblings(predicate?: Predicate, __current: VNode[] = []): VNode[] {
+        if (!predicate || predicate(this)) {
+            __current.push(this);
+        }
+        const previous = this.previousSibling();
+        return previous ? previous.previousSiblings(predicate, __current) : __current;
```
dmo-odoo: I don't really like using these obnoxious `__current` parameters. What do you think of this ?
```typescript
const prevSiblings: VNode[] = [];
let sibling = this.previousSibling();
while (sibling && (!predicate || predicate(sibling)) {
    prevSiblings.push(sibling);
    sibling = sibling.previousSibling();
}
return prevSiblings;
```

If we later want to implement this as an iterator, it would even be shorter:
```
typescript
let sibling = this.previousSibling();
while (sibling && (!predicate || predicate(sibling)) {
    yield sibling;
    sibling = sibling.previousSibling();
}
```

src/core/stores/VNode.ts
```diff
@@ -483,6 +483,34 @@ export class VNode {
...
+     *
+     * @param [predicate]
+     * @param [__current] recursive collector
+     */
+    nextSiblings(predicate?: Predicate, __current: VNode[] = []): VNode[] {
+        if (!predicate || predicate(this)) {
+            __current.push(this);
+        }
+        const next = this.nextSibling();
+        return next ? next.nextSiblings(predicate, __current) : __current;
```
dmo-odoo: Same suggestion here, obviously.

src/core/utils/Parser.ts
```diff
@@ -89,6 +89,19 @@ function parseElementNode(currentContext: ParsingContext): ParsingContext {
     const context = { ...currentContext };
 
     const node = context.node;
+    // A <br/> with no siblings is there only to make its parent visible.
```
Zinston: @dmo-odoo This is basically your suggestion from your review of https://github.com/odoo-dev/jabberwock/pull/40/commits/038aa5785a2f96bcfbf007f3b469f90707554cce. I did however not include the part that accounts for textual range nodes for commit scope reasons.

PR #35, opened 2019-11-07
[ADD] Core: handle paragraph formatting

Zinston:
Let examples speak:
```xml
<paragraph>te◆xt</paragraph>
```
=> CTRL+ALT+DIGIT1 =>
```typescript
changeType(VNodeType.HEADING1)
```
=>
```xml
<heading1>▶text◀</heading1>
```

and

```xml
<heading1>te▶xt</heading1>
<heading2>te◀xt</heading2>
```
=> CTRL+ALT+DIGIT0 =>
```typescript
changeType(VNodeType.PARAGRAPH)
```
=>
```xml
<paragraph>te▶xt</paragraph>
<paragraph>te◀xt</paragraph>
```

### DIFF
--- a/src/core/stores/VDocument.ts
+++ b/src/core/stores/VDocument.ts
@@ -49,6 +49,32 @@ export class VDocument {
         nextSiblings.forEach(sibling => duplicatedParent.append(sibling));
     }
     /**
+     * Change the selection's paragraph formatting. If there is no selection,
+     * first select the parent of the range nodes.
+     *
+     * Examples:
+     *
+     * - `<paragraph>te◆xt</paragraph>` =>
+     *   `changeType(VNodeType.HEADING1)` =>
+     *   `<heading1>▶text◀</heading1>`
+     *
+     * - `<heading1>te▶xt</heading1><heading2>te◀xt</heading2>`=>
+     *   `changeType(VNodeType.PARAGRAPH)` =>
+     *   `<paragraph>te▶xt</paragraph><paragraph>te◀xt</paragraph>`
+     *
+     * @param type
+     */
+    formatParagraph(type: VNodeType): void {
+        if (this.range.isCollapsed()) {
+            this.range.select(this.range.start.parent, this.range.end.parent);
+        }
+        new Set(this.range.selectedLeaves.map(node => node.parent)).forEach(parent => {
+            // TODO: this suffices for the moment, but it will not be enough
+            // anymore once we introduce the VNode class extensions.
+            parent.type = type;
+        });
+    }
+    /**
      * Insert something at range.
      *
      * @param node
--- a/src/core/stores/VRange.ts
+++ b/src/core/stores/VRange.ts
@@ -98,6 +98,13 @@ export class VRange {
         });
         return selectedNodes;
     }
+    /**
+     * Return a list of all the leaves implied in the selection between the
+     * first range node to the last (non-included).
+     */
+    get selectedLeaves(): VNode[] {
+        return this.selectedNodes.filter(node => !node.children.length);
+    }
 
     //--------------------------------------------------------------------------
     // Private
--- a/src/core/utils/CorePlugin.ts
+++ b/src/core/utils/CorePlugin.ts
@@ -1,13 +1,20 @@
 import { JWPlugin } from '../JWPlugin';
 import JWEditor from '../JWEditor';
 import { RelativePosition, VRangeDescription, Direction } from '../stores/VRange';
-import { InsertIntent, RangeIntent, FormatIntent, KeydownIntent } from '../types/Intents';
+import {
+    InsertIntent,
+    RangeIntent,
+    FormatIntent,
+    KeydownIntent,
+    FormatParagraphIntent,
+} from '../types/Intents';
 import { VNode, VNodeType } from '../stores/VNode';
 
 export class CorePlugin extends JWPlugin {
     editor: JWEditor;
     handlers = {
         intents: {
+            formatParagraph: 'formatParagraph',
             enter: 'enter',
             insert: 'insert',
             deleteBackward: 'deleteBackward',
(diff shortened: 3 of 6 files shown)


Zinston:
@dmo-odoo Updated.

src/core/stores/VDocument.ts
```diff
@@ -38,6 +38,34 @@ export class VDocument {
...
+     *   `changeType(VNodeType.PARAGRAPH)` =>
+     *   `<paragraph>te▶xt</paragraph><paragraph>te◀xt</paragraph>`
+     *
+     * @param type
+     */
+    formatParagraph(type: VNodeType): void {
+        if (this.range.isCollapsed()) {
+            this.range.select(this.range.start.parent, this.range.end.parent);
+        }
+        new Set(this.range.selectedLeaves.map(node => node.parent)).forEach(parent => {
```
dmo-odoo: What is most costly, assigning the same property of the same object multiple times, or creating and handling a set to call it only once ? Actually, I'm not sure I understand what is going on here...

src/core/utils/CorePlugin.ts
```diff
@@ -33,6 +35,15 @@ export class CorePlugin extends JWPlugin {
...
     //--------------------------------------------------------------------------
 
+    /**
+     * Change the type of the selection. If there is no selection, change the
+     * type of the range nodes' parent.
+     *
+     * @param intent
+     */
+    formatParagraph(intent: FormatParagraphIntent): void {
+        this.editor.vDocument.formatParagraph(intent.payload.value);
```
dmo-odoo: Why on `VDocument` ?

src/core/utils/EventManager.ts
```diff
@@ -98,7 +129,19 @@ export class EventManager {
...
+                    // payload.key
+                    if (Object.keys(ctrlAltShortcuts).includes(payload.key)) {
+                        name = ctrlAltShortcuts[payload.key].name;
+                        payload.value = ctrlAltShortcuts[payload.key].value;
+                    } else {
+                        return;
+                    }
+                } else {
+                    return;
+                }
```
dmo-odoo: I think this is not rebased on master, as this part is missing the applyFormat part added by nby. Both this and the applyFormat shortcuts will need to be converted to use the keymap.

src/core/stores/VRange.ts
```diff
@@ -96,6 +96,13 @@ export class VRange {
         });
         return selectedNodes;
     }
+    /**
+     * Return a list of all the leaves implied in the selection between the
+     * first range node to the last (non-included).
+     */
+    get selectedLeaves(): VNode[] {
+        return this.selectedNodes.filter(node => !node.children.length);
```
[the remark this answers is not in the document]
Zinston: That will be sweet...

src/core/utils/CorePlugin.ts
Zinston: That was the design as discussed: CorePlugin makes the link with VDocument - VDocument handles the changes on itself. I'm aware this will change with the commands redesign and the probable elimination of CorePlugin but in the meantime this has been the design.

src/core/stores/VDocument.ts
Zinston: Good question actually, I don't know.
As to what's going on:
- Taking each selected node
- Taking their parents
- Putting that in a set (_.uniq)
- Applying the new types on them

src/core/utils/EventManager.ts
Zinston: Indeed.

src/core/stores/VDocument.ts
```diff
@@ -38,6 +38,34 @@ export class VDocument {
...
+     *
+     * - `<heading1>te▶xt</heading1><heading2>te◀xt</heading2>`=>
+     *   `changeType(VNodeType.PARAGRAPH)` =>
+     *   `<paragraph>te▶xt</paragraph><paragraph>te◀xt</paragraph>`
+     *
+     * @param type
+     */
+    formatParagraph(type: VNodeType): void {
+        if (this.range.isCollapsed()) {
+            this.range.select(this.range.start.parent, this.range.end.parent);
```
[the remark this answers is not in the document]
Zinston: On VRange then? Like `selectParent` or `selectUp` ? That could be useful. Maybe wait to see if we need it a second time?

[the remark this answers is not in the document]
dmo-odoo: Maybe that might be the default behavior of `select` ? You'd do `this.range.select()` and that would do that. We already have a way to express "select nothing" with `this.range.collapse()`, so it might make sense that `select` *always* select something, and this would be the default behavior if you don't specify *what* to select.

src/core/utils/CorePlugin.ts
dmo-odoo: Well yes, for very "basic" operations like `insert`, but even `deleteForward`/`deleteBackward` are in `CorePlugin` so why would this one be in `VDocument` ? (To be 100% fair, I just realized I merged `applyFormat` in `VDocument`, which is wrong as well IMO since it's exactly the same use case as this one.)

dmo-odoo:
@Zinston I decided to make the changes I suggest on this branch by myself while I was waiting for nby this morning. I'm closing this PR in favor of https://github.com/odoo-dev/jabberwock/pull/43. It's still open for discussions, I just didn't want to force push this branch and I didn't want to keep two PRs for the same thing either.

Besides, I had to change the shortcuts to use 1, 2, 3 etc rather than &, é, " otherwise they would not trigger on Mac. I think this is a non-negligeable issue that will need to be tackled in the KeyMap PR. Maybe this PR (and thus https://github.com/odoo-dev/jabberwock/pull/43) should wait for KeyMap one ?

src/core/utils/CorePlugin.ts
Zinston: To me putting the delete functions in CorePlugin was precisely the mistake. Since it was done while I was on vacation I assumed that was just an oversight 😉 But anyways with the rethinking of Intents and Commands, I believe that pattern will change and maybe CorePlugin will even disappear...

dmo-odoo: Well, I think we changed our minds a hundred times about CorePlugin so I can't really blame anyone for this. However, the commands PR will have no impact on this issue. This is more of a "separation of concerns" issue IMO. Like formatting in bold might only be the concern of the "Bold" plugin, and changing the paragraph might be the concern of the "Paragraph" plugin. I think we need to discuss this sooner rather than later lest we introduce more confusion on the matter.

PR #37, opened 2019-11-13
[ADD] Core: handle keymaps

Zinston:
To add a custom keymap:
```typescript
const keyMap: Map<string, IntentInformation> = new Map();
// add shortcuts to the map...
// eg:
keyMap.set('Ctrl+Alt+K', {
    name: 'insertText',
    arguments: { value: 'Do cats eat bats? Do bats eat cats?' },
});
keyMap.set('Ctrl+Alt+V', {
    name: 'insertText',
    arguments: { value: '⚔' },
});

// before editor start:
editor.loadConfig({ keyMap: keyMap });
```

Or to add or replace shortcuts on the existing keyMap, replace everything above with:
```typescript
editor.keyMap.set({
    'Ctrl+Alt+K': {
        name: 'insertText',
        arguments: { value: 'Do cats eat bats? Do bats eat cats?' },
    },
    'Ctrl+Alt+V': {
        name: 'insertText',
        arguments: { value: '⚔' },
    },
});
```

Note: The META key is not technically supported - we consider it is equivalent to the CTRL key.

### DIFF
--- a/packages/core/src/EventManager.ts
+++ b/packages/core/src/EventManager.ts
@@ -1,21 +1,28 @@
 import { EventNormalizer, DomRangeDescription } from './EventNormalizer';
 import { LineBreakNode } from './VNodes/LineBreakNode';
 import JWEditor from './JWEditor';
+import { CommandConfig, KeyMapping } from './KeyMapping';
 
 interface SetRangeParams {
     domRange: DomRangeDescription;
 }
 
+interface EventManagerOptions {
+    keyMap?: KeyMapping;
+}
+
 export class EventManager {
     editor: JWEditor;
     eventNormalizer: EventNormalizer;
+    keyMap: KeyMapping;
 
-    constructor(editor: JWEditor) {
+    constructor(editor: JWEditor, options: EventManagerOptions = {}) {
         this.editor = editor;
         this.eventNormalizer = new EventNormalizer(
             editor.editable,
             this._onNormalizedEvent.bind(this),
         );
+        this.keyMap = options.keyMap;
     }
 
     //--------------------------------------------------------------------------
@@ -27,6 +34,7 @@ export class EventManager {
      */
     _onNormalizedEvent(customEvent: CustomEvent): void {
         const payload = customEvent.detail;
+        let shortcutMatch: CommandConfig;
         switch (customEvent.type) {
             case 'enter':
                 if (customEvent.detail.shiftKey) {
@@ -45,32 +53,15 @@ export class EventManager {
                 // TODO: keydown should be matched with existing shortcuts. If
                 // it matches an command shortcut, trigger the corresponding
                 // command, otherwise do not trigger any command.
-                if (
-                    payload.ctrlKey &&
-                    !payload.altKey &&
-                    !payload.shiftKey &&
-                    !payload.metaKey &&
-                    payload.key === 'b'
-                ) {
-                    return this.editor.execCommand('applyFormat', { format: 'bold' });
-                } else if (
-                    payload.ctrlKey &&
-                    !payload.altKey &&
-                    !payload.shiftKey &&
-                    !payload.metaKey &&
-                    payload.key === 'i'
-                ) {
-                    return this.editor.execCommand('applyFormat', { format: 'italic' });
-                } else if (
-                    payload.ctrlKey &&
-                    !payload.altKey &&
-                    !payload.shiftKey &&
-                    !payload.metaKey &&
-                    payload.key === 'u'
-                ) {
-                    return this.editor.execCommand('applyFormat', { format: 'underline' });
+
+                // TODO: use payload.code (Digit[0-6]) instead of
+                // payload.key
+                if (this.keyMap) {
+                    shortcutMatch = this.keyMap.fromKeypress(payload);
+                    if (shortcutMatch) {
+                        this.editor.execCommand(shortcutMatch.name, shortcutMatch.arguments);
+                    }
                 }
-                // TODO: keydown should be matched with existing shortcuts.
                 return;
             }
             default:
--- a/packages/core/src/EventNormalizer.ts
+++ b/packages/core/src/EventNormalizer.ts
@@ -877,8 +877,8 @@ export class EventNormalizer {
         compiledEvent.type = compiledEvent.type || ev.type;
         compiledEvent.key = ev.key;
         compiledEvent.altKey = ev.altKey;
-        compiledEvent.ctrlKey = ev.ctrlKey;
-        compiledEvent.metaKey = ev.metaKey;
+        // Meta key not supported: we consider it equivalent to Ctrl key.
+        compiledEvent.ctrlKey = ev.ctrlKey || ev.metaKey;
         compiledEvent.shiftKey = ev.shiftKey;
     }
     /**
--- a/packages/core/src/KeyMapping.ts
+++ b/packages/core/src/KeyMapping.ts
@@ -1,0 +1,78 @@
+import { defaultKeyMap } from './DefaultKeyMap';
+
+export interface Keypress {
+    key: string;
+    ctrlKey?: boolean;
+    altKey?: boolean;
+    shiftKey?: boolean;
+}
+export interface CommandConfig {
+    name: string;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    arguments: Record<string, any>;
+}
+export type ShortcutIdentifier = string; // format: /[Ctrl+][Alt+][Shift+]key/
+export type KeyMap = Map<ShortcutIdentifier, CommandConfig>;
+
+export class KeyMapping {
+    keyMap: KeyMap = new Map();
+    constructor(customKeyMap?: KeyMap) {
+        const map = customKeyMap || defaultKeyMap;
+        map.forEach((value, key) => this.set(key, value));
+    }
+
+    //--------------------------------------------------------------------------
+    // Public
+    //--------------------------------------------------------------------------
+
+    /**
+     * Return the intent information corresponding to the given keypress
+     * information, if any.
+     *
+     * @param keypress
+     */
+    fromKeypress(keypress: Keypress): CommandConfig {
+        return this.keyMap.get(this._keypressToIdentifier(keypress));
+    }
+    /**
+     * Set one or more new shortcuts in the map.
+     *
+     * @param shortcut
+     * @param commandConfig
+     */
+    set(shortcuts: Record<ShortcutIdentifier, CommandConfig>): void;
+    set(shortcut: ShortcutIdentifier, commandConfig?: CommandConfig): void;
+    set(
+        shortcut: Record<ShortcutIdentifier, CommandConfig> | ShortcutIdentifier,
+        commandConfig?: CommandConfig,
+    ): void {
+        if (typeof shortcut === 'object') {
+            Object.keys(shortcut).forEach(key => this.set(key, shortcut[key]));
+        } else {
+            this.keyMap.set(shortcut.toUpperCase(), commandConfig);
+        }
+    }
+
+    //--------------------------------------------------------------------------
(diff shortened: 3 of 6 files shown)


Zinston:
@dmo-odoo Ready

src/core/utils/DefaultKeyMap.ts
```diff
@@ -0,0 +1,53 @@
...
+            format: 'bold',
+        },
+    },
+    'ctrl+i': {
+        name: 'applyFormat',
+        arguments: {
+            format: 'italic',
+        },
+    },
+    'ctrl+u': {
```
dmo-odoo: I just checked in Google Drive on MacOS and in Sublime Text on Linux and both use the following typography: `Modifier+K`

That is the first letter of the modifier is capitalized and the actual key is capitalized as well, even though you're not supposed to type the capitalzed letter. So `Ctrl+K` triggers on `Ctrl+k`, otherwise it would require `Ctrl+Shift+k` to truly type `Ctrl+K`. In fact, I just realized the letters are capitalzed on the keyboards, so pressing the `K` key actually enters `k`, you need `Shift+K` to truly insert `K`, funny.

I still have a personal preference for all capitalized shortcuts like `CTRL+SHIFT+K` though, because somehow seeing it that way makes it more obvious in my brain that each of them are a key of their own, as they are like a "constant" of the keyboard if you like. The cognitive links in my brain activates faster when I see them all capitalized, but maybe it's just me. Both `Ctrl+Shift+K` and `CTRL+SHIFT+K` are acceptable to me, but not `ctrl+shift+k`.

If you do not like "all caps" then I suggest to use `Ctrl+Shift+K` in plugin definitions, and to `toUpperCase` it for the system rather than 'toLowerCase' like it is now. The system itself won't really care anyway... :p

src/core/utils/EventManager.ts
```diff
@@ -92,35 +96,29 @@ export class EventManager {
...
-                    !payload.metaKey &&
-                    payload.key === 'u'
-                ) {
-                    name = 'applyFormat';
-                    payload = { format: 'underline' };
-                } else {
-                    return;
+
+                // TODO: use payload.code (Digit[0-6]) instead of
+                // payload.key
```
dmo-odoo: We still need to discuss this if I remember correctly.

src/core/utils/EventManager.ts
```diff
@@ -92,35 +96,29 @@ export class EventManager {
...
+                intentInformation = this.keyMap.fromKeypress({
```
dmo-odoo: I suggest maybe `matchShortcut` or `matchHotkey` depending on how you chose to name the identifier earlier ?

src/core/utils/EventManager.ts
```diff
@@ -92,35 +96,29 @@ export class EventManager {
...
+                    key: payload.key,
+                    modifiers: {
+                        ctrlKey: payload.ctrlKey,
+                        altKey: payload.altKey,
+                        shiftKey: payload.shiftKey,
+                    },
+                });
```
dmo-odoo: Why create a new interface for something that is standard and that obviously forces us to convert it from a given format to another? I suggest changing the argument type of this function in such way that the payload can be given to it directly. So basically:
```
{
    key: string;
    ctrlKey: boolean;
    altKey: boolean;
    shiftKey: boolean;
}
```

src/core/utils/EventNormalizer.ts
```diff
@@ -873,8 +873,8 @@ export class EventNormalizer {
         compiledEvent.type = compiledEvent.type || ev.type;
         compiledEvent.key = ev.key;
         compiledEvent.altKey = ev.altKey;
-        compiledEvent.ctrlKey = ev.ctrlKey;
-        compiledEvent.metaKey = ev.metaKey;
+        // Meta key not supported: we consider it equivalent to Ctrl key.
```
dmo-odoo: Oh sorry, I meant the *manager* should consider it the same for the purpose of our application, not the *normalizer* which is supposed to be ignorant of what we do with its signals ! ;-) (And besides, I'd do this only on Mac as this is very specific to MacOS.)

src/core/utils/KeyMapping.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export interface KeyModifiers {
+    ctrlKey?: boolean;
+    altKey?: boolean;
+    shiftKey?: boolean;
+}
+export interface Keypress {
+    key: string;
+    modifiers: KeyModifiers;
+}
```
dmo-odoo: To echo the things said before, I would remove the `modifiers` intermediary and just have the modifiers booleans directly on this interface.

src/core/utils/KeyMapping.ts
```diff
@@ -0,0 +1,62 @@
...
+            identifier += 'ctrl+';
+        }
+        if (keypress.modifiers.altKey) {
+            identifier += 'alt+';
+        }
+        if (keypress.modifiers.shiftKey) {
+            identifier += 'shift+';
+        }
+        return identifier + keypress.key.toLowerCase();
+    }
```
dmo-odoo: The capitalization discussion will have repercussions here, obviously.

src/core/utils/KeyMapping.ts
```diff
@@ -0,0 +1,62 @@
...
+export interface IntentInformation {
+    name: string;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    arguments: Array<any> | Record<string, any>;
+}
+export type ShortcutIdentifier = string; // format: /[ctrl+][alt+][shift+]key/
+export type KeyMap = Map<ShortcutIdentifier, IntentInformation>;
+
+export class KeyMapping {
+    keyMap: KeyMap;
```
dmo-odoo: Why the intermediate type ? `keyMap` is a `Map<ShortcutIdentifier, IntentInformation>`.

src/core/utils/KeyMapping.ts
```diff
@@ -0,0 +1,62 @@
...
+export interface Keypress {
+    key: string;
+    modifiers: KeyModifiers;
+}
+export interface IntentInformation {
+    name: string;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    arguments: Array<any> | Record<string, any>;
+}
+export type ShortcutIdentifier = string; // format: /[ctrl+][alt+][shift+]key/
```
dmo-odoo: Oh, the type alias I referred to earlier in this review actually exists ! :D

src/core/utils/KeyMapping.ts
```diff
@@ -0,0 +1,62 @@
...
+    shiftKey?: boolean;
+}
+export interface Keypress {
+    key: string;
+    modifiers: KeyModifiers;
+}
+export interface IntentInformation {
+    name: string;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    arguments: Array<any> | Record<string, any>;
```
dmo-odoo: Nicolas' PR will soon replace Intents by Commands. I guess this will become `CommandConfig` or something like that when altered in his branch ? I'm thinking about this because my brain can't seem to get what `IntentInformation` is supposed to mean but I'm not sure it makes a lot of sense to spend time into finding a better name while Intents will soon ceast to exist in their current form.

src/core/utils/EventManager.ts
```diff
@@ -92,35 +96,29 @@ export class EventManager {
...
+                if (intentInformation) {
+                    name = intentInformation.name;
+                    if (Array.isArray(intentInformation.arguments)) {
+                        payload.value =
+                            (intentInformation.arguments.length === 1 &&
+                                intentInformation.arguments[0]) ||
+                            intentInformation.arguments;
```
dmo-odoo: Are you serious ? 🗡

src/core/JWEditor.ts
```diff
@@ -8,11 +8,12 @@ import { Renderer } from './utils/Renderer';
 import { OwlUI } from '../../src/ui/OwlUI';
 import { DevTools } from '../../src/plugins/DevTools/DevTools';
 import '../../src/plugins/DevTools/DevTools.css';
-import { FormatType } from './stores/VNode';
+import { KeyMapping, IntentInformation } from './utils/KeyMapping';
 
 export interface JWEditorConfig {
     debug?: boolean;
     theme?: string;
+    keyMap?: Map<string, IntentInformation>;
```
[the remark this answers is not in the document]
Zinston: Thought I had done that... Weird! I guess I removed it at one point for some reason...

src/core/utils/DefaultKeyMap.ts
Zinston: Yeah I tried a few different things and the lowercase felt "lightest" but that may have been cause I was starting to come down with the flu and felt uppercase letters were aggressing me 😅 Using whatever is standard definitely makes sense.

src/core/utils/EventManager.ts
Zinston: Well... I gotta admit that it's... Not great.

src/core/utils/EventNormalizer.ts
Zinston: Problem is, when doing that, it seems the EventManager doesn't receive the right event, because somehow the normalizer has modified it. Can that be?

src/core/utils/KeyMapping.ts
Zinston: Ah ! There it was!

src/core/utils/KeyMapping.ts
Zinston: Anticipating for custom keymaps, so you can import the type.

Zinston:
@dmo-odoo Rebased and applied comments. Now we need to discuss the comments that I didn't address.

packages/core/src/EventManager.ts
```diff
@@ -28,6 +35,7 @@ export class EventManager {
      */
     _onNormalizedEvent(customEvent: CustomEvent): void {
         const payload = customEvent.detail;
```
dmo-odoo: Damn it, this is (obviously) not updated according to the `EventNormalizer` branch ! (I just realized that's what was actually blocking this PR.)

packages/core/src/EventNormalizer.ts
```diff
@@ -877,8 +877,8 @@ export class EventNormalizer {
         compiledEvent.type = compiledEvent.type || ev.type;
         compiledEvent.key = ev.key;
         compiledEvent.altKey = ev.altKey;
-        compiledEvent.ctrlKey = ev.ctrlKey;
-        compiledEvent.metaKey = ev.metaKey;
+        // Meta key not supported: we consider it equivalent to Ctrl key.
+        compiledEvent.ctrlKey = ev.ctrlKey || ev.metaKey;
```
dmo-odoo: We talked about this the day before I departed, but I can't remember what we decided.

packages/core/src/KeyMapping.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+export interface CommandConfig {
+    name: string;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    arguments: Record<string, any>;
+}
+export type ShortcutIdentifier = string; // format: /[Ctrl+][Alt+][Shift+]key/
+export type KeyMap = Map<ShortcutIdentifier, CommandConfig>;
+
+export class KeyMapping {
```
dmo-odoo: Do we really need this class ? Considering the only component using the keymap is the event manager, I would just give a `Map<ShortcutIdentifier, CommandConfig>` to the event manager, give it a `matchShortcut` and `setShortcut` functions and be done with it, to reduce intermediary concepts.

dmo-odoo:
I mark this PR as "blocked" because the new normalizer PR must come first and this will have to be rebased on master afterwards.

dmo-odoo:
I'm closing this in favor of https://github.com/odoo-dev/jabberwock/pull/79

PR #41, opened 2019-11-18
[ADD] Utils: add predicates library

dmo-odoo:
To avoid having too much isWhatever functions on the VNode class and
to be able to use them as traversal predicates directly, without having
to use a useless arrow function like node => node.isWhatever().

Note: I had to extract `VNodeType` out of `VNode` because`VNode` depended on `Predicates` to implement its method but `Predicates` depended on `VNode` because it needed `VNodeType`.

### DIFF
--- a/src/core/stores/VNode.ts
+++ b/src/core/stores/VNode.ts
@@ -1,7 +1,6 @@
 import { BasicHtmlRenderingEngine, RenderingEngine } from '../utils/BasicHtmlRenderingEngine';
 import { withRange, VDocument } from './VDocument';
-
-export type Predicate = (node: VNode) => boolean;
+import { Predicate, isRange, isLeaf, not } from '../utils/Predicates';
 
 export enum VNodeType {
     ROOT = 'ROOT',
@@ -17,7 +16,6 @@ export enum VNodeType {
     CHAR = 'CHAR',
     LINE_BREAK = 'LINE_BREAK',
 }
-export const RangeTypes = [VNodeType.RANGE_TAIL, VNodeType.RANGE_HEAD];
 
 export interface VNodeProperties {
     atomic: boolean;
@@ -90,7 +88,7 @@ export class VNode {
         if (withRange) {
             return this._children;
         }
-        return this._children.filter(child => !child.isRange());
+        return this._children.filter(not(isRange));
     }
     /**
      * Return the length of this VNode.
@@ -190,12 +188,6 @@ export class VNode {
     isAfter(vNode: VNode): boolean {
         return vNode.isBefore(this);
     }
-    /**
-     * Return true if this VNode is a range node.
-     */
-    isRange(): boolean {
-        return RangeTypes.includes(this.type);
-    }
 
     //--------------------------------------------------------------------------
     // Browsing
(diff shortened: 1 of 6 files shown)


src/core/stores/VNode.ts
```diff
@@ -90,7 +76,7 @@ export class VNode {
         if (withRange) {
             return this._children;
         }
-        return this._children.filter(child => !child.isRange());
+        return this._children.filter(child => !isRange(child));
```
Zinston: `filter(not(isRange))`

src/core/stores/VNodeType.ts
```diff
@@ -0,0 +1,14 @@
+export enum VNodeType {
```
Zinston: Why not in `/typings` ?

src/core/utils/Predicates.ts
```diff
@@ -0,0 +1,15 @@
+import { VNode } from '../stores/VNode';
+import { VNodeType } from '../stores/VNodeType';
+
+const RangeTypes = [VNodeType.RANGE_TAIL, VNodeType.RANGE_HEAD];
+export const isRange = (node: VNode): boolean => {
```
Zinston: Docstrings ?

src/core/utils/Renderer.ts
```diff
@@ -53,12 +54,12 @@ export class Renderer {
...
      */
     _isSameTextNode(a: VNode, b: VNode): boolean {
-        if (a.isRange() || b.isRange()) {
+        if (isRange(a) || isRange(b)) {
             // A Range node is always considered to be part of the same text
             // node as another node in the sense that the text node must not
             // be broken up just because it contains the range.
             return true;
-        } else if (a.type !== VNodeType.CHAR || b.type !== VNodeType.CHAR) {
+        } else if (!isChar(a) || !isChar(b)) {
```
Zinston: :thumbsup:

src/core/utils/Predicates.ts
```diff
@@ -0,0 +1,21 @@
+import { VNode } from '../stores/VNode';
+import { VNodeType } from '../stores/VNodeType';
+
+export type Predicate = (node: VNode) => boolean;
```
Zinston: Does it really need to be exported?

dmo-odoo: It's used in VNode: https://github.com/odoo-dev/jabberwock/pull/41/files#diff-a8695298aff4077e84b4272d035925e1R3

src/core/utils/Predicates.ts
Goaman: I would write it with the `function` keyword otherwise the function is anonymous and it makes it more complicated to debug. 
```typescript
export function isRange(node: VNode): boolean => {
//...
}
```

src/core/stores/VNodeType.ts
Goaman: If find it strange to define types in differents ways:
- in the same file,
- in a separate file with the word type,
- in a separate file under the folder types,

Why not keep the types in the same file? It is directly related to VNode, it should reside in VNode.

dmo-odoo: From the PR description:

> Note: I had to extract VNodeType out of VNode becauseVNode depended on Predicates to implement its method but Predicates depended on VNode because it needed VNodeType.

I am open to suggestions about how to solve this issue without exporting VNodeType in a different file.

PR #48, opened 2019-11-28
[IMP] Core: add ability to parameterize the range for the commands

Goaman:
Before this commit, it was impossible to use a specific range when executing a command. The command where always using `vDocument.range`.

We want to allow the parameterization of the range for future features.

### DIFF
--- a/packages/core/src/CorePlugin.ts
+++ b/packages/core/src/CorePlugin.ts
@@ -1,18 +1,27 @@
 import { JWPlugin } from './JWPlugin';
 import JWEditor from './JWEditor';
-import { VRangeDescription, Direction, RelativePosition } from './VRange';
+import { VRangeDescription, Direction, RelativePosition, RangePosition } from './VRange';
 import { VNode } from './VNode';
 
-export interface InsertParams {
+export interface RangeFromTo {
+    start: VNode;
+    startPosition?: RangePosition;
+    end?: VNode;
+    endPosition?: RangePosition;
+}
+export interface OptionalRangeParams {
+    range?: RangeFromTo;
+}
+export interface InsertParams extends OptionalRangeParams {
     value: VNode;
 }
-export interface InsertTextParams {
+export interface InsertTextParams extends OptionalRangeParams {
     value: string;
 }
 export interface VRangeParams {
     vRange: VRangeDescription;
 }
-export interface FormatParams {
+export interface FormatParams extends OptionalRangeParams {
     format: 'bold' | 'italic' | 'underline';
 }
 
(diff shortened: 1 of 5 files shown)


packages/core/src/VDocument.ts
```diff
@@ -21,28 +22,82 @@ export class VDocument {
...
+            const fakeRange = new VRange(Direction.FORWARD);
+            if (rangeFromTo.startPosition && rangeFromTo.startPosition === RangePosition.AFTER) {
+                rangeFromTo.start.after(fakeRange.start);
+            } else {
+                rangeFromTo.start.before(fakeRange.start);
+            }
+            if (rangeFromTo.endPosition && rangeFromTo.endPosition === RangePosition.AFTER) {
+                rangeFromTo.end.after(fakeRange.end);
+            } else {
+                rangeFromTo.end.before(fakeRange.end);
```
dmo-odoo: Can't you just use `fakeRange.select` ?

packages/core/test/VDocument.test.ts
```diff
@@ -1,9 +1,169 @@
...
+                    testEditor({
+                        contentBefore: 'ab[]',
+                        stepFunction: async (editor: JWEditor) => {
+                            const aNode = editor.vDocument.root.next(node => node.value === 'a');
+                            const rangeParams: RangeFromTo = {
+                                start: aNode,
+                            };
+                            await editor.vDocument.withCustomRange(rangeParams, range => {
+                                editor.vDocument.insertText('c', range);
+                                editor.renderer.render(editor.vDocument, editor.editable);
```
dmo-odoo: Now that we have `execCommand`, I see no reason to use any other way of executing comands, especially considering `execCommand` includes the call to `render`.

packages/core/src/CorePlugin.ts
```diff
@@ -1,18 +1,27 @@
...
+import { VRangeDescription, Direction, RelativePosition, RangePosition } from './VRange';
 import { VNode } from './VNode';
 
-export interface InsertParams {
+export interface RangeFromTo {
+    start: VNode;
+    startPosition?: RangePosition;
+    end?: VNode;
+    endPosition?: RangePosition;
+}
```
dmo-odoo: We already have an interface that is very similar to this one, except that it has `direction` as well. Couldn't we just put `direction` as optional ? Check if there are parts of the code where we make the assumption that `direction` is always there.

packages/core/test/CorePlugin.test.ts
```diff
@@ -0,0 +1,51 @@
...
+                        const bNode = editor.vDocument.root.next(node => node.value === 'b');
+                        const params: OptionalRangeParams = {
+                            range: {
+                                start: bNode,
+                                end: bNode,
+                            },
+                        };
+
+                        editor.execCommand('deleteBackward', params);
+                        editor.renderer.render(editor.vDocument, editor.editable);
```
Goaman: why render twice as we use `execCommand`?

dmo-odoo:
I'm closing this since the feature is included in https://github.com/odoo-dev/jabberwock/pull/70 which also introduces other changes.

PR #50, opened 2019-11-28
Master indent nby

Goaman:


packages/core/src/CorePlugin.ts
```diff
@@ -129,6 +138,130 @@ export class CorePlugin extends JWPlugin {
...
+        });
+    }
+
+    /**
+     * Dedent lines.
+     *
+     * If there is more than one line selected, for each of the lines, remove up to 4 spaces in the
+     * beggining of the line.
+     */
+    dedent(optionalRangeParams: OptionalRangeParams): void {
```
Zinston: "outdent" is the proper term.

packages/core/src/EventManager.ts
```diff
@@ -124,6 +124,18 @@ export class EventManager {
...
+            case 'tab':
+                if (!payload.shiftKey && !payload.ctrlKey && !payload.altKey && !payload.metaKey) {
+                    return this.editor.execCommand('indent', {});
+                } else if (
+                    payload.shiftKey &&
+                    !payload.ctrlKey &&
+                    !payload.altKey &&
+                    !payload.metaKey
+                ) {
+                    return this.editor.execCommand('dedent', {});
```
Zinston: "outdent"

packages/utils/src/utils.ts
```diff
@@ -22,3 +22,10 @@ export const utils = {
         return Array.prototype.slice.call(collection);
     },
 };
+
+/**
+ * Remove duplicated values of `array`.
+ */
+export function distinct<T>(array: T[]): T[] {
```
Zinston: `uniq` ?

packages/core/test/CorePlugins.test.ts
```diff
@@ -0,0 +1,130 @@
+import { testEditor } from '../../utils/src/testUtils';
+import JWEditor from '../src/JWEditor';
+
+const indentCommand = (editor: JWEditor) => editor.execCommand('indent');
+const dedentCommand = (editor: JWEditor) => editor.execCommand('dedent');
```
Zinston: "outdent" ;-)

packages/core/src/CorePlugin.ts
```diff
@@ -129,6 +138,130 @@ export class CorePlugin extends JWPlugin {
...
+            // Only dedent when there is at leat two lines selected
+            if (firstChars.length > 1) {
+                firstChars.forEach((firstCharVNode: VNode) => {
+                    const spaceChars = [];
+                    let currentCharVNode = firstCharVNode;
+                    for (let i = 0; i < numberOfSpacePerIndent; i++) {
+                        if (
+                            currentCharVNode &&
+                            currentCharVNode.type === VNodeType.CHAR &&
+                            (currentCharVNode.value === ' ' || currentCharVNode.value === '\u00A0')
```
Zinston: How about `^/\s$/g.test(currentCharVNode.value)`?

dmo-odoo:
Blocked by https://github.com/odoo-dev/jabberwock/pull/48

packages/core/src/CorePlugin.ts
```diff
@@ -128,6 +141,121 @@ export class CorePlugin extends JWPlugin {
...
+     * ```
+     * will return the char node: a,d,g,i,k
+     */
+    getFirstForEachLineInRange(range: VRange): VNode[] {
+        const firstChars = [];
+        const lineBreaks = range.selectedNodes.filter(node => node instanceof LineBreakNode);
+        const parents = distinct(
+            range.selectedNodes.filter(n => n.children.length === 0).map(node => node.parent),
+        );
+        if (lineBreaks.length || parents.length) {
```
Goaman: Why testing `lineBreaks.length`? Don't `parents.length` always contain at least one parent?
Also, `lineBreaks.length` serve no purpose because even if `lineBreaks` has no elements, it will simply not add any chars in the instruction below.

packages/core/test/CorePlugins.test.ts
```diff
@@ -0,0 +1,130 @@
+import { testEditor } from '../../utils/src/testUtils';
+import JWEditor from '../src/JWEditor';
+
+const indent = (editor: JWEditor) => editor.execCommand('indent');
+const outdent = (editor: JWEditor) => editor.execCommand('outdent');
+
+describe('core', () => {
+    describe('utils', () => {
+        describe('CorePlugin', () => {
+            describe('tab', () => {
```
Goaman: `describe('indent/outdent'` rather than 'describe('tab''?
Tab is just one way to indent/outdent. A button can also do the job.

packages/core/test/CorePlugins.test.ts
```diff
@@ -0,0 +1,130 @@
...
+                                        },
+                                    });
+                                },
+                                contentAfter: 'ab<br>cd[]',
+                            });
+                        });
+                    });
+                });
+            });
+            it('should do nothing when another modifier then shift is triggered', () => {});
```
Goaman: This test will provide a false positive I think.
To investigate.

packages/core/src/CorePlugin.ts
```diff
@@ -130,41 +129,32 @@ export class CorePlugin extends JWPlugin {
...
-        } else {
-            const firstCharsOfLines = this.getFirstForEachLineInRange(range);
-            // If there is at least two lines
-            if (firstCharsOfLines.length > 1) {
-                // For each `firstCharsOfLines`, add all the `indentChars` in the document
-                firstCharsOfLines.forEach((firstCharNode: VNode) => {
-                    this.editor.vDocument.withRange(VRange.at(firstCharNode), range => {
-                        this.editor.vDocument.insertText(range, this.indentChars);
-                    });
+        const isSegmentBreak = (node: VNode): boolean => {
```
Goaman: The following lines are reapeated in both function:
```
const isSegmentBreak = (node: VNode): boolean => {
const numberOfSpacePerIndent = this.indentChars.length;	            return node instanceof LineBreakNode || isChar(node.firstChild());
};
const segmentBreaks = range.selectedNodes.filter(isSegmentBreak);
const tabLength = this.indentChars.length;
// Only outdent when there is at leat two lines selected	        // Only outdent when there is at leat two lines selected
if (firstChars.length > 1) {	        if (segmentBreaks.length) {
    firstChars.forEach((firstCharVNode: VNode) => {	            // The first line is usually only partially selected, which means
        const spaceChars = [];	            // the segment break preceding it was not in `range.selectedNodes`.
        let currentCharVNode = firstCharVNode;	            segmentBreaks.unshift(range.start.previous(isSegmentBreak));
```

Is the function `getFirstForEachLineInRange` containing that logic more flexible?

packages/core/src/CorePlugin.ts
```diff
@@ -130,41 +129,32 @@ export class CorePlugin extends JWPlugin {
...
+            return node instanceof LineBreakNode || isChar(node.firstChild());
+        };
+        const segmentBreaks = range.selectedNodes.filter(isSegmentBreak);
+        if (segmentBreaks.length) {
+            // The first line is usually only partially selected, which means
+            // the segment break preceding it was not in `range.selectedNodes`.
+            segmentBreaks.unshift(range.start.previous(isSegmentBreak));
+            segmentBreaks.forEach(br => {
```
Goaman: `br` might not be a `br`. Maybe calling it `segmentBreak`?

PR #53, opened 2019-12-06
[𝐼𝑀𝒫] 𝒱𝒩𝑜𝒹𝑒: 𝓇𝑒𝒾𝓂𝓅𝓁𝑒𝓂𝑒𝓃𝓉 𝒾𝓈𝐵𝑒𝒻𝑜𝓇𝑒 𝒾𝓃 𝒶𝓃 𝑒𝒻𝒻𝒾𝒸𝒾𝑒𝓃𝓉 𝓌𝒶𝓎

Zinston:
This reimplements the `isBefore` method of `VNode`, which aims at determining whether a given node is first in the traversal of the tree (depth first, pre-order). Before this, the method was implemented naively, traversing the entire tree.
The new method is very efficient, which could justify removing the `direction` property of `VRange` in favor of a simple call to this. In the worst case scenario we need only traverse the ancestors of both
nodes all the way to the root.

Here's an explanation:

1. Generate the ancestors paths of each VNode up to the root.
2. Compare the ancestors paths starting by the root:
    1. If both are the same, move on to the next ancestors in the paths.
    2. If one of the ancestors is undefined, then the VNode that
       originally generated this path is itself part of the ancestors
       path of the other VNode. This VNode is definitely first in
       traversal since the other is a descendent of it.
    3. Otherwise, compare the ancestors indices. The smaller of the two
       indices gives us the first VNode in traversal.

**Examples:**
Consider the following DOM:
```html
<root>
    <div>
        <span>textA<br/>textB</span>textC
    </div>
    <div>textD<b>textE</b>textF</div>
</root>
```
Its equivalent in tree form, with indices:
```
            0
          /   \
        0       1
       / \    / | \
      0   1  0  1  2
     /|\        |
    0 1 2       0
```
**Example 1: which is first: textA or textC?**
let A = textA and B = textC
<pre>
            0
          /   \
        0       1
       / \    / | \
      0   <b>B</b>  0  1  2
     /|\        |
    <b>A</b> 1 2       0
</pre>
pathA = [0, 0, 0, 0]
pathB = [0, 0, 1]
=> 0 < 1
=> A is first

**Example 2: which is first: span or textA?**
let A = span and B = textA
<pre>
            0
          /   \
        0       1
       / \    / | \
      <b>A</b>   1  0  1  2
     /|\        |
    <b>B</b> 1 2       0
</pre>
pathA = [0, 0, 0]
pathB = [0, 0, 0, 0]
=> pathA[3] is undefined
=> A is first

### DIFF
--- a/packages/core/src/VNode.ts
+++ b/packages/core/src/VNode.ts
@@ -209,20 +209,32 @@ export class VNode {
      * Return true if this VNode comes before the given VNode in the pre-order
      * traversal.
      *
-     * TODO: Make a less naive version of this, performing an efficient search
-     * instead of a full traversal.
-     *
      * @param vNode
      */
     isBefore(vNode: VNode): boolean {
-        return !!this.next((node: VNode) => node.id === vNode.id);
+        const path = this._pathToRoot(this);
+        const otherPath = this._pathToRoot(vNode);
+        let ancestor = path.pop();
+        let otherAncestor = otherPath.pop();
+        // Compare the ancestors of each nodes one by one, in a path to the
+        // root. While the ancestors are the same in both, continue on the path.
+        while (ancestor && otherAncestor && ancestor === otherAncestor) {
+            ancestor = path.pop();
+            otherAncestor = otherPath.pop();
+        }
+        // If we reached the end of one of the paths, that is when one of the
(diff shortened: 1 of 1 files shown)


Zinston:
@dmo-odoo :sunglasses:

PR #54, opened 2019-12-09
[ADD] Format: support all format tags

Zinston:
Prior to this PR, the only formats that were supported were `bold`, `italic` and `underline`. This adds support for `strong`, `emphasis`, `subscript`, `superscript` and `strikethrough`.
Some refactoring of Format is included, that became relevant once the list of Formats became bigger.


packages/core/src/VNodes/CharNode.ts
```diff
@@ -20,26 +21,23 @@ export function makeChar(char: string): Char {
...
     char: Char;
     // Format
     bold = false;
     italic = false;
     underline = false;
+    strikethrough = false;
+    subscript = false;
+    superscript = false;
+    strong = false;
+    emphasis = false;
```
dmo-odoo: I just went quickly through this PR changes on my phone while waiting for the alarm clock. I'm uneasy with this PR because I feel it is strengthening the fundamental issue I already had with formats : they are hard coded. Would it be possible to have a mechanism such that plugins can define new formats? That would add the key to the format property of VNode (which would be [key of string: boolean] then) and register the proper format tag in the utils. Maybe the format utils should become the "Format" plugin and the others formats would be separated like "Bold", "Superscript" etc? Of course they would not do a lot of things except define what is bold and how to represent it, as all the heavy lifting of parsing and rendering formats would be offloaded to the Format Plugin.

Zinston: That would be nice indeed. Now @Gorash wanted to avoid having objects on the nodes as much as possible for the sake of performance of the memory system, which is why I took all these properties out of the `format` object, which I then removed. I'm looking into an alternative way of doing something like what you're suggesting.

dmo-odoo: I don't think we should make api design choices based on the performance of the memory. If we really can't have our cake and eat it too, then I think it should be the other way around. For example (that's just a suggestion since I don't currently have a deep understanding of the issue), we could have a "simpler" object type in the memory than "object". For example "record" (to match Typescript's terminology), that would be more efficient than "object" because it would only have to handle some specific cases (assigning string keys to a single type) rather than the "everything is possible, yeah!" of standard Javascript objects. My idea basically: if using pure "objects" in memory is too costly and there really is nothing we can do about it in the first place, then we might want to introduce a different object type in memory and explicitly label the "format" key as such. cc @Gorash (not sure mentions work on the phone, if not, please correct this comment)

PR #55, opened 2019-12-09
Master node classes age tests chm

Gorash:
add tests: 100% coverage for VNodes

### DIFF
--- a/packages/core/src/VNodes/CharNode.ts
+++ b/packages/core/src/VNodes/CharNode.ts
@@ -1,4 +1,4 @@
-import { VNode, VNodeType } from '../VNode';
+import { VNode, VNodeType } from './VNode';
 import { utils } from '../../../utils/src/utils';
 
 /**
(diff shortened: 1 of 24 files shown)


packages/core/src/VNodes/VNode.ts
```diff
@@ -35,23 +35,6 @@ export class VNode {
         this.name = this.type;
         id++;
     }
-    /**
```
Zinston: Do not remove.

packages/core/src/VNodes/CharNode.ts
```diff
@@ -63,7 +63,7 @@ export class CharNode extends VNode {
      *
      * @override
      */
-    shallowDuplicate(): VNode {
```
Zinston: Split this commit in two: one with the refactorings, one with only tests.

packages/core/test/Dispatcher.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-nested-callbacks */
```
Zinston: Why?

packages/core/test/VNodes.test.ts
```diff
@@ -0,0 +1,107 @@
+/* eslint-disable max-nested-callbacks */
```
Zinston: Why is this in this commit?

PR #56, opened 2019-12-10
Master test devtools chm

Gorash:
add coverage for devtools + fix (remove VNode from owl's useState)

### DIFF
--- a/packages/plugin-devtools/src/components/CommandsComponent.ts
+++ b/packages/plugin-devtools/src/components/CommandsComponent.ts
@@ -21,7 +21,7 @@ interface CommandsProps {
 
 export class CommandsComponent extends OwlUIComponent<CommandsProps> {
     state: CommandsState = {
-        currentTab: 'selected',
+        currentTab: 'queue',
         registry: this.env.editor.dispatcher.commands,
         handlers: this.env.editor.dispatcher.handlers,
         selectedCommandIndex: null, // Index of the selected command in the stack
@@ -45,14 +45,6 @@ export class CommandsComponent extends OwlUIComponent<CommandsProps> {
             return '<' + value.nodeName.toLowerCase() + '>';
         }
         return '' + value;
-    }
-    /**
-     * Return the command handler corresponding to the given token.
-     *
-     * @param commandIdentifier
-     */
-    getHandlers(commandIdentifier: CommandIdentifier): CommandHandler[] {
-        return this.env.editor.dispatcher[commandIdentifier];
     }
     /**
      * Handle keydown event to navigate in the command stack.
--- a/packages/plugin-devtools/src/components/InfoComponent.ts
+++ b/packages/plugin-devtools/src/components/InfoComponent.ts
@@ -1,10 +1,8 @@
 import { OwlUIComponent } from '../../../owl-ui/src/OwlUIComponent';
-import { VRange } from '../../../core/src/VRange';
 import { VNode } from '../../../core/src/VNodes/VNode';
 
 interface InfoState {
     currentTab: string;
-    range: VRange;
 }
 export class InfoComponent extends OwlUIComponent<{}> {
     aboutMeProps = ['id', 'name', 'type', 'length', 'atomic', 'text'];
@@ -12,7 +10,6 @@ export class InfoComponent extends OwlUIComponent<{}> {
     customPropsBlacklist = this.aboutMeProps.concat(this.familyProps).concat(['renderingEngines']);
     state: InfoState = {
         currentTab: 'vNode',
-        range: this.env.editor.vDocument.range,
     };
     localStorage = ['currentTab'];
     /**
--- a/packages/plugin-devtools/src/components/InspectorComponent.ts
+++ b/packages/plugin-devtools/src/components/InspectorComponent.ts
@@ -4,19 +4,12 @@ import { TreeComponent } from './TreeComponent';
 import { OwlUIComponent } from '../../../owl-ui/src/OwlUIComponent';
 import { VNode } from '../../../core/src/VNodes/VNode';
 
-////////////////////////////// todo: use API ///////////////////////////////////
-
-interface InspectorState {
-    selectedNode: VNode;
-    selectedPath: VNode[]; // From highest parent to currently selected node
-}
-
 export class InspectorComponent extends OwlUIComponent<{}> {
     static components = { InfoComponent, PathComponent, TreeComponent };
-    state: InspectorState = {
-        selectedNode: this.env.editor.vDocument.root,
-        selectedPath: this._getPath(this.env.editor.vDocument.root),
+    state = {
+        selectedNodeID: 0,
     };
+    selectedNode = this.env.editor.vDocument.root;
 
     /**
      * Handle keyboard navigation in DevTools (arrows to move in the tree)
@@ -24,7 +17,7 @@ export class InspectorComponent extends OwlUIComponent<{}> {
      * @param {KeyboardEvent} event
      */
     onKeydown(event: KeyboardEvent): void {
-        const selected: VNode = this.state.selectedNode;
+        const selected: VNode = this.selectedNode;
         let newSelection: VNode;
         switch (event.code) {
             case 'ArrowDown':
@@ -39,10 +32,13 @@ export class InspectorComponent extends OwlUIComponent<{}> {
             case 'ArrowRight':
                 newSelection = selected.nextSibling();
                 break;
+            default:
+                return;
         }
         if (newSelection) {
             event.preventDefault();
-            this._selectNode(newSelection);
+            this.selectedNode = newSelection;
+            this.state.selectedNodeID = newSelection.id;
         }
     }
     /**
@@ -56,13 +52,9 @@ export class InspectorComponent extends OwlUIComponent<{}> {
      */
     selectNode(event: CustomEvent): void {
         const vNode: VNode = event.detail.vNode;
-        this._selectNode(vNode);
+        this.selectedNode = vNode;
+        this.state.selectedNodeID = vNode.id;
     }
-
-    //--------------------------------------------------------------------------
-    // Private
-    //--------------------------------------------------------------------------
-
     /**
      * Return the path between the given `vNode` and the root vNode, as an array
      * of VNode, sorted from the highest parent (the root) to the given vNode.
@@ -70,7 +62,7 @@ export class InspectorComponent extends OwlUIComponent<{}> {
      * @param {VNode} vNode
      * @returns {VNode []}
      */
-    _getPath(vNode: VNode): VNode[] {
+    getPath(vNode: VNode): VNode[] {
         const path: VNode[] = [vNode];
         let parent: VNode = vNode.parent;
         while (parent) {
@@ -78,14 +70,5 @@ export class InspectorComponent extends OwlUIComponent<{}> {
             parent = parent.parent;
         }
         return path;
-    }
-    /**
-     * Update the state to account for the selection of the given vNode
-     *
-     * @param {VNode} vNode
-     */
-    _selectNode(vNode: VNode): void {
-        this.state.selectedNode = vNode;
-        this.state.selectedPath = this._getPath(vNode);
     }
 }
--- a/packages/plugin-devtools/test/devtools.test.ts
+++ b/packages/plugin-devtools/test/devtools.test.ts
@@ -1,0 +1,926 @@
+/* eslint-disable max-nested-callbacks */
+import { expect } from 'chai';
+import JWEditor from '../../core/src/JWEditor';
+import { VNode } from '../../core/src/VNodes/VNode';
+import { click, nextTickFrame, keydown } from '../../utils/src/testUtils';
+
+async function openDevTools(devtools: HTMLElement): Promise<void> {
+    await click(devtools.querySelector('devtools-navbar'));
+}
+
+describe('plugin-devtools', () => {
+    let wrapper: HTMLElement;
+    let devtools: HTMLElement;
+    let editor: JWEditor;
+
+    beforeEach(async () => {
+        wrapper = document.createElement('test-wrapper');
+        wrapper.style.display = 'block';
+        document.body.appendChild(wrapper);
+
+        const root = document.createElement('test-container');
+        root.style.display = 'block';
+        root.innerHTML = '<h1>Title</h1><p>a<b>b</b>c</p><p>def<br>gh<i>i</i></p><div>div</div>';
+        wrapper.appendChild(root);
+
+        Object.keys(localStorage).forEach(key => {
+            if (key.indexOf('OwlUI') !== -1) {
+                localStorage.removeItem(key);
+            }
+        });
+
+        editor = new JWEditor(root);
+        editor.loadConfig({
+            debug: true,
+        });
+
+        await editor.start();
+
+        devtools = editor.el.querySelector('jw-devtools');
+    });
+    afterEach(async () => {
+        editor.stop();
+        document.body.removeChild(wrapper);
+    });
+
+    describe('DevToolsComponent', () => {
+        it('should display the devtools', async () => {
+            expect(!!devtools).to.equal(true);
+        });
+        it('should display the devtools closed as bottom bar', async () => {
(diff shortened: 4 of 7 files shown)


packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+        const root = document.createElement('test-container');
+        root.style.display = 'block';
+        root.innerHTML = '<h1>Title</h1><p>a<b>b</b>c</p><p>def<br>gh<i>i</i></p><div>div</div>';
+        wrapper.appendChild(root);
+
+        Object.keys(localStorage).forEach(key => {
+            if (key.indexOf('OwlUI') !== -1) {
+                localStorage.removeItem(key);
+            }
+        });
```
dmo-odoo: Well, then you don't test that the local storage works so this is not true 100% coverage.

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+        editor.loadConfig({
+            debug: true,
+        });
+
+        await editor.start();
+
+        devtools = editor.el.querySelector('jw-devtools');
+    });
+    after(async () => {
+        editor.stop();
```
dmo-odoo: You should reset the values of the editor and devtools variables.

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+            expect(devtools.classList.contains('closed')).to.equal(true);
+        });
+        it('should open the devTools', async () => {
+            const navbar = devtools.querySelector('devtools-navbar');
+            navbar.dispatchEvent(new MouseEvent('mousedown', { bubbles: true }));
+            await nextTick();
+            navbar.dispatchEvent(new MouseEvent('mouseup', { bubbles: true }));
+            await nextTick();
+            navbar.dispatchEvent(new MouseEvent('click', { bubbles: true }));
+            await nextTick();
```
dmo-odoo: We need `await testUtils.click(navbar)` which does this.

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+            navbar.dispatchEvent(new MouseEvent('mousedown', begin));
+            await longNextTick();
+            window.dispatchEvent(new MouseEvent('mousemove', step));
+            await longNextTick();
+            window.dispatchEvent(new MouseEvent('mousemove', end));
+            await longNextTick();
+            navbar.dispatchEvent(new MouseEvent('mouseup', end));
+            await nextTick();
+            navbar.dispatchEvent(new MouseEvent('click', end));
+            await nextTick();
```
dmo-odoo: Maybe, just maybe, we might need a testUtils.dragAndDrop function. It wouldn't be of great use now but it might be if we have more dragging stuff coming later.

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+            button.dispatchEvent(new MouseEvent('click', { bubbles: true }));
+            await longNextTick();
+            expect(devtools.classList.contains('closed')).to.equal(false);
+        });
+    });
+    describe('InspectorComponent', () => {
+        it('should change panel to "Inspector" with top button', async () => {
+            const button = devtools.querySelector('devtools-navbar button');
+            expect(button.textContent).to.equal('Inspector');
+            expect(button.classList.contains('selected')).to.equal(false, 'button is not selected');
```
dmo-odoo: Wait, how is it false ? This is another describe, it should not depend on previous tests !

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+            const button = devtools.querySelector('devtools-info devtools-navbar button');
+            expect(button.textContent).to.equal('VNode');
+            expect(button.classList.contains('selected')).to.equal(true, 'button is selected');
+        });
+        it('should display the root by default in info', async () => {
+            const about = devtools.querySelector('devtools-info .about');
+            const aResult =
+                '<div class="about">' +
+                '<span class="type">VNode</span> ROOT<span class="id">0</span>' +
+                '</div>';
```
dmo-odoo: Could we test this without being 100% coupled with the DOM structure of the devtools ? Because that is subject to change.

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+                '</tbody>' +
+                '</table>' +
+                '<div class="divider">👪 My Family</div>' +
+                '<table>' +
+                '<tbody>' +
+                '<tr><td>index</td><td>0</td></tr>' +
+                '<tr><td>children</td><td><ol><li>HEADING1</li><li>PARAGRAPH</li><li>PARAGRAPH</li><li>?</li></ol></td></tr>' +
+                '</tbody>' +
+                '</table>' +
+                '</div>';
```
dmo-odoo: Same here, if I change *anything* in the DOM structure of the Devtools, this test will fail. I would prefer that the test tests that each property is properly displayed inside the div with class "properties" but nothing more, otherwise we won't be able to change anything in the display without having to adapt the tests. Granted, these tests are strong, but I fear they are in fact too strong. Not sure about that, we'll need to discuss it with the others.

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+            expect(node.classList.contains('folded')).to.equal(true, 'node is folded');
+            expect(node.querySelector('.children').children.length).to.equal(
+                0,
+                'children are not loaded',
+            );
+
+            const about = devtools.querySelector('devtools-info .about');
+            const aResult =
+                '<div class="about">' +
+                '<span class="type">VNode</span> PARAGRAPH<span class="id">9</span>' +
```
dmo-odoo: For example, if we change how VNode ids are generated, all tests doing this will fail...

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+            );
+            const span = node.firstElementChild;
+
+            node.dispatchEvent(
+                new KeyboardEvent('keydown', {
+                    key: 'ArrowDown',
+                    code: 'ArrowDown',
+                    bubbles: true,
+                }),
+            );
```
dmo-odoo: Maybe this could be an helper too. `await testUtils.keydown('ArrowDown', node)`

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+            await longNextTick();
+            expect(span.classList.contains('selected')).to.equal(false, 'unselect');
+            const next = node.nextElementSibling;
+            expect(next.firstElementChild.classList.contains('selected')).to.equal(
+                true,
+                'select the next',
+            );
```
dmo-odoo: You can do `to.be.true` and the label at the end is not really useful IMO. This applies to a lot of lines... :p

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+                new KeyboardEvent('keydown', {
+                    key: 'ArrowDown',
+                    code: 'ArrowDown',
+                    bubbles: true,
+                }),
+            );
+            await longNextTick();
+            expect(span.classList.contains('selected')).to.equal(false, 'unselect');
+            expect(
+                last.lastElementChild.firstElementChild.firstElementChild.classList.contains(
```
dmo-odoo: 👀

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+            const name = node.querySelector('.element-name');
+            const pos = name.getBoundingClientRect();
+            name.dispatchEvent(
+                new MouseEvent('click', { clientX: pos.left, clientY: pos.top, bubbles: true }),
+            );
+            await longNextTick();
+            expect(node.classList.contains('folded')).to.equal(false, 'node is unfolded');
+            expect(node.querySelector('.children').children.length).to.equal(
+                3,
+                'children are not loaded',
```
dmo-odoo: Wait, there are 3 children but they "are not loaded" ?

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+        });
+        it('should open/close the last paragraph with enter', async () => {
+            const node = devtools.querySelector(
+                'devtools-node .children devtools-node.element:nth-child(3)',
+            );
+            node.dispatchEvent(new KeyboardEvent('keydown', { key: 'Enter', code: 'Enter' }));
+            await longNextTick();
+            expect(node.classList.contains('folded')).to.equal(false, 'node is unfolded');
+            expect(node.querySelector('.children').children.length).to.equal(
+                7,
```
dmo-odoo: Same as "3"

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+                'children are not loaded',
+            );
+
+            node.dispatchEvent(new KeyboardEvent('keydown', { key: 'Enter', code: 'Enter' }));
+            await longNextTick();
+            expect(node.classList.contains('folded')).to.equal(true, 'node is folded');
+            expect(node.querySelector('.children').children.length).to.equal(
+                0,
+                'children are removed',
```
dmo-odoo: You mean "not displayed anymore", right ?

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+                'devtools-tree > devtools-node > .children > devtools-node:nth-child(2)',
+            );
+            const name = p.firstElementChild;
+            const pChildren = p.lastElementChild;
+
+            expect(pChildren.childNodes.length).to.equal(3);
+
+            const pos = name.getBoundingClientRect();
+            name.dispatchEvent(
+                new MouseEvent('dblclick', { clientX: pos.left, clientY: pos.top, bubbles: true }),
```
dmo-odoo: Not sure this works on Safari, I need to test it.

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+            let about = devtools.querySelector('devtools-info .about');
+            let aResult =
+                '<div class="about">' +
+                '<span class="type">VNode</span> PARAGRAPH<span class="id">9</span>' +
+                '</div>';
+            expect(about.outerHTML).to.equal(aResult);
+
+            const li = [].slice
+                .call(devtools.querySelectorAll('devtools-info .properties li'))
+                .pop();
```
dmo-odoo: Why not use Array.prototype.slice if you really want to do that ? Otherwise, you could also do `Array.from(...).slice().pop()`

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+        it('should change sub panel to "Registry"', async () => {
+            const panel = devtools.querySelector('devtools-panel.active');
+            const button = panel.querySelector('devtools-navbar button:not(.selected)');
+            expect(button.textContent).to.equal('Registry');
+            button.dispatchEvent(new MouseEvent('click', { bubbles: true }));
+            await longNextTick();
+            expect(button.classList.contains('selected')).to.equal(true, 'button is selected');
+            const subpanel = devtools.querySelector('devtools-panel.active mainpane-contents');
+            expect(subpanel.firstElementChild.nodeName).to.equal('DIV');
+            expect([].map.call(subpanel.childNodes, (n: Node) => n.textContent)).to.deep.equal([
```
dmo-odoo: Same here, you could do `Array.from(subpanel.childNodes).map(...)` directly.

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+                new KeyboardEvent('keydown', {
+                    key: 'ArrowDown',
+                    code: 'ArrowDown',
+                    bubbles: true,
+                }),
+            );
+            await longNextTick();
+            expect(selected.nextElementSibling.classList.contains('selected')).to.equal(true);
+        });
+        it('should not change the selection with other key', async () => {
```
dmo-odoo: Isn't there already a test that tests this ?

packages/plugin-devtools/src/components/InspectorComponent.ts
```diff
@@ -6,14 +6,17 @@ import { VNode } from '../../../core/src/VNode';
...
+interface InspectorValues {
     selectedNode: VNode;
     selectedPath: VNode[]; // From highest parent to currently selected node
 }
 
 export class InspectorComponent extends OwlUIComponent<{}> {
     static components = { InfoComponent, PathComponent, TreeComponent };
-    state: InspectorState = {
+    state = {
+        redrawWithoutLinkVNodes: 0,
```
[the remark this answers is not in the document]
Gorash: force redraw when select a new node, because the state does not change. Can't put the VNode into owl (for proxy)

packages/plugin-devtools/src/components/InfoComponent.ts
```diff
@@ -2,14 +2,18 @@ import { OwlUIComponent } from '../../../owl-ui/src/OwlUIComponent';
...
+    range: VRange;
+}
 interface InfoState {
     currentTab: string;
-    range: VRange;
 }
 export class InfoComponent extends OwlUIComponent<{}> {
+    values: InfoValue = {
+        range: this.env.editor.vDocument.range,
+    };
```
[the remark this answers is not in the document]
Gorash: i don't see how

packages/plugin-devtools/src/components/CommandsComponent.ts
```diff
@@ -21,7 +21,7 @@ interface CommandsProps {
 
 export class CommandsComponent extends OwlUIComponent<CommandsProps> {
     state: CommandsState = {
-        currentTab: 'selected',
+        currentTab: 'queue',
```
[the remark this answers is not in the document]
Gorash: 0.o

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -0,0 +1,735 @@
...
+import { expect } from 'chai';
+import JWEditor from '../../core/src/JWEditor';
+
+async function nextTick(): Promise<void> {
+    return new Promise((resolve): number => setTimeout(resolve));
+}
+
+async function longNextTick(): Promise<void> {
+    return new Promise((resolve): void => {
+        setTimeout(resolve, 50);
```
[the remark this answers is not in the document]
Gorash: :+1: will use requestAnimationFrame

packages/plugin-devtools/test/devtools.test.ts
Gorash: No, here we test that it doesn't matter (=> 100% coverage)

packages/plugin-devtools/test/devtools.test.ts
dmo-odoo: Maybe this error is elsewhere as well, please check

packages/plugin-devtools/test/devtools.test.ts
dmo-odoo: Please change to "removed from devtools dom" (or devtools "display")

packages/plugin-devtools/test/devtools.test.ts
Gorash: I wrote the tests for what is in plugin-devtools, here part of the code is in another folder, it will be another test which will not use this before

packages/plugin-devtools/test/devtools.test.ts
Gorash: Yes, and if that changes it is normal to update the tests. this has already allowed an error to be detected (used in Antoine's branch)

Gorash:
ping @dmo-odoo to merge ;)

PR #57, opened 2019-12-11
[POC] Format: support all format tags, with classes, and make it modular

Zinston:
_This PR replaces https://github.com/odoo-dev/jabberwock/pull/54._
_This is a WIP_

Prior to this PR, the only formats that were supported were `bold`, `italic` and `underline`. This adds support for `strong`, `emphasis`, `subscript`, `superscript` and `strikethrough`.
In the process it makes Formats modular and adds support for classes on them.

TODO:
- add delete/backspace/enter tests in waiting
- the way `formatManager` is instantiated in `JWEditor` then passed around everywhere is less than ideal -> find a better way
- manage nested formats with different classes
- support style and other attributes
- the current structure of `Bold`, `Italic` and the other format files is too cumbersome
- doc


packages/plugin-devtools/src/components/InfoComponent.ts
```diff
@@ -2,15 +2,19 @@ import { OwlUIComponent } from '../../../owl-ui/src/OwlUIComponent';
...
+}
 interface InfoState {
     currentTab: string;
-    range: VRange;
 }
 const vNodeKeys = Object.keys(new VNode('default'));
 export class InfoComponent extends OwlUIComponent<{}> {
+    values: InfoValue = {
+        range: this.env.editor.vDocument.range,
+    };
```
dmo-odoo: @Zinston Couldn't these "values" be props ? We already have `state` and `props`, so I'm asking to avoid introducing yet another place to find the data.

packages/plugin-devtools/src/components/InspectorComponent.ts
```diff
@@ -6,14 +6,17 @@ import { VNode } from '../../../core/src/VNodes/VNode';
...
+interface InspectorValues {
     selectedNode: VNode;
     selectedPath: VNode[]; // From highest parent to currently selected node
 }
 
 export class InspectorComponent extends OwlUIComponent<{}> {
     static components = { InfoComponent, PathComponent, TreeComponent };
-    state: InspectorState = {
+    state = {
+        redrawWithoutLinkVNodes: 0,
```
dmo-odoo: ?

examples/index/index.css
```diff
@@ -13,3 +13,9 @@ jw-editable {
     padding: 10px;
     height: 100vh;
 }
+
+jw-editable .drop-cap {
```
dmo-odoo: ?

dmo-odoo: Since this PR is a wip, I'm not sure it's ready for review, so I'll stop here.

Zinston: This is for the sake of example. The example page's first letter has a span with that class so it looks like the first letter in a chapter of a book.

packages/plugin-devtools/src/components/InfoComponent.ts
Zinston: This is @Gorash 's commit, cherry-picked from another branch.

packages/plugin-devtools/src/components/InspectorComponent.ts
Zinston: Same comment as above.

examples/index/index.css
Zinston: It's a wip because we need to look into it together so I wouldn't stop the review quite yet.

PR #60, opened 2019-12-14
WIP2 Custom nodes

Zinston:
This branch is https://github.com/odoo-dev/jabberwock/pull/52 with review in progress.


Zinston:
@dmo-odoo @Goaman @Gorash

packages/core/src/VNodes/RangeNode.ts
```diff
@@ -0,0 +1,10 @@
+import { VNode, VNodeType } from './VNode';
+
+export class RangeNode extends VNode {
+    properties = {
+        atomic: true,
+    };
+    constructor(tailOrHead: 'tail' | 'head') {
```
Goaman: Why use string rather than an ENUM?

packages/core/src/Renderer.ts
```diff
@@ -127,7 +128,7 @@ export class Renderer {
     _renderTextNode(context: RenderingContext): RenderingContext {
         // If the node has a format, render the format nodes first.
         const renderedFormats = [];
-        const firstChar = context.currentVNode;
+        const firstChar = context.currentVNode as CharNode;
```
Goaman: How can we be sure that the currentVNode is a CharNode?
We might need a type guard to check if it is really one, otherwise, a bug might slip here.

Goaman: The name `firstChar` is misleading, why not currentCharNode?

packages/core/src/VNodes/CharNode.ts
```diff
@@ -1,10 +1,67 @@
...
+ * @param char
+ */
+export function makeChar(char: string): Char {
+    if (char.length === 1) {
+        return char as Char;
+    } else {
+        throw new Error('Cannot make a Char out of anything else than a string of length 1.');
+    }
+}
+
```
Goaman: The logic is too complicated to understand for just checking that the char has is only of length one.
I would simply check it in the constructor or in a getter.

packages/core/src/VNodes/CharNode.ts
```diff
@@ -1,10 +1,67 @@
...
+    }
+
+    //--------------------------------------------------------------------------
+    // Public
+    //--------------------------------------------------------------------------
+
+    /**
+     * Return the length of this VNode.
+     */
+    get length(): number {
```
Goaman: Why do we need this? Its length can be inferred from its type.

packages/core/src/VNodes/CharNode.ts
```diff
@@ -1,10 +1,67 @@
...
+        return 1;
+    }
+    /**
+     * Return this VNode's inner text (concatenation of all descendent
+     * char nodes values).
+     *
+     * @param __current
+     */
+    text(__current = ''): string {
```
Goaman: Why having a variable that holds state inside its signature?
The creation of a variable that concatenates multiples string should belong to its caller.

packages/core/src/VNodes/VNode.ts
```diff
@@ -45,23 +45,23 @@ export class VNode {
...
         this.type = type;
         this.originalTag = originalTag;
-        this.value = value;
         this.format = format || {
             bold: false,
             italic: false,
             underline: false,
         };
         this._updateProperties();
+        this.name = this.type;
```
Goaman: Why its `type` becomes its `name`?

packages/core/src/VNodes/LineBreakNode.ts
```diff
@@ -1,10 +1,57 @@
...
+    // Lifecycle
+    //--------------------------------------------------------------------------
+
+    /**
+     * Render the VNode to the given format.
+     *
+     * @param [to] the name of the format to which we want to render (default:
+     * html)
+     */
+    render<T>(to = 'html'): T {
```
Goaman: The rendering API should be discussed.

packages/core/src/VNodes/LineBreakNode.ts
```diff
@@ -1,10 +1,57 @@
...
+        return new LineBreakNode();
+    }
+    /**
+     * Locate where to set the range, when it targets this VNode, at a certain
+     * offset. This allows us to handle special cases.
+     *
+     * @param domNode
+     * @param offset
+     */
+    locateRange(domNode: Node, offset: number): [VNode, RelativePosition] {
```
Goaman: Discuss in group about who should locateRange?

packages/core/src/VNodes/LineBreakNode.ts
```diff
@@ -29,6 +25,19 @@ export class LineBreakNode extends VNode {
...
+    //--------------------------------------------------------------------------
+    // Public
+    //--------------------------------------------------------------------------
+
+    /**
+     * Return true if the VNode is atomic (ie. it may not have children).
+     *
+     * @override
+     */
+    get atomic(): boolean {
```
Goaman: Is `isAtomic` a good name alternative?

packages/utils/src/range.ts
```diff
@@ -0,0 +1,22 @@
+type DomDirection = 'ltr' | 'rtl';
```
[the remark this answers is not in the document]
Zinston: The reason can be found in the commit message where that change belongs. We quickly get into circular dependency issues when putting too much in the core. More specifically, we had a circular dependency issue when these range utils were in VRange and Range.

packages/utils/src/utils.ts
```diff
@@ -1,3 +1,5 @@
+export let isWithRange = false;
```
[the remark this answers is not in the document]
Zinston: Same answer as above ;-)

packages/core/src/Parser.ts
```diff
@@ -3,9 +3,13 @@ import { Format } from '../../utils/src/Format';
...
 import { VRangeDescription } from './VRange';
 import { Direction, RelativePosition } from '../../utils/src/range';
-import { VNode, FormatType, VNodeType } from './VNode';
+import { VNode, FormatType, VNodeType } from './VNodes/VNode';
 import { DomRangeDescription } from './EventNormalizer';
 import { utils } from '../../utils/src/utils';
+import { CharNode } from './VNodes/CharNode';
+import { SimpleElementNode } from './VNodes/SimpleElementNode';
+import { LineBreakNode } from './VNodes/LineBreakNode';
+import { RootNode } from './VNodes/RootNode';
```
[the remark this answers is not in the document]
Zinston: There is a TODO in the commit for that. That's out of scope for this PR but making the Parser node-agnostic is indeed needed, and the subject of a subsequent PR.

packages/utils/src/utils.ts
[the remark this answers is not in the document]
Zinston: Indeed. We noticed that too ;-)

packages/core/src/Parser.ts
```diff
@@ -74,29 +78,31 @@ function parseRange(range: Range | DomRangeDescription, direction?: Direction):
...
  * @param node
- * @returns the `VNodeType` of the given DOM node
  */
-function getNodeType(node: Node): VNodeType {
-    switch (node.nodeName) {
+function VNodeFromTag(tag: string): VNode {
+    switch (tag) {
         case 'P':
-            return VNodeType.PARAGRAPH;
+            return new SimpleElementNode(VNodeType.PARAGRAPH, tag);
```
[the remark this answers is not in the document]
Zinston: This is changed in a subsequent commit. Like so: `new SimpleElementNode('P')`.

packages/core/src/Parser.ts
```diff
@@ -129,6 +130,7 @@ function parseElementNode(currentContext: ParsingContext): ParsingContext {
             underline: format.underline || node.nodeName === 'U',
         });
     } else {
+        Object.assign(parsedNode.format, context.format[0]);
```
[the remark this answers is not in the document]
Zinston: This is also irrelevant if you look further in the PR. This commit was just moving stuff around, this is old code.

packages/core/src/VNodes/RangeNode.ts
Zinston: Again, removed later. What happened with this review?

packages/core/src/Renderer.ts
Zinston: We are sure because this method is called after checking that the node we are rendering is a text node.

Zinston: Because `firstChar` maps to the first character in the text node.

packages/core/src/VNodes/CharNode.ts
Zinston: Also removed already.

packages/core/src/VNodes/CharNode.ts
Zinston: _Because_ its length can be inferred from its type. This is an override of `VNode`'s `length` method, because in the case of a `CharNode` it shouldn't go look for how many children it has, but it's always 1. That said, I _should_ have put `@override` in the doc string.

packages/core/src/VNodes/CharNode.ts
Zinston: Tail recursion.

packages/core/src/VNodes/VNode.ts
Zinston: Outdated.

packages/utils/src/range.ts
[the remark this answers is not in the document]
Goaman: I saw the problem linked to the circular dependency but I don't think to solve it, the files and logic should belong to the utils folder.

packages/core/src/VNodes/VNode.ts
```diff
@@ -4,32 +4,33 @@ import { RelativePosition } from '../../../utils/src/range';
...
+    HEADING5 = 'heading5',
+    HEADING6 = 'heading6',
+    CHAR = 'char',
+    LINE_BREAK = 'line_break',
 }
 let id = 0;
 
 export class VNode {
-    readonly type: VNodeType;
+    readonly type: VNodeType | string;
```
Goaman: Why can it be a string?

packages/utils/src/utils.ts
```diff
@@ -39,11 +39,154 @@ export const utils = {
...
+    /**
+     * Return a string with the value of a text node stripped of its formatting
+     * space, applying the w3 rules for white space processing
+     * TODO: decide what exactly to do with formatting spaces:
+     * remove, keep, recompute?
+     *
+     * @see https://www.w3.org/TR/css-text-3/#white-space-processing
+     * @returns {string}
+     */
+    removeFormatSpace(node: Node): string {
```
Goaman: Should be in core is it uses a Node.

packages/utils/src/utils.ts
```diff
@@ -39,11 +39,154 @@ export const utils = {
...
+     * A segment break is a sort of line break, not considering automatic breaks
+     * that are function of the screen size. In this context, a segment is what
+     * you see when you triple click in text in the browser.
+     * Eg: `<div><p>◆one◆</p>◆two◆<br>◆three◆</div>` where ◆ = segment breaks.
+     *
+     * @param {Element} node
+     * @param {'start'|'end'} side
+     * @returns {boolean}
+     */
+    _isAtSegmentBreak(node: Node, side: 'start' | 'end'): boolean {
```
Goaman: Should be in core is it uses a Node.

packages/core/src/EventManager.ts
```diff
@@ -1,5 +1,5 @@
 import { EventNormalizer, DomRangeDescription } from './EventNormalizer';
-import { LineBreakNode } from './VNodes/LineBreakNode';
+import { LineBreakNode } from '../../plugin-linebreak/src/VNodes/LineBreakNode';
```
Zinston: `EventManager` should not be aware of `LineBreakNode`, given that it's optional.

packages/core/src/JWEditor.ts
```diff
@@ -96,21 +96,26 @@ export class JWEditor {
      *
      * @param pluginClass
      */
-    addPlugin(pluginClass: typeof JWPlugin): void {
-        const plugin: JWPlugin = new pluginClass(this);
-        this.pluginsRegistry.push(plugin);
-        // Register the commands of this plugin.
-        Object.keys(plugin.commands).forEach(key => {
-            this.dispatcher.registerCommand(key, plugin.commands[key]);
+    addPlugin(...pluginClasses: Array<typeof JWPlugin>): void {
```
Zinston: Update the doc string now it accepts several plugin classes.

packages/core/src/JWPlugin.ts
```diff
@@ -19,4 +20,5 @@ export class JWPlugin {
         // todo: namespace
         this.name = options.name || this.constructor.name;
     }
+    static parse: (node: Node) => VNode[] | Set<Attribute>;
```
Zinston: This return type is less than ideal. Maybe return a parsing context?

packages/core/src/Parser.ts
```diff
@@ -1,27 +1,27 @@
 import { VDocument } from './VDocument';
-import { Format } from '../../utils/src/Format';
+import { formatMap } from '../../plugin-format/src/formatMap';
```
Zinston: `Parser` should not be aware of `formatMap`, given that it's optional.

packages/core/src/Parser.ts
```diff
@@ -37,16 +37,24 @@ export class Parser {
...
      * Return a list of vNodes matching the given node.
      *
      * @param node
      */
-    parseNode(node: Node): VNode[] {
-        for (const VNodeClass of this._VNodes) {
-            const vNodes = VNodeClass.parse(node);
-            if (vNodes) {
-                return vNodes;
+    parseNode(node: Node): VNode[] | Set<Attribute> {
```
Zinston: This return type is less than ideal. Maybe return a parsing context?

packages/core/src/Parser.ts
```diff
@@ -61,7 +69,7 @@ export class Parser {
      * @returns the element parsed into the editor's virtual representation
      */
     parse(node: Node): VDocument {
-        const root = new FragmentNode();
+        const root = new FragmentNode('root');
```
Zinston: Remove this change.

packages/core/src/Parser.ts
```diff
@@ -130,70 +138,32 @@ export class Parser {
...
-            // internal representation but by the format of its children.
-            // For the parsing, encountering a format node generates a new format
-            // context which inherits from the previous one.
-            const format = context.format.length ? context.format[0] : {};
-            context.format.unshift({
-                bold: format.bold || node.nodeName === 'B',
-                italic: format.italic || node.nodeName === 'I',
-                underline: format.underline || node.nodeName === 'U',
+        const parseResult = this.parseNode(node);
+        if (Array.isArray(parseResult)) {
```
Zinston: Part of the problem of the double return type of `parseNode`. This actually checks if it returns `VNodes` instead of attributes.

packages/core/src/Parser.ts
```diff
@@ -130,70 +138,32 @@ export class Parser {
...
+                context.attributes.set(attribute.name, attribute.copy());
+            });
         }
         // A <br/> with no siblings is there only to make its parent visible.
         // Consume it since it was just parsed as its parent element node.
         // TODO: do this less naively to account for formatting space.
         if (node.childNodes.length === 1 && node.childNodes[0].nodeName === 'BR') {
             context.node = node.childNodes[0];
-            VDocumentMap.set(parsedNode, context.node);
+            if (Array.isArray(parseResult)) {
```
Zinston: Same as above.

packages/core/src/Parser.ts
```diff
@@ -206,28 +176,13 @@ export class Parser {
             !node.nextSibling.nextSibling
         ) {
             context.node = node.nextSibling;
-            VDocumentMap.set(parsedNode, context.node);
+            if (Array.isArray(parseResult)) {
```
Zinston: Same as above.

packages/core/src/Parser.ts
```diff
@@ -255,9 +210,9 @@ export class Parser {
             const rootNode = currentContext.rootNode;
             do {
                 ancestor = ancestor.parentNode;
-                if (ancestor && Format.tags.includes(ancestor.nodeName)) {
+                if (ancestor && formatMap.tags.includes(ancestor.nodeName)) {
```
Zinston: `Parser` should not be aware of `formatMap`, given that it's optional.

packages/core/src/Renderer.ts
```diff
@@ -1,11 +1,9 @@
 import { VDocumentMap } from './VDocumentMap';
 import { VDocument } from './VDocument';
-import { isMarker, isChar } from '../../utils/src/Predicates';
 import { VNode } from './VNodes/VNode';
-import { Format } from '../../utils/src/Format';
 import { VRange } from './VRange';
-import { CharNode } from './VNodes/CharNode';
 import { RelativePosition } from '../../utils/src/range';
+import { HTMLRendering } from './BasicHtmlRenderingEngine';
```
Zinston: Well this _is_ actually an HTML renderer to be exact.

packages/core/src/Renderer.ts
```diff
@@ -222,12 +131,24 @@ export class Renderer {
...
+            rendering.vNodes,
+            context.parentNode,
+        );
+        context.currentVNode = rendering.vNodes[rendering.vNodes.length - 1];
         return context;
     }
+    // Append the rendered nodes to the DOM and map them.
+    _addRenderedNodes(domNodes: Node[], vNodes: VNode[], parent: Node): void {
+        domNodes.forEach(renderedNode => {
+            vNodes.forEach((vNode, index) => {
```
Zinston: O(n<sup>2</sup>) * recursion. Though granted `n` in this case should remain quite small.
This is solved if `render` returns a rendering context.

packages/core/src/Renderer.ts
```diff
@@ -222,12 +131,24 @@ export class Renderer {
...
      * @param context
      */
-    _renderVNode(context: RenderingContext): RenderingContext {
-        if (isChar(context.currentVNode)) {
-            context = this._renderTextNode(context);
-        } else {
-            context = this._renderElement(context);
-        }
+    _renderNext(context: RenderingContext): RenderingContext {
+        const rendering = context.currentVNode.render<HTMLRendering>('html');
```
Zinston: This is not a great return type. How about a rendering context?

packages/core/src/VDocument.ts
```diff
@@ -1,10 +1,11 @@
 import { VNode } from './VNodes/VNode';
 import { VRange } from './VRange';
-import { CharNode, FormatType, FORMAT_TYPES } from './VNodes/CharNode';
+import { CharNode } from '../../plugin-char/src/VNodes/CharNode';
```
Zinston: WIP! Nothing in this file is expected to work for the moment. The main question is how to make this plugin-agnostic.
For example I thought of using `execCommand` to let `Char` create the `CharNodes` but `execCommand` doesn't return anything.

packages/core/src/VDocumentMap.ts
```diff
@@ -1,4 +1,4 @@
-import { Format } from '../../utils/src/Format';
+import { formatMap } from '../../plugin-format/src/formatMap';
```
Zinston: `VDocumentMap` should not be aware of `formatMap`, given that it's optional.

packages/core/src/VDocumentMap.ts
```diff
@@ -62,7 +62,7 @@ export const VDocumentMap = {
         // Only if element is not a format and not already in the map to prevent
         // overriding a VNode if it is representing by multiple Nodes. Only the
         // first Node is mapped to the VNode.
-        if (!Format.tags.includes(domNode.nodeName) && !toDom.has(vNode)) {
+        if (!formatMap.tags.includes(domNode.nodeName) && !toDom.has(vNode)) {
```
Zinston: Same as above.

packages/core/src/VNodes/Attribute.ts
```diff
@@ -0,0 +1,21 @@
...
+    }
+    render(rendered: Node[]): Node[] {
+        return rendered.map(node => {
+            if (this.value && node instanceof Element) {
+                node.setAttribute(this.name, this.value);
+            }
+            return node;
+        });
+    }
+    copy(): Attribute {
```
Zinston: `clone`?

packages/core/src/VNodes/VNode.ts
```diff
@@ -15,21 +16,20 @@ let id = 0;
 
 export class VNode {
     static readonly atomic: boolean = false;
-    readonly type: VNodeType;
+    readonly type: VNodeType = VNodeType.NODE;
```
Zinston: typing not needed

packages/core/src/VNodes/VNode.ts
```diff
@@ -15,21 +16,20 @@ let id = 0;
...
     readonly id = id;
     parent: VNode;
     renderingEngines: Record<string, RenderingEngine> = {
         html: BasicHtmlRenderingEngine,
     };
-    name: string;
+    name: string = this.constructor.name;
```
Zinston: typing not needed

packages/core/src/VNodes/VNode.ts
```diff
@@ -57,12 +57,18 @@ export class VNode {
     get atomic(): boolean {
         return (this.constructor as typeof VNode).atomic;
     }
+    get attributesObject(): { [key: string]: AttributeValue } {
```
Zinston: This is a crappy idea, forget it. It was just for the DevTools to display the attributes properly because there still is that Owl bug that prevents us from using `get` on the map but there has to be a better way anyway.

packages/core/test/Parser.test.ts
```diff
@@ -1,85 +1,85 @@
-import { expect } from 'chai';
-import { CharNode } from '../src/VNodes/CharNode';
-import { Parser } from '../src/Parser';
-import { FragmentNode } from '../src/VNodes/FragmentNode';
-import { LineBreakNode } from '../src/VNodes/LineBreakNode';
-import { VElement } from '../src/VNodes/VElement';
+// import { expect } from 'chai';
```
Zinston: TODO: restore these tests.

packages/core/src/VRange.ts
```diff
@@ -1,6 +1,6 @@
 import { VNode } from './VNodes/VNode';
 import { RelativePosition, Direction, withMarkers } from '../../utils/src/range';
-import { RangeNode } from './VNodes/RangeNode';
+import { RangeNode } from '../../plugin-range/src/VNodes/RangeNode';
```
Zinston: `VRange` should not be aware of `RangeNode` given that it's optional.

packages/core/test/VNodes.test.ts
```diff
@@ -1,1169 +1,1169 @@
...
-import { VNode, VNodeType } from '../src/VNodes/VNode';
-import { CharNode } from '../src/VNodes/CharNode';
-import { LineBreakNode } from '../src/VNodes/LineBreakNode';
-import { RangeNode } from '../src/VNodes/RangeNode';
-import { VElement } from '../src/VNodes/VElement';
-import { FragmentNode } from '../src/VNodes/FragmentNode';
-import { withMarkers } from '../../utils/src/range';
-import { JWPlugin } from '../src/JWPlugin';
-import JWEditor from '../src/JWEditor';
+// import { expect } from 'chai';
```
Zinston: TODO: restore these tests.

packages/plugin-char/src/VNodes/CharNode.ts
```diff
@@ -0,0 +1,144 @@
...
+    //--------------------------------------------------------------------------
+    // Private
+    //--------------------------------------------------------------------------
+
+    /**
+     * Return true if this VNode has the same format properties as `b`.
+     *
+     * @param b
+     */
+    _isSameAs(b: VNode): boolean {
```
Zinston: Will this still work?

packages/plugin-devtools/assets/DevTools.xml
```diff
@@ -54,9 +54,9 @@
             <span t-elif="props.vNode.char" t-on-click="onClickNode"
                 t-on-dblclick="onDblClickNode"
                 class="selectable-line" t-att-class="{
-                    bold: props.vNode.format and props.vNode.format.bold,
-                    italic: props.vNode.format and props.vNode.format.italic,
-                    underline: props.vNode.format and props.vNode.format.underline,
+                    bold: props.vNode.attributesObject.bold,
```
Zinston: See comment above. Dump this.

packages/plugin-devtools/src/components/InfoComponent.ts
```diff
@@ -66,7 +66,7 @@ export class InfoComponent extends OwlUIComponent<{}> {
                         value = this._objectRepr(vNode[key]);
                     }
                 } else if (typeof vNode[key] === 'string') {
-                    value = '"' + value + '"';
+                    value = '"' + vNode[key] + '"';
```
Zinston: :eyes:

packages/plugin-devtools/src/components/PathComponent.ts
```diff
@@ -1,23 +1,9 @@
...
-            }
-            if (vNode.format.italic) {
-                format += '.i';
-            }
-            if (vNode.format.underline) {
-                format += '.u';
-            }
-        }
-        return name + format;
+        return vNode.name.toLowerCase();
```
Zinston: TODO: restore this, in an agnostic way...

packages/plugin-devtools/src/components/TreeComponent.ts
```diff
@@ -3,7 +3,7 @@ import { OwlUIComponent } from '../../../owl-ui/src/OwlUIComponent';
 import { VNode } from '../../../core/src/VNodes/VNode';
 import { VRangeDescription } from '../../../core/src/VRange';
 import { Direction, RANGE_TAIL_CHAR, RANGE_HEAD_CHAR } from '../../../utils/src/range';
-import { LineBreakNode } from '../../../core/src/VNodes/LineBreakNode';
+import { LineBreakNode } from '../../../plugin-linebreak/src/VNodes/LineBreakNode';
```
Zinston: `TreeComponent` should not be aware of `LineBreakNode` given that it's optional.

packages/plugin-linebreak/src/VNodes/LineBreakNode.ts
```diff
@@ -25,10 +21,10 @@ export class LineBreakNode extends VElement {
      */
     render<T>(to = 'html'): T {
         const t = this.renderingEngines[to].render(this) as T;
-        if (to === 'html' && !this.nextSibling() && t instanceof DocumentFragment) {
+        if (to === 'html' && !this.nextSibling()) {
             // If a LINE_BREAK has no next sibling, it must be rendered as two
             // BRs in order for it to be visible.
-            t.appendChild(document.createElement('br'));
+            ((t as unknown) as HTMLRendering).fragment.appendChild(document.createElement('br'));
```
Zinston: :eyes:

packages/plugin-youtube/src/VNodes/YoutubeNode.ts
```diff
@@ -56,8 +39,8 @@ export class YoutubeNode extends VNode {
      */
     render<T>(to = 'html'): T {
         const t = this.renderingEngines[to].render(this) as T;
-        if (to === 'html' && t instanceof DocumentFragment) {
-            const iframe = t.firstChild as Element;
+        if (to === 'html') {
+            const iframe = ((t as unknown) as HTMLRendering).fragment.firstChild as Element;
```
Zinston: :eyes:

packages/plugin-youtube/src/Youtube.ts
```diff
@@ -3,4 +3,22 @@ import { YoutubeNode } from './VNodes/YoutubeNode';
...
+                        width: element.getAttribute('width'),
+                        height: element.getAttribute('height'),
+                        frameborder: element.getAttribute('frameborder'),
+                        allow: element.getAttribute('allow'),
+                        allowfullscreen: element.getAttribute('allowfullscreen') !== null,
+                    }),
+                ];
+            }
+        }
+        return null;
```
Zinston: unnecessary

packages/utils/src/Predicates.ts
```diff
@@ -13,4 +13,4 @@ export function not(predicate: Predicate): Predicate {
 
 export { isLeaf } from '../../core/src/VNodes/VNode';
 export { isMarker } from '../../core/src/VNodes/VNode';
-export { isChar } from '../../core/src/VNodes/CharNode';
+export { isChar } from '../../plugin-char/src/VNodes/CharNode';
```
Zinston: nope

packages/core/src/VRange.ts
Zinston: Actually shouldn't `RangeNode` stand among `VElement` and `FragmentNode` as a generic node type? That is, not in a plugin. It's not exactly optional, is it?

packages/plugin-char/src/VNodes/CharNode.ts
Zinston: It seems to but `<i>’Twas brillig, <b>and</b> the slithy toves</i>` doesn't parse properly: we get `<i>’Twas brillig, </i><b>and</b> the slithy toves` instead.

examples/index/index.ts
```diff
@@ -1,15 +1,16 @@
...
+import { MarkdownRenderer } from '../../packages/plugin-markdownrenderer/src/MarkdownRenderer';
 import { Youtube } from '../../packages/plugin-youtube/src/Youtube';
 import './index.css';
 
 jabberwocky.init(basicEditor.editable);
 
 basicEditor.loadConfig({
     debug: true,
-    // renderTo: 'markdown',
+    renderTo: 'markdown',
```
Zinston: @dmo-odoo
I'm facing a conundrum.

1. nodes are not aware of renderers
2. renderers are not aware of nodes
3. nodes render themselves
4. renderers handle generic steps of rendering (node creation, mapping, attributes)

-> Where do we create links?

Eg:
A markdown renderer must render Formats in a special way, therefore it must be
aware and able to extend anything that handles formats. But we could have
different plugins handling formats.

------
Also, do they all render ultimately to html to display in the DOM? Albeit
differently?

Eg:
VDocument
```html
<paragraph>
    <Char.b:a/>
</paragraph>
<paragraph>
    <Char:b/>
</paragraph>

```
HTMLRenderer
```html
<p><b>a</b></p>
<p>b</p>
```
MarkdownRenderer
```html
<p><b>*a*</b></p>
<p>b</p>
```
or
```html
<div>*a*<br>b</div>
```
...or 
```javascript
"*a*\nb"
```
?

Zinston:
@dmo-odoo Two things I find a bit to be a pity are that (1) now that VNodeTypes are just 'node', 'marker' or 'fragment', we only see that in the title portion of the info panel in DevTools ; and that (2) now that the node's name is mostly just its DOM nodeName, we see just that in DevTools ('H1' instead of 'heading1'), which makes it non-obvious that it's actually been parsed and it didn't just use the default generic parsing.
Also we gotta do something about that shitty `toString` I made. In hindsight, there's no need for it: I'd rather it just returned `this.name`.

PR #62, opened 2019-12-20
WIP Agnostic parser/renderer

Zinston:


Zinston:
@dmo-odoo Not easy to pick this back up... I rebase on master, which is a start (and not a minor one!). The problem we're trying to solve is complex and important. This PR introduces some key ideas to solve it but does not yet achieve it. In the meantime, the situation (and therefore the problem) has changed since the merge of node classes.

dmo-odoo:
There are now two open PRs for this feature. Could you please keep only one open and archive the other by closing it? Thanks!

PR #65, opened 2019-12-24
Replace Range by Selection

Gorash:
In our implementation, we can calculate very quickly if anchor is before
focus. Removing start, end, startContainer... removes unnecessary
complexity. In addition we can use the dom selection instead of looking
for the range (and its direction) in the dom selection.

The methods all use the anchor if the range is collapsed. If it is not, as
for the 'deleteSelection', it is necessary to call on the direction to know
(in the contrary case, for a backward selection, the range can remain in
the node which is removed from the arch.
We make sure that the range is calapsed in forward after removing the
nodes.


dmo-odoo:
I'm closing this PR in favor of https://github.com/odoo-dev/jabberwock/pull/70

PR #67, opened 2019-12-24
[ADD] Core: handle basic lists

Zinston:
This PR introduces `ListNodes`, which are parsed from `OL` and `UL`.

All currently relevant list tests from we3 were also converted and extended for consistency.

A few peculiarities about our treatment of lists:

##### (1) The virtual document abstracts away the concept of list item (`LI`).
To make this possible:
- each block in a list is rendered as a list item in the DOM
- each list item that contains text without wrapping (eg.: `<li>a</li>`) is wrapped in a `VElement: P`

This allows us to minimize special cases and handle virtually everything
that goes on within lists in the same basic way as outside of lists.

##### (2) Paragraph elements (`P`) within list items are not rendered as such.
As a consequence of (1).
eg: `<li><p>a</p></li>` will be rendered as `<li>a</li>`

These are meaningless and error prone so we choose to ignore them, which
allows us to simplify the treatment of lists. This could change in the
future but it's no priority.

##### _Note: a lot of the logic in this should move or be removed once node behaviors can be handled via plugins._

<hr>

##### To do later
- handle indent (for indent PR or when parser and renderer are agnostic)
- handle formats (for format PR)
- when parser and renderer are agnostic, move most of this to list plugin


examples/utils/jabberwocky.xml
```diff
@@ -1,8 +1,25 @@
...
+                <ul>
+                    <li>a.d.a</li>
+                </ul>
+            </li>
+        </ul>
+    </li>
     <li>b</li>
+    <li>
+        <ol>
+            <li>b.1</li>
```
Zinston: Backspace at the beginning of this line doesn't work: we get
```html
<paragraph>
    b<paragraph>b.1</paragraph> ...
</paragraph>
```
which is then rendered as
```html
<li>
    b<p>b.1</p> ...
</li>
```

packages/core/src/Parser.ts
```diff
@@ -288,6 +290,36 @@ export class Parser {
                 return;
             }
         }
+        // Adapt the context for the parsing of a list
```
Zinston: This would be in `ListPlugin`'s parsing hook.

Zinston:
todo:
- insert/convert-to ul/ol with ctrl+shift+7/8
- fix backspace nested lists
- refactor, reorganize
- handle indent
- handle formats (for format PR)
- when parser and renderer are agnostic, move most of this to list plugin

Zinston:
@dmo-odoo Ready :-)

packages/core/test/VDocument.test.ts
```diff
@@ -615,6 +615,177 @@ describe('stores', () => {
                         });
                     });
                 });
+                describe('Lists', () => {
```
dmo-odoo: I'm so looking forward to when these tests will be moved to their respective plugins rather than cluttering VDocument... :)

packages/core/test/Renderer.test.ts
```diff
@@ -77,5 +79,144 @@ describe('utils', () => {
...
+                const renderer = new Renderer();
+                renderer.render(editor.vDocument, editor.editable);
+                /* eslint-disable prettier/prettier */
+                expect(editor.editable.innerHTML).to.equal([
+                    '<ul>',
+                        '<li>a</li>',
+                        '<li>',
+                            '<ul>',
+                                '<li>a.a</li>',
+                                '<li>a.b</li>',
```
dmo-odoo: My prior question about P remains, since this example introduces no p. What about a list item containing two P elements ?

packages/core/src/CorePlugin.ts
```diff
@@ -44,6 +47,9 @@ export class CorePlugin extends JWPlugin {
         applyFormat: {
             handler: this.applyFormat.bind(this),
         },
+        list: {
```
dmo-odoo: I guess you already thought about using `createList` but chose to use `list` instead, I'm looking forward to knowing why :)

packages/core/src/EventManager.ts
```diff
@@ -69,6 +69,22 @@ export class EventManager {
...
+                ) {
+                    return this.editor.execCommand('list', { type: 'ordered' });
+                } else if (
+                    payload.ctrlKey &&
+                    !payload.altKey &&
+                    payload.shiftKey &&
+                    !payload.metaKey &&
+                    payload.key === '8'
+                ) {
+                    return this.editor.execCommand('list', { type: 'unordered' });
```
dmo-odoo: I would add `TODO: remove this when we have a toolbar.` Because these are not "standardly recognized" shortcuts.

packages/core/src/VDocument.ts
```diff
@@ -156,4 +157,57 @@ export class VDocument {
             }
         }
     }
+    /**
+     * Insert a list at range.
+     *
+     * @param type
+     */
+    list(type: 'ordered' | 'unordered'): void {
```
dmo-odoo: I *think* I would have created an enum for the type to avoid having to rewrite the possible types at multiple places, like you had to do in the `ListNode` definition and here.

packages/core/src/VDocument.ts
```diff
@@ -156,4 +157,57 @@ export class VDocument {
...
+        let selectedBlocks: Set<VNode>;
+        if (this.range.isCollapsed()) {
+            selectedBlocks = new Set([this.range.start.parent]);
+        } else {
+            selectedBlocks = new Set(
+                this.range.selectedNodes.map(node => {
+                    return node.atomic ? node.parent : node;
```
dmo-odoo: I'm not sure I understand why this distinction must be made.

Zinston: It's wrong. This doesn't handle nested blocks within list items. I'll rewrite it.

packages/core/src/VNodes/VNode.ts
```diff
@@ -183,6 +183,13 @@ export class VNode {
     // Browsing
     //--------------------------------------------------------------------------
 
+    ancestor(predicate: Predicate): VNode {
```
Zinston: docstring

packages/core/src/VRange.ts
```diff
@@ -87,6 +87,18 @@ export class VRange {
         });
         return selectedNodes;
     }
+    get selectedParents(): VNode[] {
```
Zinston: Get rid of this.

packages/core/test/VDocument.test.ts
```diff
@@ -285,6 +290,352 @@ describe('stores', () => {
                 });
             });
         });
+        describe.only('toggleList', () => {
```
Zinston: Remove "only".

packages/core/test/VDocument.test.ts
```diff
@@ -285,6 +290,352 @@ describe('stores', () => {
...
+                            await testEditor({
+                                contentBefore: '<p>ab</p><ul><li>cd</li><li>ef[]gh</li></ul>',
+                                stepFunction: toggleUnorderedList,
+                                contentAfter: '<p>ab</p><ul><li>cd</li></ul><p>ef[]gh</p>',
+                            });
+                        });
+                    });
+                });
+                describe('Ordered', () => {
+                    describe('Insert', () => {
```
Zinston: Add missing tests from Unordered.

packages/core/test/VDocument.test.ts
```diff
@@ -2,12 +2,17 @@ import JWEditor from '../src/JWEditor';
 import { testEditor } from '../../utils/src/testUtils';
```
Zinston: Add:

- `<div><p>a[]bc</p></div>`
- `<div><p>abc</p><p>d[]ef</p></div>`
- `<div><p>a[bc</p></div><p>d]ef</p>`
- `<ul><li>abc</li><li>def<ul><li>ghi</li></ul></li><li>jkl</li></ul>` and bigger nests

And check from we3's `TestList.js`:
- ~~`ulTests`~~ (done)
- `olTests`
- `checklistTests` (for when we implement checklists)
- `conversionTests`
- `indentTests`
- `outdentTests`

packages/core/test/VDocument.test.ts
```diff
@@ -2,12 +2,17 @@ import JWEditor from '../src/JWEditor';
...
 import { FormatParams } from '../src/CorePlugin';
 import { LineBreakNode } from '../src/VNodes/LineBreakNode';
+import { ListType } from '../src/VNodes/ListNode';
```
Zinston: Add:

- `<div><p>a[]bc</p></div>`
- `<div><p>abc</p><p>d[]ef</p></div>`
- `<div><p>a[bc</p></div><p>d]ef</p>`
- `<ul><li>abc</li><li>def<ul><li>ghi</li></ul></li><li>jkl</li></ul>` and bigger nests

And check from we3's `TestList.js`:
- ~~`ulTests`~~ (done)
- `olTests`
- `checklistTests` (for when we implement checklists)
- `conversionTests`
- `indentTests`
- `outdentTests`

packages/core/src/CorePlugin.ts
```diff
@@ -1,7 +1,7 @@
 import { JWPlugin } from './JWPlugin';
 import JWEditor from './JWEditor';
 import { VRangeDescription } from './VRange';
-import { Direction, RelativePosition } from '../../utils/src/range';
+import { RelativePosition, Direction } from '../../utils/src/range';
```
Zinston: @dmo-odoo This is your commit. Could you please write a message and put yourself as author in it? Thanks.

packages/core/src/VNodes/ListNode.ts
```diff
@@ -1,4 +1,5 @@
 import { VElement } from './VElement';
+import { VNode } from './VNode';
```
Zinston: Same here, if you have anything particular to change in the commit message, given that you authored this version, please proceed :-)
Also, obviously, put yourself as author...

Zinston:
@dmo-odoo Ready bis :-)

packages/core/src/Parser.ts
```diff
@@ -288,8 +290,76 @@ export class Parser {
...
+                VDocumentMap.set(paragraph, nextContext.node);
+                nextContext.parentVNode.append(paragraph);
+                nextContext.parentVNode = paragraph;
+                nextContext.node = adjacentInlineChildren[0];
+                if (adjacentInlineChildren.length === 1 && nextContext.node.nodeName === 'BR') {
+                    VDocumentMap.set(paragraph, nextContext.node);
+                    return this._nextParsingContext(nextContext);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
Zinston: What if the child is a `H1`? Then we need to create a `H1` instead of a `P`.

packages/core/src/Parser.ts
```diff
@@ -288,8 +290,76 @@ export class Parser {
...
+            // new parentVNode to move up one level (<ol><li>a<ul> where ul is
+            // what we're about to parse, ol is the new parentVNode, and the li
+            // is essentially what we have as parentVNode before correction).
+            nextContext.parentVNode = nextContext.parentVNode.parent;
+        } else if (nextContext.node.nodeName === 'LI') {
+            // Parse a list item.
+            if (this._isIndentedList(nextContext.node.parentNode)) {
+                // We are parsing the children of an indented list, we need to
+                // correct the new parentVNode to be that indented list.
+                nextContext.parentVNode = nextContext.parentVNode.parent.lastChild();
```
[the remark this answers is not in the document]
Zinston: Tbh I don't remember the detail... I did check and see that it _is_ indeed necessary, and that to be more complete and not do useless reassignation, I could have added the following check: `&& !nextContext.parentVNode.is(ListNode)`.

Here's what I think... I believe this was meant to handle `<ul><li>abc<ol><li>def</li></ol></li></ul>`:
1) ul -> ListNode:UL into root
2) li -> VElement:P into ListNode:UL + CharNodes:abc into VElement:P
3) ol -> ListNode:OL into ListNode:UL
4) li -> VElement:P into ListNode:OL (...)
At step 4, we need the P to be inserted into the OL. But the regular workflow for blocks within a list is to append them to the list so our default parent is the UL. This is us starting a new list context basically.
Something like that... :sweat_smile:

packages/core/src/Parser.ts
```diff
@@ -288,8 +290,76 @@ export class Parser {
...
+    /**
+     * Return true if the given DOM node is an indented list node.
+     *
+     * @param node
+     */
+    _isIndentedList(node: Node): boolean {
+        return (
+            this._isList(node) &&
+            node.parentNode &&
+            (node.parentNode.nodeName === 'LI' || this._isList(node.parentNode))
```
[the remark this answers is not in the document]
Zinston: No `LI` is not among the permitted contents of a `LI`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li

packages/core/src/VNodes/ListNode.ts
```diff
@@ -0,0 +1,18 @@
+import { VElement } from './VElement';
+
+export enum ListType {
+    ORDERED = 'ordered',
+    UNORDERED = 'unordered',
+}
```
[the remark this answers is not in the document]
Zinston: No. I just think it's nice ^^ That way DevTools can render something nicer than 0 and 1...

packages/core/src/VNodes/ListNode.ts
Zinston: Naive rebasing... Sorry!

packages/core/src/VDocument.ts
```diff
@@ -152,4 +153,199 @@ export class VDocument {
...
+
+    /**
+     * Convert the given nodes to a list with the given `htmlTag`.
+     *
+     * @param nodes
+     * @param htmlTag
+     */
+    _convertToList(nodes: Array<VNode>, htmlTag: string): void {
+        // Lists cannot have atomic nodes as direct children.
+        nodes = Array.from(new Set(nodes.map(node => (node.atomic ? node.parent : node))));
```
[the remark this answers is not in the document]
Zinston: You're talking HTML and this code is VDocument.
HTML: `<ul><li>a<br/>[b]<br/>c</li></ul>`
VDocument: `ListNode:UL > VElement:P > aLineBreak[b]LineBreak[c]`
All the atomics are children of the VElement:P, not of the list. It's non-atomic elements within lists that allow us the discriminate when rendering to `LI`s.

Were you thrown off by the name `nodes`? I _could_ rename to `vNodes` if that can avoid this kind of confusion... :smiling_imp:

packages/core/src/VDocument.ts
```diff
@@ -152,4 +153,199 @@ export class VDocument {
...
+            if (!lastList.hasChildren()) {
+                lastList.remove();
+            }
+        }
+
+        // Create and insert the list before the common ancestor to the first
+        // last node, unless the first node is within a list, in which case
+        // insert before the list instead to avoid inserting the new list within
+        // a list.
+        let list = new ListNode(htmlTag);
```
Zinston: This is where I need `htmlTag`. Although granted it might make more sense if `ListNode` would take a `ListType` instead...

packages/core/src/VDocument.ts
```diff
@@ -152,4 +153,199 @@ export class VDocument {
...
+        let selectedNodes: Array<VNode>;
+        if (this.range.isCollapsed()) {
+            // Toggle the parent if the range is collapsed.
+            selectedNodes = [this.range.start.parent];
+        } else {
+            selectedNodes = this.range.selectedNodes;
+        }
+
+        // Check if all selected nodes are within a list of the same type.
+        const htmlTag = type === ListType.ORDERED ? 'OL' : 'UL';
```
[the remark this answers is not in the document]
Zinston: I think you're right, I don't need it for the check just below. I do need it in `_convertToList` though, but that might be bad design. See comment below.

packages/core/src/VDocument.ts
```diff
@@ -152,4 +153,199 @@ export class VDocument {
...
+        // lists.
+        const duplicatedNodes = [];
+        const first = nodes[0];
+        const firstList = this._listAncestor(first);
+        if (firstList) {
+            let child = first;
+            first.ancestor(ancestor => {
+                duplicatedNodes.push(ancestor.splitAt(child));
+                child = ancestor;
+                return ancestor === firstList;
```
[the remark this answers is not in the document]
Zinston: `_recursiveSplitUpToTheFirstAncestorThatIsAList`
Thanks for the tip!

packages/core/src/VDocument.ts
```diff
@@ -152,4 +153,199 @@ export class VDocument {
...
+                child = ancestor;
+                return ancestor === lastList;
+            });
+            if (!lastList.hasChildren()) {
+                lastList.remove();
+            }
+        }
+
+        // Create and insert the list before the common ancestor to the first
+        // last node, unless the first node is within a list, in which case
```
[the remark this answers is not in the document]
Zinston: Oops. "first _and_ last".

packages/core/src/VDocument.ts
```diff
@@ -152,4 +153,199 @@ export class VDocument {
...
+                        childOfList = node;
+                    } else {
+                        childOfList = node.ancestor(ancestor => {
+                            return this._isList(ancestor.parent);
+                        });
+                    }
+                    return childOfList;
+                }),
+            ),
+        );
```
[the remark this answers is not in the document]
Zinston: `distinct` == `_.uniq` == `$.unique`? I'm in favor!

packages/core/src/VDocument.ts
```diff
@@ -152,4 +153,199 @@ export class VDocument {
...
+
+        // Split the lists and move their contents to move out of them.
```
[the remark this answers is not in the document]
Zinston: Omg. Well. `and move their appropriate contents out of them`? What's the word I'm looking for? Relevant? Appropriate? Involved?

packages/core/src/VNodes/ListNode.ts
```diff
@@ -16,6 +15,22 @@ export class ListNode extends VElement {
...
+    // Lifecycle
+    //--------------------------------------------------------------------------
+
+    /**
+     * Return a new VNode with the same type and attributes as this VNode.
+     *
+     *  @override
+     */
+    shallowDuplicate(): ListNode {
+        return new ListNode(this.htmlTag);
```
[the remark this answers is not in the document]
Zinston: Well it is, sort of... That's an annoying thing with the current implementation: VNode has to return `new VNode()`, CharNode has to return `new CharNode(this.char, this.format)`, etc.

packages/core/src/Parser.ts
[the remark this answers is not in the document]
dmo-odoo: Well I was paraphrasing, you can check if `LI`'s child is a text node and, if yes create a `P`, if no then do not parse anything and let the parser handle the `P`, it will put it inside the `ListNode:UL` at that point.

packages/core/src/VNodes/ListNode.ts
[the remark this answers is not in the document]
dmo-odoo: Alright

packages/core/src/VNodes/ListNode.ts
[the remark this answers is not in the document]
dmo-odoo: Oh yeah right... I'll see if I can find a way of having it do something similar to 'new typeof this' :p

packages/core/src/VDocument.ts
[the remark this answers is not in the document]
dmo-odoo: I  have a slight preference for naming it `distinct` because I think the wording is more obvious but I'm open to discussions.

packages/core/src/VDocument.ts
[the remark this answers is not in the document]
dmo-odoo: I see what you did there... 👀 
Though HTML or VDocument is irrelevant to what I was saying. The following VDocument snippet still exhibit the issue I was referring to:
```
<VElement:P><CharNode:a/><LineBreakNode/>[<CharNode:b>]<LineBreakNode/><CharNode:c/></VElement:P>
```

packages/core/src/VDocument.ts
```diff
@@ -152,4 +153,199 @@ export class VDocument {
...
+
+        // Move the nodes to the list
+        withMarkers(() => {
+            // Remove the children of nested lists (they will be moved together
+            // with their parent) and convert nested lists to the new type.
+            const nestedLists = nodes.filter(this._isNestedList.bind(this));
+            nodes = nodes.filter(node => {
+                return !node.ancestor(ancestor => nestedLists.includes(ancestor));
+            });
+            nestedLists.forEach(nestedList => ((nestedList as ListNode).htmlTag = htmlTag));
```
dmo-odoo: Aren't you supposed to change `listType` ? You might want to redefine `htmlTag` as a getter on `listType` or the opposite to keep them synchronized.

packages/core/src/VDocument.ts
```diff
@@ -152,4 +153,199 @@ export class VDocument {
...
+        }
+
+        // Move the nodes to the list
+        withMarkers(() => {
+            // Remove the children of nested lists (they will be moved together
+            // with their parent) and convert nested lists to the new type.
+            const nestedLists = nodes.filter(this._isNestedList.bind(this));
+            nodes = nodes.filter(node => {
+                return !node.ancestor(ancestor => nestedLists.includes(ancestor));
+            });
```
dmo-odoo: I think this might deserve a new variable name.

packages/core/src/VDocument.ts
```diff
@@ -152,4 +153,199 @@ export class VDocument {
...
+                }
+            });
+        });
+
+        // If the new list is after or before a list of the same type, merge
+        // them (eg: <ol><li>a</li></ol><ol><li>b</li></ol> =>
+        // <ol><li>a</li><li>b</li></ol>).
+        const previousSibling = list.previousSibling();
+        if (this._isList(previousSibling) && (previousSibling as ListNode).htmlTag === htmlTag) {
+            list.children.slice().forEach(child => previousSibling.append(child));
```
dmo-odoo: FYI on my commonAncestor branch, children is always a filtered copy so no more need to slice. You basically can't alter a node's children by playing with the array manually, which was already the case outside of `withMarkers` but the behavior was inconsistent.

packages/core/src/VDocument.ts
```diff
@@ -152,4 +153,199 @@ export class VDocument {
...
+            list.remove();
```
dmo-odoo: I'm wondering if we might get rid of `withMarkers` by calling `mergeWith` enough time to empty the list (since `mergeWith` uses `withMarkers` internally.)

packages/core/src/VDocument.ts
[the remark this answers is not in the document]
Zinston: I don't get it. In your example, we indeed need to take the VElement:P.
LineBreakNodes are atomic, if that's the info you're missing.

Wait... Do you mean you're wondering if we should split at the line breaks ? If so, no we shouldn't.

packages/core/src/VNodes/ListNode.ts
```diff
@@ -16,6 +15,22 @@ export class ListNode extends VElement {
             return [new ListNode(node.nodeName)];
         }
     }
+    get listType(): ListType {
```
Zinston: See ? Getter :-D

packages/core/src/VDocument.ts
Zinston: ListType _is_ a getter so they remain in sync. Now as I said above I think instantiating lists with listType instead of htmlTag might make more sense so I did that locally yesterday evening. Then I had to overload htmlTag as a getter to keep in sync instead of the opposite. This line is thus now `nestedLists.forEach(nestedList => ((nestedList as ListNode).listType = type));`

packages/core/src/VDocument.ts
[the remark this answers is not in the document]
Zinston: No I like `distinct`. I never was a fan of `uniq` and indeed "dinstinct" is more semantically accurate than "unique".

packages/core/src/VNodes/ListNode.ts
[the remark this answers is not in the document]
Zinston: That's a job for... DMO!

packages/core/src/Parser.ts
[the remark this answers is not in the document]
Zinston: Not enough. I would have to test for every inline or every block.
I need to create a P if the first child of the LI is a text node _or any other inline node_, or a BR.
I guess I technically _could_ do that, but is it better?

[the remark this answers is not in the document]
Zinston: Ok so if we do that, I got this:
```typescript
// Adapt the context for the parsing of a list
if (
this._isIndentedList(nextContext.node) &&
!(nextContext.parentVNode instanceof ListNode)
) {
// We're about to parse an indented list, we may need to correct the
// new parentVNode to move up to the parent list (<ol><li>a<ul>...,
// where ul is what we're about to parse, ol is the new parentVNode,
// and the li is essentially what we have as parentVNode before
// correction).
nextContext.parentVNode = nextContext.parentVNode.ancestor(node => node.is(ListNode));
} else if (nextContext.node.nodeName === 'LI') {
// Parse a list item.
if (this._isIndentedList(nextContext.node.parentNode)) {
    // We are parsing the children of an indented list, we need to
    // correct the new parentVNode to be that indented list.
    nextContext.parentVNode = nextContext.parentVNode.parent.lastChild();
}
const children = Array.from(nextContext.node.childNodes);
// A list item with no children should be skipped.
if (!children.length) {
    return this._nextParsingContext(nextContext);
}
// An empty text node as first child should be skipped.
if (this._isEmptyTextNode(children[0])) {
    return this._nextParsingContext({ ...nextContext, ...{ node: children[0] } });
}
// A list item containing only a BR should be replaced with an
// empty paragraph.
if (children.length === 1 && children[0].nodeName === 'BR') {
    const paragraph = new VElement('P');
    nextContext.parentVNode.append(paragraph);
    VDocumentMap.set(paragraph, nextContext.node);
    VDocumentMap.set(paragraph, children[0]);
    return this._nextParsingContext({ ...nextContext, ...{ node: children[0] } });
}
// Inline elements in a list item should be wrapped in a paragraph.
if (!utils.isBlock(children[0]) || children[0].nodeName === 'BR') {
    const paragraph = new VElement('P');
    nextContext.parentVNode.append(paragraph);
    nextContext.parentVNode = paragraph;
    VDocumentMap.set(paragraph, nextContext.node);
}
// Now we can move on to the list item's contents, to be added to
// the paragraph created above, or to the list itself in the case of
// blocks.
nextContext.node = children[0];
```
It passes the tests.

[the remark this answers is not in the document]
Zinston: OMG apparently with this I can remove the indented list check above! No idea why it was there or why this fixes but well... good!

packages/core/src/Parser.ts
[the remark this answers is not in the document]
Zinston: See below. If the solution is approved we can resolve this.

packages/core/src/VDocument.ts
```diff
@@ -152,4 +153,199 @@ export class VDocument {
...
+                lastList.remove();
+            }
+        }
+
+        // Create and insert the list before the common ancestor to the first
+        // last node, unless the first node is within a list, in which case
+        // insert before the list instead to avoid inserting the new list within
+        // a list.
+        let list = new ListNode(htmlTag);
+        const listAncestor = this._listAncestor(nodes[0]);
```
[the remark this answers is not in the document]
Zinston: You can't use `firstListAncestor.before(list)`: `firstListAncestor` was split, `first` is now in the duplicate.
So we'd have:
```typescript
// If the last node is within a list, we need to split that list.
const last = nodes[nodes.length - 1];
const lastListAncestor = this._listAncestor(last);
if (lastListAncestor) {
    const duplicatedAncestors = this._splitUpToAncestor(last, lastListAncestor);
    duplicatedNodes = [...duplicatedNodes, ...duplicatedAncestors];
}

// If the first node is within a list, we need to split that list.
// Then insert the new list to which the nodes will be appended.
const first = nodes[0];
const firstListAncestor = this._listAncestor(first);
if (firstListAncestor) {
    const duplicatedAncestors = this._splitUpToAncestor(first, lastListAncestor);
    duplicatedNodes = [...duplicatedNodes, ...duplicatedAncestors];

    // Insert the new list before the first node's list ancestor.
    // It was split so we need its duplicate as that is where the node is.
    const newListAncestor = duplicatedAncestors[duplicatedAncestors.length - 1];
    newListAncestor.before(newList);
} else {
    // Insert the new list before the common ancestor to the first and
    // last nodes.
```
with
```typescript
/**
 * Split a node and its ancestors up until the given ancestor.
 *
 * @param node
 * @returns a list of the clones generated during the splits.
 */
_splitUpToAncestor(node: VNode, ancestor: VNode): VNode[] {
    const duplicatedNodes = [];
    let child = node;
    let parent = child.parent;
    do {
        duplicatedNodes.push(parent.splitAt(child));
        child = parent;
        parent = child.parent;
    } while (child !== ancestor);
    if (!ancestor.hasChildren()) {
        ancestor.remove();
    }
    return duplicatedNodes;
}
```
OK?

packages/core/src/VDocument.ts
Zinston: Good question. Is it really worth investigating though?

Zinston:
@dmo-odoo Branch is ready again. There are changes I kept in a separate commit. If approved, it can be merged in the commit that precedes it. There's also still your commit to describe, but I made you author of both your commits.

packages/core/src/VDocument.ts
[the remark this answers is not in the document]
dmo-odoo: Relevant

packages/core/src/VDocument.ts
```diff
@@ -155,4 +153,198 @@ export class VDocument {
...
+     */
+    _isNestedList(node: VNode): node is ListNode {
+        return node.is(ListNode) && !!this._listAncestor(node);
+    }
+    /**
+     * Return a node's ancestor that is a `ListNode`, if any.
+     *
+     * @param node
+     */
+    _listAncestor(node: VNode): ListNode {
```
dmo-odoo: return `node is ListNode` as type

packages/core/src/VDocument.ts
```diff
@@ -155,4 +153,198 @@ export class VDocument {
...
+            const commonAncestor = this._commonAncestor(first, last);
+            const reference = first.ancestor(node => node.parent === commonAncestor) || first;
+            commonAncestor.insertBefore(newList, reference);
+        }
+
+        // Move the nodes to the list
+        withMarkers(() => {
+            // Remove the children of nested lists (they will be moved together
+            // with their parent) and convert nested lists to the new type.
+            const nestedLists = nodes.filter(this._isNestedList.bind(this));
```
dmo-odoo: `_isNestedList` should not be on `this`, it should be a pure function, same for `listAncestor` then you will be able to remove the type castings.

packages/core/src/VDocument.ts
dmo-odoo: No, I agree that iti s out of scope.

dmo-odoo:
I agree with the changes you suggested, you can squash them so that it will be easier to re-review it after you rebase the branch on the new master.

packages/core/src/VDocument.ts
```diff
@@ -155,4 +153,198 @@ export class VDocument {
...
+    _isNestedList(node: VNode): node is ListNode {
+        return node.is(ListNode) && !!this._listAncestor(node);
+    }
+    /**
+     * Return a node's ancestor that is a `ListNode`, if any.
+     *
+     * @param node
+     */
+    _listAncestor(node: VNode): ListNode {
+        return node.ancestor(ancestor => ancestor.is(ListNode)) as ListNode;
```
dmo-odoo: Just write `return node.ancestor(ListNode)` and typescript will understand that what you want is a `ListNode` ... ;-)

packages/core/src/VDocument.ts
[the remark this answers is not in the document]
Zinston: Ok then.

packages/core/src/Parser.ts
```diff
@@ -11,6 +11,9 @@ import { FragmentNode } from './VNodes/FragmentNode';
 import { FormatType, CharNode } from './VNodes/CharNode';
 import { ListNode } from './VNodes/ListNode';
 
+const listTags = ['UL', 'OL'];
+const listContextTags = listTags.concat('LI');
```
Zinston: I don't understand. How does that help with the agnostic parser? This should still be read from the List plugin wouldn't it?

packages/core/src/Renderer.ts
```diff
@@ -1,13 +1,15 @@
 import { VDocumentMap } from './VDocumentMap';
 import { VDocument } from './VDocument';
-import { VNode, RelativePosition, isMarker } from './VNodes/VNode';
+import { VNode, RelativePosition, isMarker, NodePredicate } from './VNodes/VNode';
```
Zinston: Title: you mean 'future agnostic _renderer_' right?

PR #70, opened 2020-01-02
Selection vs Range

dmo-odoo:
Before this commit, it was impossible to use a specific range when executing a command. The command was always using vDocument.range.

We want to allow the parameterization of the range for future features.


packages/core/src/VRange.ts
```diff
@@ -95,227 +45,101 @@ export class VRange {
...
+            reference.prepend(this.start);
+        } else if (position === RelativePosition.AFTER && reference !== this.end) {
+            // We check that `reference` isn't `this.end` to avoid a backward
+            // collapsed range.
+            reference.after(this.start);
         } else {
-            return this._setHead(reference, position);
+            reference.before(this.start);
         }
+        return this;
```
Goaman: Why do we need to return `this`? It is strange to mix a jQuery style return function with the rest of the code that is not jQuery like and return void.

packages/core/src/VRange.ts
```diff
@@ -95,227 +45,101 @@ export class VRange {
...
+     * Because of that, calling this method will always result in a range that
+     * is at least the size that it was prior to calling it, and usually bigger.
      *
-     * @param reference
-     * @param [direction] default: Direction.FORWARD
-     * @see _setHead
+     * @param node The node to extend the range to.
      */
-    extendTo(reference: VNode, direction = Direction.FORWARD): void {
+    extendTo(node: VNode): this {
```
Goaman: Change `node` to `targetNode` ?

packages/plugin-devtools/src/components/TreeComponent.ts
```diff
@@ -13,6 +13,7 @@ interface NodeProps {
 
 interface NodeState {
     folded: boolean;
+    gotcha: boolean;
```
Goaman: What is `gotcha`?

packages/core/src/VDocument.ts
```diff
@@ -156,4 +157,27 @@ export class VDocument {
...
+     * Create a temporary range from the given description and call the given
+     * callback with the newly created range as argument. The range is
+     * automatically destroyed after calling the callback.
+     *
+     * @param description
+     * @param callback
+     */
+    withRange<T>(description: RangeDescription, callback: (range: VRange) => T): T {
+        const { startNode, startPosition, endNode, endPosition } = description;
+        const range = new VRange();
```
Goaman: `vRange` `vrange`? ;)

packages/core/src/VDocument.ts
```diff
@@ -156,4 +157,27 @@ export class VDocument {
...
+
+    /**
+     * Create a temporary range from the given description and call the given
+     * callback with the newly created range as argument. The range is
+     * automatically destroyed after calling the callback.
+     *
+     * @param description
+     * @param callback
+     */
+    withRange<T>(description: RangeDescription, callback: (range: VRange) => T): T {
```
Goaman: `rangeDescription`?

packages/core/src/VRange.ts
```diff
@@ -10,6 +17,51 @@ export class VRange {
...
     //--------------------------------------------------------------------------
 
+    /**
+     * Describe a collapsed range at the given location, targetting a reference
+     * VNode and specifying the position relative to that VNode.
+     *
+     * @param reference
+     * @param position
+     */
+    static at(reference: VNode, position = RelativePosition.BEFORE): RangeDescription {
```
Goaman: at is a very small name. Hard to grep. `selectAt` `getRangeDescriptionAt`?

packages/core/src/VRange.ts
```diff
@@ -10,6 +17,51 @@ export class VRange {
...
+     * `select(a, RelativePosition.AFTER, b, RelativePosition.BEFORE)` => `a[]b`
+     * `select(a, RelativePosition.BEFORE, b, RelativePosition.AFTER)` => `[ab]`
+     * `select(a, RelativePosition.AFTER, b, RelativePosition.AFTER)` => `a[b]`
+     *
+     * @param startNode
+     * @param [startPosition] default: `RelativePosition.BEFORE`
+     * @param [endNode] default: `startNode`
+     * @param [endPosition] default: `RelativePosition.AFTER`
+     */
+    static select(startNode: VNode, endNode?: VNode): RangeDescription;
```
Goaman: `getRangeDescription`?
I first thought that select would actually do something in the VDOM. Then I saw it is a static function and the only thing it does is return an object.
`select` implies a mutation somewhere.

packages/core/test/VDocument.test.ts
```diff
@@ -1859,5 +1861,107 @@ describe('stores', () => {
...
         });
+        describe('withRange', () => {
+            describe('rangeParam with VRange.at', () => {
+                it('should work with VRange.at', () => {
+                    testEditor({
+                        contentBefore: 'ab[]',
+                        stepFunction: async (editor: JWEditor) => {
+                            const aNode = editor.vDocument.root.next(node => node.name === 'a');
+                            await editor.vDocument.withRange(VRange.at(aNode), range => {
+                                editor.execCommand('insertText', range, { value: 'c' });
```
Goaman: The tests are not working.
`execCommand` only takes two arguments.

Change:
```typescript
editor.execCommand('insertText', range, { value: 'c' });
```
By:
```typescript
editor.execCommand('insertText', { value: 'c', range: range });
```

packages/core/src/VRange.ts
dmo-odoo: https://github.com/odoo-dev/jabberwock/issues/71

packages/core/src/CorePlugin.ts
```diff
@@ -13,10 +13,10 @@ export type DeleteBackwardParams = RangeParams;
 export type DeleteForwardParams = RangeParams;
 
 export interface InsertParams extends RangeParams {
-    value: VNode;
+    node: VNode;
```
Goaman: `vNode` `vnode`?

packages/core/src/VNodes/VNode.ts
```diff
@@ -60,13 +66,12 @@ export class VNode {
...
-     * offset. This allows us to handle special cases.
+     * Transform the given DOM location into its VDocument counterpart.
      *
-     * @param domNode
-     * @param offset
+     * @param domNode DOM node corresponding to this VNode
+     * @param offset The offset of the location in the given domNode
      */
-    locateRange(domNode: Node, offset: number): [VNode, RelativePosition] {
+    locate(domNode: Node, offset: number): [VNode, RelativePosition] {
```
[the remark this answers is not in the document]
dmo-odoo: I think it *might* be possible that a VNode return another VNode than itself in this function. @Zinston might have ideas of such nodes ?

packages/core/test/VDocument.test.ts
```diff
@@ -127,12 +113,15 @@ describe('stores', () => {
                     await testEditor({
                         contentBefore: '[]a',
                         stepFunction: (editor: JWEditor) => {
-                            const params: FormatParams = {
+                            const formatParams: FormatParams = {
                                 format: 'bold',
                             };
+                            editor.execCommand('applyFormat', formatParams);
```
Goaman: As you changed the execCommand above, maybe remove it here as well?
```typescript
this.editor.vDocument.applyFormat(formatParams);
```

packages/core/src/VNodes/VNode.ts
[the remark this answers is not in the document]
Zinston: My first thought goes to atomic nodes. If we do end up having atomic groups or atomic nodes with shadow children (quarks? 😊), they would typically return another node than themselves. Eg: a child of an iframe would return the iframe or one of its siblings.
I do think it would be bad practice though and to be avoided as much as possible. If we can ensure that our always returns `this`, it would be great, though in that case indeed just returning the relative position would suffice. I'm talking ideally here but I know we've had plenty of discussions on atomic nodes where we had disagreements on their definition and scope.

packages/core/src/Renderer.ts
```diff
@@ -1,11 +1,11 @@
...
 import { VDocument } from './VDocument';
 import { isMarker, isChar } from '../../utils/src/Predicates';
-import { VNode } from './VNodes/VNode';
+import { VNode, RelativePosition } from './VNodes/VNode';
 import { Format } from '../../utils/src/Format';
-import { VRange } from './VRange';
+import { VSelection } from './VSelection';
 import { CharNode } from './VNodes/CharNode';
-import { RelativePosition } from '../../utils/src/range';
+import { Direction } from './VSelection';
```
Zinston: `import { VSelection, Direction } from './VSelection';`

packages/core/src/Renderer.ts
```diff
@@ -170,43 +169,48 @@ export class Renderer {
...
      * Return the location in the DOM corresponding to the location in the
-     * VDocument of the given range VNode. The location in the DOM is expressed
-     * as a tuple containing a reference Node and a relative position with
-     * respect to the reference Node.
+     * VDocument of the given VNode. The location in the DOM is expressed as a
+     * tuple containing a reference Node and a relative position with respect to
+     * the reference Node.
      *
-     * @param rangeNode
+     * @param vNode
```
Zinston: Didn't we say `node`?

packages/core/src/VDocument.ts
```diff
@@ -27,46 +27,49 @@ export class VDocument {
...
     insertParagraphBreak(): void {
-        // Remove the contents of the selection if needed.
-        if (!this.range.isCollapsed()) {
+        const range = this.selection.range;
+        // Remove the contents of the range if needed.
+        if (!range.isCollapsed()) {
             this.deleteSelection();
         }
-        this.range.start.parent.splitAt(this.range.start);
+        range.startContainer.splitAt(range.start);
```
Zinston: Nice!

packages/core/src/VDocument.ts
```diff
@@ -75,17 +78,18 @@ export class VDocument {
...
+        const range = this.selection.range;
         let format: FormatType = {};
         if (this.formatCache) {
             return this.formatCache;
-        } else if (this.range.isCollapsed()) {
-            const charToCopyFormat = (this.range.start.previousSibling(isChar) ||
-                this.range.start.nextSibling(isChar) || {
+        } else if (range.isCollapsed()) {
+            const charToCopyFormat = (range.start.previousSibling(isChar) ||
+                range.start.nextSibling(isChar) || {
```
Zinston: Out of scope but this bit here is wrong and will have to disappear soon... Just wanted to pull your attention on it.

packages/core/src/VDocument.ts
```diff
@@ -98,18 +102,19 @@ export class VDocument {
...
      */
     deleteSelection(): void {
+        const range = this.selection.range;
         withMarkers(() => {
-            const nodes = this.range.selectedNodes;
+            const nodes = range.selectedNodes;
             if (!nodes.length) return;
-            this.range.collapse(this.range.start); // Reset the direction of the range.
-            let reference = this.range.end;
+            range.collapse();
```
Zinston: Is this still needed if we don't handle the direction specifically? It seems that we were collapsing only to reset the direction.

packages/core/src/VNodes/VNode.ts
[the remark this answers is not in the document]
Zinston: What's the status here finally?

packages/core/src/VRange.ts
```diff
@@ -95,227 +45,101 @@ export class VRange {
...
+     * @param targetNode The node to extend the range to.
      */
-    extendTo(reference: VNode, direction = Direction.FORWARD): void {
+    extendTo(targetNode: VNode): this {
         let position: RelativePosition;
-        if (direction === Direction.FORWARD) {
-            if (reference.hasChildren()) {
-                reference = reference.next();
-                reference = reference.firstLeaf();
+        if (targetNode.isBefore(this.start)) {
```
Zinston: I find this harder to read. How about naming it? `const isExtendingBackward = targetNode.isBefore(this.start);`
Wait... if we're extending to a node that is _before_ the start of the range, we're extending backward indeed. So why was it `direction === Direction.FORWARD` before?

packages/core/src/VSelection.ts
```diff
@@ -0,0 +1,179 @@
...
+        return this;
+    }
+    /**
+     * Extend the selection from its anchor to the given location, targetting a
+     * `reference` VNode and specifying the `direction` of the extension.
+     *
+     * @param reference
+     * @param [direction] default: Direction.FORWARD
+     */
+    extendTo(reference: VNode, direction = Direction.FORWARD): this {
```
Zinston: So this basically duplicates the logic of `range.extendTo` but with the possibility of giving it a direction?

packages/plugin-devtools/src/components/TreeComponent.ts
```diff
@@ -91,6 +91,16 @@ export class TreeComponent extends OwlUIComponent<NodeProps> {
             event.stopImmediatePropagation();
         }
     }
+    repr(): string {
+        if (this.props.vNode.type === 'marker') {
+            // Representation of marker nodes might change depending on the
+            // context in which they are referenced, even though themselves did
```
Zinston: "even though they did"?

packages/core/src/CorePlugin.ts
```diff
@@ -56,56 +66,56 @@ export class CorePlugin extends JWPlugin {
     /**
      * Insert a paragraph break.
      */
-    insertParagraphBreak(): void {
-        this.editor.vDocument.insertParagraphBreak();
+    insertParagraphBreak(params: InsertParagraphBreakParams): void {
```
Zinston: Oh ok so my earlier comment can be ignored. Nice.

packages/core/test/VDocument.test.ts
```diff
@@ -109,7 +100,7 @@ describe('stores', () => {
                         await testEditor({
                             contentBefore: '<b>a</b>[b]<b>c</b>',
                             stepFunction: (editor: JWEditor) => {
-                                editor.vDocument.insertText('d');
+                                editor.execCommand('insertText', { value: 'd' });
```
Zinston: I think you can remove the `render` underneath.

packages/core/test/VDocument.test.ts
```diff
@@ -119,7 +110,7 @@ describe('stores', () => {
                         await testEditor({
                             contentBefore: '<b>[a</b>b]<b>c</b>',
                             stepFunction: (editor: JWEditor) => {
-                                editor.vDocument.insertText('d');
+                                editor.execCommand('insertText', { value: 'd' });
```
Zinston: Idem.

packages/core/src/VDocument.ts
```diff
@@ -101,8 +97,7 @@ export class VDocument {
      * Truncate the tree by removing the selected nodes and merging their
      * orphaned children into the parent of the first removed node.
      */
-    deleteSelection(): void {
-        const range = this.selection.range;
+    deleteSelection(range: VRange): void {
```
Zinston: Why not use the document's selection's range by default if nothing is passed?

packages/core/src/VDocument.ts
```diff
@@ -137,8 +132,7 @@ export class VDocument {
      * If the selection is collapsed, set the format on the selection so we know
      * in the next insert which format should be used.
      */
-    applyFormat(formatName: string): void {
-        const range = this.selection.range;
+    applyFormat(range = this.selection.range, formatName: string): void {
```
Zinston: I'm surprised Typescript let you do this. So `range` is optional but `formatName` isn't. That means if I want to use this with the default value of `range` I have to call it like this: `applyFormat(undefined, 'bold')` ?

packages/core/src/VRange.ts
```diff
@@ -1,10 +1,71 @@
-import { VNode, RelativePosition } from './VNodes/VNode';
+import { VNode, RelativePosition, Point } from './VNodes/VNode';
 import { withMarkers } from '../../utils/src/markers';
 import { MarkerNode } from './VNodes/MarkerNode';
 
 export class VRange {
-    readonly start = new MarkerNode();
-    readonly end = new MarkerNode();
+    readonly start: VNode = new MarkerNode();
```
Zinston: Why are you typing it? Why as `VNode` instead of `MarkerNode`?

packages/core/src/VDocument.ts
```diff
@@ -4,6 +4,7 @@ import { isChar } from '../../utils/src/Predicates';
 import { withMarkers } from '../../utils/src/markers';
 import { FragmentNode } from './VNodes/FragmentNode';
 import { VSelection } from './VSelection';
+import { VRange } from './VRange';
```
Zinston: I don't understand the title of this commit. Could you elaborate?

packages/core/src/CorePlugin.ts
Zinston: :+1:

packages/core/src/VNodes/VNode.ts
```diff
@@ -10,9 +10,9 @@ export enum RelativePosition {
 }
 
 export enum VNodeType {
-    NODE = 'node',
-    MARKER = 'marker',
-    FRAGMENT = 'fragment',
+    NODE = 'Node',
```
Zinston: :eyes:

Zinston: image

packages/plugin-devtools/src/components/PathComponent.ts
```diff
@@ -4,7 +4,7 @@ import { CharNode } from '../../../core/src/VNodes/CharNode';
 
 export class PathComponent extends OwlUIComponent<{}> {
     getNodeRepr(vNode: VNode): string {
-        const name: string = (vNode.type && vNode.type.toLowerCase()) || '?';
+        const name: string = vNode.name || '?';
```
Zinston: :dark_sunglasses:

Zinston: image

packages/core/src/CorePlugin.ts
```diff
@@ -86,11 +85,11 @@ export class CorePlugin extends JWPlugin {
      * Delete in the backward direction (backspace key expected behavior).
      */
     deleteBackward(): void {
-        const range = this.editor.vDocument.range;
+        const range = this.editor.vDocument.selection.range;
```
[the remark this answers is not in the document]
dmo-odoo: Commands work on *a* range, not *the* range. It is the command responsibility to decide which is the default range it uses when none is provided. Some commands do not need the range so setting a default value for `params.range` through `execCommand` would make it weird to have range in their params if they don't use it, plus it would be impossible to detect whether a range was provided or not. Maybe I want to write a command which does something at the start of the document unless you give it another range to work on ? Or maybe I want to do something on the first heading node ? So I could write:
```
const nodes = params.range ? params.range.selectedNodes() : this.editor.vDocument.root.firstDescendant(h1);
```
I think a more obvious issue is the fact that `this.editor.vDocument` is a pain to read and write and that might be part of why you are skeptical of the `this.editor.vDocument.selection.range`. I agree that something might be done to alleviate that issue.

packages/core/src/EventNormalizer.ts
```diff
@@ -581,26 +579,27 @@ export class EventNormalizer {
...
         // length between the previous and current content in order to compute
         // their corresponding indices in the previous DOM.
         const insertPreviousStart = insertStart;
         const insertPreviousEnd = insertEnd + previousLength - currentLength;
-        const insertionRange: DomRangeDescription = {
-            startContainer: previousNodes[insertPreviousStart].origin,
-            startOffset: previous.offsets[insertPreviousStart],
-            endContainer: previousNodes[insertPreviousEnd].origin,
-            endOffset: previous.offsets[insertPreviousEnd],
+        const impliedSelection: DomSelectionDescription = {
```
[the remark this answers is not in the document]
dmo-odoo: It is not a range and it is not *the* selection. It is implied by the mutations yes. For example, you write `Christofe` then you click on the suggestion of your spell-checking keyboard which says `Christophe`. We implement this as a `setSelection(Christofe)` followed by `insertText(Christophe)` so the selection is *implied* because you never *actually* selected anything in the text. I can change `implied` to something else if you still think the name is wrong after this explanation. Now that I think about it and the commands can receive a range as parameter, it might be more correct to trigger an insertText with this *insertion range* than fake a `setSelection` to use it as the range of the command.

packages/core/src/Parser.ts
```diff
@@ -85,42 +84,57 @@ export class Parser {
...
-     * @param range
+     * @param selection
      * @param [direction]
      */
-    parseRange(range: DomRangeDescription): VRangeDescription;
-    parseRange(range: Range, direction: Direction): VRangeDescription;
-    parseRange(range: Range | DomRangeDescription, direction?: Direction): VRangeDescription {
-        const start = this._locate(range.startContainer, range.startOffset);
-        const end = this._locate(range.endContainer, range.endOffset);
+    parseSelection(selection: Selection | DomSelectionDescription): VSelectionDescription {
```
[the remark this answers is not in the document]
dmo-odoo: Well `Selection` is the true class of the DOM, while `DomSelectionDescription` is when we describe a selection by specifying an `anchorNode`, a `focusNode` etc. This method is called by the `EventManager` to convert the description sent by the `EventNormalizer` and it is called by the `Parser` itself, in which case it gives it the true `Selection` of the DOM. Neither type is a subtype of the other because `Selection` has a lot more than the 4 properties we use *but* it is missing `direction`. I think we could recompute `direction` here though so it could be possible to define an interface that would suit both cases. The `EventNormalizer` has to compute the direction anyway to give it to us here so it's not like we would add additional computional work by computing it here instead.

packages/core/src/VNodes/VNode.ts
dmo-odoo: Now that I realized (thanks to the list PR) that you are defining these to display them in the DevTools, I might revert this change such that the devtools looks a bit more like the "css" part of Chrome's.

packages/core/src/VNodes/VNode.ts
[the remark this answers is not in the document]
dmo-odoo: I'm keeping `locate` for now. I like the 1<->1 conversion between `[Node, number]` and `[VNode, RelativePosition]`.

packages/core/src/VRange.ts
dmo-odoo: It's hard to read from the diff but I actually swapped the `if` and `else` part ! ;-) I'll make a variable for this check.

packages/core/src/VSelection.ts
dmo-odoo: It does not duplicate the logic of `range.extendTo`. The range one is moving either start or end depending on which one is closer to the node to include. This one moves the focus of the selection such that it selects the given node. Example:
```
<p>ab[cd]ef</p>
```
Calling `range.extendTo(b)` will produce:
```
<p>a[bcd]ef</p>
```
While `selection.extendTo(b)` will produce:
```
<p>a]b[cdef</p>
```
Basically `selection.extendTo` is the equivalent of the DOM's `selection.extend`, while the range one is "smarter" in the way that it trully *extends* the range, conserving what was *already* in it.

packages/core/src/VDocument.ts
dmo-odoo: It's an opinionated choice, actually. This one is not a command per se, you can't `execCommand` it. It is merely a "public helper" that is called by the other commands. If `range` is not required, you can write this:
```
function myCommand(params: CommandParams) {
    const range = params.range || selection.range;
    [do stuff]
    [do more stuff]
    [do yet more stuff, your brain hurts]
    this.deleteSelection();
    [so much stuff]
}
```
... and you just introduced a bug if a range is passed to your function, because you are not forwarding it to `deleteSelection`.

packages/core/src/VDocument.ts
dmo-odoo: Nice catch. I inverted the params for the three commands where I did it wrong.

packages/core/src/EventNormalizer.ts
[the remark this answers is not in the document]
Zinston: Right, that does make more sense. Seems like this kind of thing is even the whole point of this PR isn't it?

packages/core/src/Parser.ts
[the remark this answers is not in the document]
Zinston: I think it's the `Normalizer`'s job to make a `DomSelectionDescription` object from the native `Selection`: it's normalizing it precisely so our parser always receives the same type of object and not the native stuff.

packages/core/src/VRange.ts
Zinston: Oh ok now it makes more sense! You're right it's not a backward selection but there is a direction to the extension (left or right if the text is ltr). Because whether that direction is ltr or rtl is dependent on whether the text itself is ltr or rtl, we chose to name these things backward and forward instead. Like `deleteForward` and `deleteBackward`. Nothing to do with the selection either but it's still backward or forward!

packages/core/src/VSelection.ts
Zinston: Ok, got it. That's a bit confusing thought, I think it might require some extra doc effort.

packages/core/src/VDocument.ts
Zinston: Ok right that makes sense.

packages/core/src/VSelection.ts
Gorash: When we will use `extendTo` ?
There is no test for these methods and no code calls them

PR #72, opened 2020-01-13
[IMP] Parser: make parser node-agnostic

Zinston:
#### Notes
1. The following nodes become managed by plugins: `Char`, `LineBreak`, `Heading`, `Paragraph`
2. `Heading` and `Paragraph` now being handled individually, `VElement` becomes the model for an unknown `VNode`. As such, it has no parser but is instead instantiated whenever the node could not be parsed by anyone else.
3. The `Parser` now receives `ParsingPredicate`s from the plugins. These return a parsing function or undefined. The first predicate to return a parsing function (in the order that the plugins were added to the editor) parses the node and returns an array of `VNode` (extensions).
4. Parsing functions receive the current parsing context and return a new one, together with a `ParsingMap` (see below). In `Parser`, `nextParsingContext` then sees if that parsing context has to be added to its `_contextStack` (see below), if the stack has to be popped, or if it can simply move on to the next node to parse.
5. The aforementioned `_contextStack` is there to ensure nodes are parsed according to the context given by their parent. This allows us to handle inherited contexts, like the format state.
6. The aforementioned `ParsingMap` maps the `vNode`s that the parsing function created with their respective dom nodes. This way the `Parser` can then add them to the `VDocumentMap`.
6. Currently, appending a `VNode` to its parent is done within the parsing functions. It could instead return the nodes to append, if any, so the `Parser` handles that part. This would make for better separation of concerns, but uglier, chunkier return statements.

#### Todo
- Move format logic out of there (for a 'format' PR)
- Throw an error if a parsing function changes the node to parse in its context beyond acceptable bounds (ie. beyond its own parent, as that could break `nextParsingContext`.
- Do something similar to this with the `Renderer`.


Zinston:
@dmo-odoo Here's a first acceptable version. I believe it has everything we discussed. Obviously the commits will need to be rearranged but I thought this could actually be easier for you to review this way for now.

packages/core/src/Parser.ts
```diff
@@ -217,33 +221,34 @@ export class Parser {
...
         const parentVNode = this.currentContext.parentVNode;
         const format = this.currentContext.format[0];
         this.parseNode();
-        const parsedVNodes = this.currentContext.lastParsed as CharNode[];
-        parsedVNodes.forEach((parsedVNode: CharNode, index: number) => {
-            parsedVNode.format = { ...parsedVNode.format, ...format };
-            VDocumentMap.set(parsedVNode, node, index);
-            parentVNode.append(parsedVNode);
-        });
+        Array.from(this.currentContext.parsingMap.keys()).forEach(
```
dmo-odoo: The context being somewhat hidden in `this` is particulary hurting here, because we have no way to tell that calling `this.parseNode()` had any impact on `this.currentContext.parsingMap` !

packages/core/src/Parser.ts
```diff
@@ -217,33 +221,34 @@ export class Parser {
...
+            (parsedVNode: CharNode, index: number) => {
```
dmo-odoo: I'm surprised typescript lets you write this. How can it be sure that it is a `CharNode` rather than `VNode`?

packages/core/src/Parser.ts
```diff
@@ -178,10 +182,10 @@ export class Parser {
     _parseElementNode(): void {
         const context = { ...this.currentContext };
 
-        const node = context.node;
+        const node = context.currentNode;
         const newContext = this.parseNode();
-        const parsedNode = newContext.lastParsed[newContext.lastParsed.length - 1];
-        if (Format.tags.includes(context.node.nodeName)) {
+        const parsedNode = newContext.parsingMap.keys().next().value;
```
dmo-odoo: 👀

packages/core/src/Parser.ts
```diff
@@ -70,7 +70,12 @@ export class Parser {
         // the HTML tag of the DOM Node. This way we may not support the node
         // but we don't break it either.
         const context = { ...this.currentContext };
-        context.lastParsed = [new VElement(this.currentContext.node.nodeName)];
+        context.parsingMap = createMap([
+            [
+                new VElement(this.currentContext.currentNode.nodeName),
+                this.currentContext.currentNode,
+            ],
+        ]);
```
dmo-odoo: What was the point of this `parsingMap` again ? (I still don't understand the point of `lastParsed` either.)

packages/core/src/Parser.ts
```diff
@@ -10,16 +10,16 @@ import { VElement } from './VNodes/VElement';
...
     readonly rootNode?: Node;
     currentNode?: Node;
-    parsingMap?: Map<VNode, Node[]>;
     parentVNode?: VNode;
     format?: FormatType[];
     vDocument: VDocument;
 }
 export type ParsingPredicate = (node: Node) => ParsingFunction;
-export type ParsingFunction = (context: ParsingContext) => ParsingContext;
+export type ParsingFunction = (context: ParsingContext) => [ParsingContext, ParsingMap];
```
dmo-odoo: This is undoing the previous commit. In a good way so there is that, but I don't understand why it is not squashed into the previous commit.

packages/core/src/Parser.ts
```diff
@@ -222,61 +220,72 @@ export class Parser {
...
             parentVNode.append(parsedVNode);
         });
+        return context;
     }
-    _nextParsingContext(): void {
-        let nextContext = { ...this.currentContext };
-
-        const node = this.currentContext.currentNode;
+    _nextParsingContext(context: ParsingContext): void {
+        const node = context.currentNode;
```
dmo-odoo: Why was `nextContext` lost here ? It is cognitively important !

packages/core/src/Parser.ts
```diff
@@ -151,19 +147,21 @@ export class Parser {
      * Parse a node depending on its DOM type.
      */
     _parseNode(): void {
+        let context;
         switch (this.currentContext.currentNode.nodeType) {
             case Node.ELEMENT_NODE: {
-                this._parseElementNode();
+                context = this._parseElementNode();
```
dmo-odoo: 👍

packages/core/src/Parser.ts
```diff
@@ -222,61 +220,72 @@ export class Parser {
...
             if (parsedParent) {
-                nextContext.parentVNode = parsedParent[0];
+                context.parentVNode = parsedParent[0];
             }
-            this._contextStack.push(nextContext);
+            this._contextStack.push(context);
         } else if (node.nextSibling) {
             // Parse the siblings of the current node with the same parent, if any.
-            nextContext.currentNode = node.nextSibling;
+            this.currentContext.currentNode = node.nextSibling;
```
dmo-odoo: Weird... sometimes this function alters `context`, sometimes it alters `this.currentContext`. Red alert, red alert ! 🔴

packages/core/src/Parser.ts
```diff
@@ -222,61 +220,72 @@ export class Parser {
...
+                if (!this.currentContext) {
+                    // If we went back up the whole stack, we start with a new
+                    // context.
+                    this._contextStack.push({
+                        rootNode: context.rootNode,
+                        currentNode: ancestor.nextSibling,
+                        parentVNode: context.parentVNode,
+                        format: [],
+                        vDocument: context.vDocument,
+                    });
```
dmo-odoo: What ? There was an initial context at the beginning when the parsing started. There is no reason this one could not be retrieved here.

packages/core/src/Parser.ts
```diff
@@ -8,14 +8,14 @@ import { DomRangeDescription } from './EventNormalizer';
 import { utils } from '../../utils/src/utils';
 import { VElement } from './VNodes/VElement';
 import { FragmentNode } from './VNodes/FragmentNode';
-import { FormatType, CharNode } from '../../plugin-char/CharNode';
+import { FormatType } from '../../plugin-char/CharNode';
```
dmo-odoo: I don't understand this entire commit.

packages/core/src/CorePlugin.ts
```diff
@@ -68,20 +62,6 @@ export class CorePlugin extends JWPlugin {
     insert(params: InsertParams): void {
         this.editor.vDocument.insert(params.value);
     }
-    /**
-     * Insert text at range.
-     *
-     * @param params
-     */
-    insertText(params: InsertTextParams): void {
```
dmo-odoo: You forgot to move the `Params` types associated with these commands. And this PR is not rebased on master.

packages/plugin-char/Char.ts
```diff
@@ -33,4 +42,94 @@ export class Char extends JWPlugin {
...
+     *
+     * If the range is collapsed, add `text` to the vDocument and copy the formating of the
+     * previous char or the next char.
+     *
+     * If the range is not collapsed, replace the text with the formating that was present in the
+     * range.
+     *
+     * @param text
+     */
+    insertText(params: InsertTextParams): void {
```
dmo-odoo: I would have put `insertText` before `applyFormat` because I see `insertText` as being more "basic" than `applyFormat`.

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -733,7 +733,7 @@ describe('plugin-devtools', () => {
             const subpanel = devtools.querySelector('devtools-panel.active mainpane-contents');
             expect(subpanel.textContent).to.equal('insertTextsetRange');
         });
-        it('should select "insertParagraphBreak"', async () => {
+        it('should select "insertLineBreak"', async () => {
```
dmo-odoo: Isn't this undoing the previous commit changes ?

packages/core/test/Renderer.test.ts
```diff
@@ -1,5 +1,5 @@
 import { expect } from 'chai';
-import JWEditor from '../src/JWEditor';
+import { createBasicEditor } from '../../bundles/basic';
```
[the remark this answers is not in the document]
dmo-odoo: test
```
typescript
const test = 'test';
```
test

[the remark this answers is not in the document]
dmo-odoo: test
```typescript
test 
```
test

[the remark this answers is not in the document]
dmo-odoo: test
```typescript
const dmo;
```
test

packages/core/src/Parser.ts
```diff
@@ -149,12 +159,49 @@ export class Parser {
     _parseNode(): void {
         let context;
         switch (this.currentContext.currentNode.nodeType) {
-            case Node.ELEMENT_NODE: {
-                context = this._parseElementNode();
-                break;
-            }
+            case Node.ELEMENT_NODE:
```
Zinston: This switch can be removed.

packages/plugin-char/Char.ts
Zinston: Ok.

packages/plugin-char/Char.ts
```diff
@@ -33,4 +42,94 @@ export class Char extends JWPlugin {
...
+        const text = params.value;
+        const format = this._getCurrentFormat();
+        // Split the text into CHAR nodes and insert them at the range.
+        const characters = text.split('');
+        const nodes = characters.map(char => new CharNode(char, format));
+        this.editor.vDocument.formatCache = null;
+        nodes.forEach(node => this.editor.vDocument.insert(node));
+    }
+
+    //--------------------------------------------------------------------------
```
Zinston: fail

packages/core/src/Parser.ts
```diff
@@ -17,25 +17,32 @@ interface ParsingContext {
     format?: FormatType[];
     vDocument: VDocument;
 }
-
+export type ParsingPredicate = (node: Node) => ParsingFunction;
```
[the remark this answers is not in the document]
Zinston: What should we name them?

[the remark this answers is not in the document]
Goaman: I don't see the value of an intermediate structure (e.g. the parsing predicates).
I only see boilerplate code.
The branching could be handled directly in one simple `parse` (and for the renderer, `render`) function without introducing another concept.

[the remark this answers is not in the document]
Zinston: After discussion, we're removing this predicate thing.

packages/plugin-list/test/List.test.ts
```diff
@@ -0,0 +1,2410 @@
+import { expect } from 'chai';
```
Zinston: @Goaman I need your mocha expertise for this commit: the tests are not being concatenated with the other tests.

packages/core/src/VNodes/VNode.ts
```diff
@@ -47,11 +47,17 @@ export class VNode {
     get name(): string {
         return this.constructor.name;
     }
+    /**
```
Zinston: @dmo-odoo In case you wonder what the concrete use is of this, and what it has to do with the PR: it allows us to remove the reference to `LineBreakNode` in `DevTools`. It might be overkill though so if you want to skip it I'm not going to fight ;-)

packages/plugin-list/List.ts
```diff
@@ -1,22 +1,83 @@
...
+        while (children.length && _isEmptyTextNode(children[0])) {
+            children.shift();
+        }
+        // A list item with no children should be skipped.
+        if (!children.length) {
             return [context, parsingMap];
         }
+        // Inline elements in a list item should be wrapped in a paragraph.
+        if (!utils.isBlock(children[0]) || children[0].nodeName === 'BR') {
+            const paragraph = new ParagraphNode(); // todo: remove reference to plugin
```
Zinston: @dmo-odoo This is not ok, right? But I can't think of a way to avoid it...

PR #73, opened 2020-01-14
[ADD] VNode: add support for classes as predicates

dmo-odoo:


dmo-odoo:
@Zinston I suggest you to try this commit on your list branch.
By writing your `instanceof` predicates like the following:
```
export function isList(node: VNode): node is ListNode {
    return node instanceof ListNode;
}
```
Typescript will now recognize that the return value of `node.next(isList)` is a `ListNode`, not merely a `VNode`, thus removing the need to manually do `as ListNode` on the result.

dmo-odoo:
@Zinston Even better. I just added a magic `is` to the `Predicates` utils which creates type guards from classes. You don't even need to manually create a predicate to use type guards on classes that extends VNode.
Example:
```
const list: ListNode = item.ancestor(is(ListNode));
```

dmo-odoo:
@Zinston Retrospectively, I'm not sure that `is` method is very useful. It's not hard to write a proper type guard and at least that function will only be defined once, where `is` will generate a brand new function everytime we reach that point in the code...

dmo-odoo:
@Zinston Remember a few hours ago when I told you I tried and it couldn't be done ? Well, it turns out I just did it.

Zinston:
> @Zinston Remember a few hours ago when I told you I tried and it couldn't be done ? Well, it turns out I just did it.

image

Zinston:
> > @Zinston Remember a few hours ago when I told you I tried and it couldn't be done ? Well, it turns out I just did it.
> 
> image

Also: 
image

PR #74, opened 2020-01-15
Agnostic Renderer

Zinston:


packages/plugin-char/Char.ts
```diff
@@ -42,6 +53,52 @@ export class Char extends JWPlugin {
...
+        // Browsers don't render leading/trailing space chars otherwise.
+        text = text.replace(/^ | $/g, '\u00A0');
+
+        // Create and append the text node, update the VDocumentMap.
+        const renderedNode = document.createTextNode(text);
+        parent.appendChild(renderedNode);
+        context.parentNode.appendChild(fragment);
+        context.renderingMap = createMapFromDom([
+            [renderedNode, charNodes.concat(renderedFormats)],
+        ]);
```
Goaman: Should rather return the renderingMap rather than writing the context

packages/plugin-dom-renderer/DomRenderer.ts
```diff
@@ -25,43 +27,77 @@ export class DomRenderer extends Renderer<DocumentFragment> {
...
+        // Map the parsed nodes to the DOM nodes they represent.
+        Array.from(context.renderingMap.keys()).forEach(domNode => {
+            const renderedVNodes = context.renderingMap.get(domNode);
+            renderedVNodes.forEach((node: VNode, index: number) => {
+                VDocumentMap.set(node, domNode, index);
+            });
+        });
+
+        // We're done using the rendering map: wipe it.
+        context.renderingMap = new Map<Node, VNode[]>();
```
Goaman: If the renderingMap is returned, no need for this line

packages/plugin-dom-renderer/DomRenderer.ts
```diff
@@ -25,43 +27,77 @@ export class DomRenderer extends Renderer<DocumentFragment> {
...
+    _renderNode(context: DomRenderingContext): DomRenderingContext {
+        context = this.renderNode({ ...context });
+
+        // Map the parsed nodes to the DOM nodes they represent.
+        Array.from(context.renderingMap.keys()).forEach(domNode => {
+            const renderedVNodes = context.renderingMap.get(domNode);
+            renderedVNodes.forEach((node: VNode, index: number) => {
+                VDocumentMap.set(node, domNode, index);
+            });
+        });
```
Goaman: Should have it's own map. Maybe in another PR?

packages/core/src/JWEditor.ts
```diff
@@ -140,7 +156,7 @@ export class JWEditor {
      */
     execCommand(id: CommandIdentifier, args?: CommandArgs): void {
         this.dispatcher.dispatch(id, args);
-        this.renderer.render(this.vDocument, this.editable);
+        this.renderers[this.defaultRendererId].render(this.vDocument, this.editable);
```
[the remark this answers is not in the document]
dmo-odoo: All renderers should be called ? So if I have installed, say, a PDFRenderer, it re-generates a new PDF at every change ? I would have expected it to not do anything unless I click on "Export to PDF" or whatever. It looks like we may need a separation between "live" renderers and "exports" renderer ?

packages/core/src/JWEditor.ts
```diff
@@ -120,10 +121,18 @@ export class JWEditor {
...
-            pluginClass.renderers.forEach(Renderer => {
-                this.renderers[Renderer.id] = new Renderer();
+            pluginClass.renderers.forEach(renderer => {
+                this.renderers[renderer.name] = renderer;
             });
         }
+        // Register the parsing predicate of this plugin if it has any.
+        Object.keys(pluginClass.renderingPredicates).forEach(rendererId => {
+            if (Object.keys(this.renderers).includes(rendererId)) {
+                this.renderers[rendererId].addRenderingPredicate(
```
dmo-odoo: I guess we'll veer away from predicates here as well.

packages/core/src/JWEditor.ts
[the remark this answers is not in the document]
Goaman: They should at least have a signal that tells them that the document has changed.
Maybe instead of calling a method `render`, we could call something like `changeDocument`.

packages/core/src/Renderer.ts
```diff
@@ -1,6 +1,25 @@
...
+    currentVNode?: VNode; // Current VNode rendered at this step.
+    parentNode?: Node | DocumentFragment; // Node to render the VNode into.
+}
+export type RenderingPredicate<T> = (node: VNode) => RenderingFunction<T>;
+export type RenderingFunction<T> = (node: VNode, defaultRendering?: T) => T;
+
+export class Renderer<T> {
+    renderingPredicates = new Set<RenderingPredicate<T>>();
+    readonly name: string;
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-empty-function, @typescript-eslint/no-explicit-any
```
dmo-odoo: // dmo-disable-next-line @dmo-eslint/no-eslint-disable

packages/plugin-dom-renderer/DomRendererPlugin.ts
```diff
@@ -2,5 +2,5 @@ import { JWPlugin } from '../core/src/JWPlugin';
 import { DomRenderer } from './DomRenderer';
 
 export class DomRendererPlugin extends JWPlugin {
-    static readonly renderers = [DomRenderer];
+    static readonly renderers = [new DomRenderer()];
```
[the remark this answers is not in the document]
dmo-odoo: @Zinston Just so you know, if you make Renderer an interface (as suggested by nby) rather than a class, it would be possible to write:
```export class DomRendererPlugin extends JWPlugin implements Renderer {```

Such that a plugin can be a renderer, rather than serving no other purpose than providing a `renderers` key to another object.

You can even implements multiple interfaces, for example:
```export class MyPlugin extends JWPlugin implements Renderer, WhateverFeature, WhateverOtherFeature {```

File: packages/core/src/VDocumentMap.ts [hunk 8]
```diff
@@ -92,3 +92,21 @@ export function createMap(items: Array<[VNode, Node | Node[]]>): Map<VNode, Node
...
+                map.set(domNode, node);
+            }
+        } else if (map.has(domNode)) {
+            map.get(domNode).push(node);
+        } else {
+            map.set(domNode, [node]);
+        }
+    });
+    return map;
+}
```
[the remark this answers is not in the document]
dmo-odoo: I think it's just an artifact of the previous `createMap` paradigm of the parser branch on which this branch was based. We discussed about it yesterday and decided to remove it from the parser branch so of course it should be removed from here as well.

packages/plugin-dom-renderer/DomRenderer.ts
```diff
@@ -1,14 +1,16 @@
...
-import { VNode, isMarker } from '../core/src/VNodes/VNode';
-import { Format } from '../utils/src/Format';
+import { VNode } from '../core/src/VNodes/VNode';
 import { VRange } from '../core/src/VRange';
-import { CharNode } from '../plugin-char/CharNode';
 import { RelativePosition } from '../utils/src/range';
 import { VElement } from '../core/src/VNodes/VElement';
 
-export class DomRenderer extends Renderer<DocumentFragment> {
+export interface DomRenderingContext extends AbstractRenderingContext {
```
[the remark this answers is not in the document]
dmo-odoo: Yeah, actually, I would have moved nearly everything from Renderer.ts to DomRenderer.ts and just keep the "render" method in the Renderer interface. If something else can be generalized later, we can do it at that point, but now it's a bit too early and erroneous.

packages/plugin-dom-renderer/DomRenderer.ts
dmo-odoo: Yes, this is out of scope of this PR.

packages/plugin-dom-renderer/DomRendererPlugin.ts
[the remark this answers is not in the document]
Zinston: This particular plugin only does that but nothing should stop you from making a plugin that declares renderers _and_ does other things too... Also, I made `Renderer` a class so that it could handle `addRenderingFunction` and we wouldn't have to rewrite it for each renderer.

packages/core/src/Renderer.ts
Zinston: Well. Agreed. But do you know how I can do without?

packages/plugin-dom-renderer/DomRenderer.ts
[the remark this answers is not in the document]
Zinston: How do we define RenderingFunction then? Since each renderer can define it differently?

packages/plugin-dom/DomRenderer.ts
```diff
@@ -75,7 +75,10 @@ export class DomRenderer extends Renderer<
         for (const renderingFunction of this.renderingFunctions) {
             const renderingResult = renderingFunction({ ...context });
             if (renderingResult) {
-                return renderingResult;
+                context = renderingResult[0];
```
Zinston: This is why I call it "dirty make-it-work" ;-)
How should we handle this kind of situation? We need to be able to render a wrapper, then let the node render itself within said wrapper.
Also there is an issue of priorities. By chance right now it works, but it's only by chance: List has to take control whenever a ListNode's children have to be rendered. But if another renderingFunction renders the node, List will never be called. List _could_ handle the rendering of all its children, like it was done in the non-agnostic renderer... but then they need to be able to call `render` on each of said children.

Zinston:
@dmo-odoo Ok so I'm finally done rebasing and updating this for lists. I also updated to follow the new patterns of Parser (like the stack, and no predicates), and a few of the review comments. Now I added a few comments myself because some things require discussion. There is still work on this PR. It's no easy job but it's prioritary right now. Maybe we can finish this in tandem?

packages/core/src/JWEditor.ts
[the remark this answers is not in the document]
Gorash: We have a confusion of term / name.
I think that it is necessary to make the difference between the signal to render them (informing that they can update), and the request to the editor to obtain the result of a specific rendering.
The `renderers` (plugin thereafter) know if they are displayed or not, and know if they update.
Clicking on 'save as' will be another command.

packages/core/src/JWEditor.ts
```diff
@@ -21,10 +23,11 @@ export class JWEditor {
     dispatcher: Dispatcher;
     eventManager: EventManager;
     pluginsRegistry: JWPlugin[];
-    renderer: Renderer;
+    renderers: Record<string, Renderer> = {};
```
dmo-odoo: I suggest to give a nickname to that `string`. Maybe `RenderingOutputType` ? That's a bit long, though.

packages/core/src/JWEditor.ts
```diff
@@ -111,6 +116,14 @@ export class JWEditor {
         if (pluginClass.parsingPredicate) {
             this.parser.addParsingPredicate(pluginClass.parsingPredicate);
         }
+        // Register the renderers of this plugin if it has any.
+        // If two renderers exist with the same id, the last one to be added
+        // will replace the previous one.
+        if (pluginClass.renderers) {
+            pluginClass.renderers.forEach(Renderer => {
+                this.renderers[Renderer.id] = new Renderer();
```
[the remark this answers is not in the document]
dmo-odoo: sproutch ! 🔨

packages/core/src/JWEditor.ts
```diff
@@ -11,6 +12,7 @@ import { DevTools } from '../../plugin-devtools/src/DevTools';
 export interface JWEditorConfig {
     debug?: boolean;
     theme?: string;
+    defaultRendererId?: string;
```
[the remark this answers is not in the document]
Zinston: We don't. It sucks. But I haven't managed without yet.

packages/plugin-list/List.ts
```diff
@@ -108,60 +108,69 @@ export class List extends JWPlugin {
...
-            node.parent && node.parent.is(ListNode) && !VDocumentMap.toDom(node);
         let result;
-        if (isUnrenderedListItem) {
+        if (node.parent && node.parent.is(ListNode) && !this._isRenderingLi) {
+            this._isRenderingLi = true;
             result = this.renderListItemToDom({ ...context });
             context = result[0];
-        }
-        if (node.is(ListNode)) {
+            this._isRenderingLi = false;
```
dmo-odoo: By "remove list rendering hack", you mean "replace it by another hack", right ? 🚎

packages/core/src/Renderer.ts
```diff
@@ -1,11 +1,12 @@
+import { VNode } from './VNodes/VNode';
 import { VDocument } from './VDocument';
 
 export type RendererIdentifier = string;
-export type RenderingFunction<Context = {}, T = {}> = (context: Context) => T;
+export type RenderingFunction<To = {}> = (node: VNode, to: To) => Map<VNode, To[]>;
```
dmo-odoo: That is kinda weird... If you want to render a node into string you have to pass it a string ?

packages/plugin-dom/DomRenderer.ts
```diff
@@ -40,68 +29,61 @@ export class DomRenderer extends Renderer<
         }
 
         // Don't render `root` itself if already rendered, render its children.
-        const renderedRoot = VDocumentMap.toDom(root);
-        const parentNode = renderedRoot ? renderedRoot : target;
-        const firstVNode = renderedRoot ? root.firstChild() : root;
-
-        if (!firstVNode) return;
+        const domRoot = VDocumentMap.toDom(root);
```
dmo-odoo: Is it the root of the dom ?

packages/plugin-dom/DomRenderer.ts
```diff
@@ -40,68 +29,61 @@ export class DomRenderer extends Renderer<
...
 
-        let currentContext: DomRenderingContext = { ...rootContext };
-        do {
-            currentContext = this.renderNode(currentContext);
-        } while ((currentContext = this._nextRenderingContext(currentContext)));
+        let node = firstNode;
+        let rendered = [];
+        while (node) {
+            if (!rendered.includes(node)) {
+                rendered = [...rendered, ...this.renderNode(node, domParent).keys()];
```
dmo-odoo: I would explicitely call `renderNode` in its own statement, it is in fact the most important part of the algorithm.

packages/plugin-dom/DomRenderer.ts
```diff
@@ -40,68 +29,61 @@ export class DomRenderer extends Renderer<
...
+            }
+            node = node.nextSibling();
+        }
```
dmo-odoo: Rather than a let, a while and another reasignation, I think you could just have done:
```
for (const child of root.children) {
    ...
}
```

packages/plugin-dom/DomRenderer.ts
```diff
@@ -122,54 +104,16 @@ export class DomRenderer extends Renderer<
...
-            } else {
-                // If no next ancestor having a sibling could be found then the
-                // tree has been fully rendered.
-                return;
-            }
-        }
-        return this._contextStack[this._contextStack.length - 1];
+        });
+
+        domParent.appendChild(fragment);
```
dmo-odoo: I think the very point of backing down to a "rendering a node renders its children" mechanism is to actually do the append in the parent itself. So the api of `render` would just be `render<T>(node: VNode): T` and then the parent would take the result and append it where he wants it to be appended. The point is the parent controls where the children is placed.

packages/plugin-dom/DomRenderer.ts
```diff
@@ -122,54 +104,16 @@ export class DomRenderer extends Renderer<
...
-            let ancestor = node;
-            // Climb up the ancestor tree to the first parent having a sibling.
-            while (ancestor && !ancestor.nextSibling() && ancestor != currentContext.root) {
-                ancestor = ancestor.parent;
-                this._contextStack.pop();
+        // Render the node's children
+        let rendered = [];
+        node.children.forEach(child => {
+            if (!rendered.includes(child)) {
+                rendered = [...rendered, ...this.renderNode(child, renderedElement).keys()];
```
dmo-odoo: Yeah, really that's not ok :p Actually, with this mechanism, is the map still needed ? I think it is yes, but we said we would set it manually. That's something else though.

packages/plugin-list/List.ts
```diff
@@ -102,74 +101,66 @@ export class List extends JWPlugin {
...
-                renderingMap.set(liNode, [listItem]);
-                // TODO: this should be generic.
-                if (!listItem.hasChildren()) {
-                    liNode.appendChild(document.createElement('BR'));
-                }
+            if (node.is(VElement) && node.htmlTag === 'P') {
+                // Render the node's children
+                let rendered = [];
+                node.children.forEach(child => {
+                    if (!rendered.includes(child)) {
```
dmo-odoo: How could this happen ?

packages/plugin-list/List.ts
```diff
@@ -102,74 +101,66 @@ export class List extends JWPlugin {
...
+                            ...this.editor.renderers.dom.renderNode(child, domListItem).keys(),
+                        ];
+                    }
+                });
             } else {
                 // The node was wrapped in a "LI" but needs to be rendered as well.
-                currentContext = this.editor.renderers.dom.renderNode({
-                    ...currentContext,
-                }) as DomRenderingContext;
+                this.editor.renderers.dom.renderNode(node, domListItem) as Map<VNode, Node[]>;
```
dmo-odoo: as

packages/plugin-list/List.ts
```diff
@@ -102,74 +101,66 @@ export class List extends JWPlugin {
...
-            this._isRenderingLi = false;
-            return [{ ...currentContext }, renderingMap];
+
+            const domNodes = [domListItem];
+            // TODO: this should be generic.
+            if (!node.hasChildren()) {
+                const br = document.createElement('BR');
+                domListItem.appendChild(br);
+                domNodes.push(br);
+            }
```
dmo-odoo: In my understanding, the one and only time where you would still have to do that is in the case where you render a `P` as a `LI` because it is a child of a list and that particular `P` is empty. It's the only time where you "prevent" somewhat the node from rendering itself using the standard mechanism.

packages/core/src/EventManager.ts
```diff
@@ -25,7 +25,7 @@ export class EventManager {
     /**
      * Callback given to the normalizer.
      */
-    _onNormalizedEvent(customEvent: CustomEvent): void {
+    async _onNormalizedEvent(customEvent: CustomEvent): Promise<void> {
```
dmo-odoo: You put `async` and `Promise` but no `await`, how does that work ? I see most of the time we `return this.editor.execCommand` so I guess it works, but the `default` case does not `return`.

packages/plugin-char/test/Char.test.ts
```diff
@@ -8,17 +8,20 @@ import { Parser } from '../../core/src/Parser';
...
         text: text,
     };
-    editor.execCommand('insertText', params);
+    await editor.execCommand('insertText', params);
 };
-const applyFormat = (editor: JWEditor, format: 'bold' | 'italic' | 'underline'): void => {
+const applyFormat = async (
+    editor: JWEditor,
+    format: 'bold' | 'italic' | 'underline',
+): Promise<void> => {
```
dmo-odoo: So much git conflits ahead... 👨‍💻 😵

packages/plugin-dom/DomRenderer.ts
```diff
@@ -78,12 +79,21 @@ export class DomRenderer extends Renderer<Node> {
 
         return renderingMap;
     }
+    async renderChildren(node: VNode, domParent: Node): Promise<void> {
+        let rendered = [];
+        for (const child of node.children) {
+            if (!rendered.includes(child)) {
+                const mapping = await this.renderNode(child, domParent);
```
dmo-odoo: Thank you.

packages/plugin-list/List.ts
```diff
@@ -111,57 +111,47 @@ export class List extends JWPlugin {
             let mapping: [VNode, Node[]][] = [[node, [domListNode]]];
 
             // Render the node's children
-            node.children.forEach(child => {
-                mapping = [...mapping, ...this.renderListItemToDom(child, domListNode)];
-            });
+            for (const child of node.children) {
+                const childrenMapping = await this.renderListItemToDom(child, domListNode);
```
dmo-odoo: You mean `childMapping`, right ?

packages/plugin-list/List.ts
```diff
@@ -23,80 +24,135 @@ function _isEmptyTextNode(node: Node): boolean {
...
-    //--------------------------------------------------------------------------
-    // Public
-    //--------------------------------------------------------------------------
+        // Direct ListNode's VElement children "P" are rendered as "LI"
+        // while other nodes will be rendered inside the "LI".
+        if (node.is(VElement) && node.htmlTag === 'P') {
+            await this.editor.renderers.dom.renderChildren(node, domListItem);
+        } else {
+            // The node was wrapped in a "LI" but needs to be rendered as well.
+            await this.editor.renderers.dom.renderNode(node, domListItem);
```
dmo-odoo: The `renderNode` function is on the renderer, right ? What if it was this function which does the mapping ? It would map the result of the rendering function to the `VNode` it got in argument. No more need for manual mappings *unless* you create unrepresented nodes. Example:
```typescript
// In DomRenderer.ts
renderNode(node: VNode) {
    const result = renderingFunction(node, this.domMap);
    if (result) {
        this.domMap.set(node, result);
    }
}

// In List.ts
renderListToDom(node: ListNode, domMap: Map<VNode: Node[]>) {
    const ul = createUL();
    for (const child of node) {
        const li = createLI();
        const domChild = this.editor.render(child);
        li.appendChild(domChild);
        domMap.get(child).unshift(li);
        ul.appendChild(li);
    }
}
```

packages/core/src/JWEditor.ts
```diff
@@ -114,23 +114,29 @@ export class JWEditor {
...
             if (Object.keys(this.renderers).includes(rendererId)) {
-                this.renderers[rendererId].addRenderingFunction(
-                    pluginClass.renderingFunctions[rendererId],
-                );
+                const renderer = this.renderers[rendererId];
+                const renderingFunctions = plugin.renderingFunctions[rendererId];
+                if (Array.isArray(renderingFunctions)) {
+                    renderingFunctions.forEach(renderingFunction => {
+                        renderer.renderingFunctions.add(renderingFunction);
+                    });
```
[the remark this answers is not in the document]
Zinston: No.

packages/plugin-list/List.ts
Zinston: image

packages/plugin-list/List.ts
Zinston: So you want me to move this into the above `if`?

packages/core/src/EventManager.ts
Zinston: We return an asynchronous function call so that checks out. The default case I'm not sure...

packages/plugin-char/test/Char.test.ts
Zinston: :sos:

packages/plugin-list/List.ts
Zinston: Thanks for using ```typescript.

PR #77, opened 2020-01-20
[WIP] Format

Zinston:


packages/core/src/JWEditor.ts
```diff
@@ -48,6 +48,10 @@ export class JWEditor {
         // The editable property of the editor is the original editable element
         // before start is called, and becomes the clone after start is called.
         this.editable = editable;
+
+        // CorePlugin is a special mandatory plugin that handles the matching
+        // between the core commands and the VDocument.
+        this.addPlugin(CorePlugin);
```
dmo-odoo: Nice catch, it made no sense where it previously was.

packages/plugin-bold/BoldPlugin.ts
```diff
@@ -18,4 +22,15 @@ export class BoldPlugin extends JWPlugin {
...
     }
+    /**
+     * Apply the `format` to the range.
+     *
+     * If the range is collapsed, set the format on the range so we know in the next insert
+     * which format should be used.
+     */
+    applyFormat(params: FormatParams): void {
+        if (params.formatName === 'bold') {
+            this.editor.vDocument.applyFormat(new Bold());
```
dmo-odoo: Ah ah, see, you actually do create the `new` here, so the override of `applyFormat` isn't really useful since it only brings the conversion from `string` to `Format` when you could have given a `Format` since the start :p

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -692,15 +692,15 @@ describe('plugin-devtools', () => {
...
-                'insertLineBreak',
                 'insert',
                 'insertParagraphBreak',
                 'setRange',
                 'deleteBackward',
                 'deleteForward',
                 'selectAll',
+                'applyFormat',
+                'insertText',
+                'insertLineBreak',
```
dmo-odoo: Damn you, Christophe ! 💢

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -733,7 +733,7 @@ describe('plugin-devtools', () => {
             const subpanel = devtools.querySelector('devtools-panel.active mainpane-contents');
             expect(subpanel.textContent).to.equal('insertTextsetRange');
         });
-        it('should select "insertLineBreak"', async () => {
+        it('should select "setRange"', async () => {
```
dmo-odoo: It's `setSelection` now, but I guess you would have noticed it when the test breaks after rebasing :p

dmo-odoo: Now that I think about it, even though `setSelection` sets the *selection* of the VDocument, it is more or less equivalent to "setting the current range of the user edition" so I would not be against renaming it back to `setRange` if someone is emotional about using `setRange`. I mean, the selection *is* the default range that most commands use, so in a way it is not wrong at all. Fuck, now I want to rename it `setRange` and have it act upon `VDocument.selection.range` which would make sense actually... Damn it ! (What do you think ?)

packages/core/src/Parser.ts
```diff
@@ -78,6 +79,7 @@ export class Parser {
         // the HTML tag of the DOM Node. This way we may not support the node
         // but we don't break it either.
         const parsedNode = new VElement(this.currentContext.currentNode.nodeName);
+        parsedNode.attributes = parseAttributes(context.currentNode);
```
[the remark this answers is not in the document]
Zinston: I assumed we were always going to parse from the DOM anyway given that our data is always going to come from the DOM. But in retrospect that assumption was wrong: we could allow editing from markdown for instance, in which case we'd still be parsing the DOM but interpreting the text contents as markdown.

packages/plugin-linebreak/LineBreak.ts
```diff
@@ -27,10 +28,17 @@ export class LineBreak extends JWPlugin {
...
         const node = context.currentNode;
         const domNodes = [node];
+        const attributes = parseAttributes(context.currentNode);
         if (!node.nextSibling) {
             // A <br/> with no siblings is there only to make its parent visible.
             // Consume it since it was just parsed as its parent element node.
             // TODO: do this less naively to account for formatting space.
+            attributes.forEach((value: string, name: string) => {
+                if (!context.parentVNode.attributes) {
+                    context.parentVNode.attributes = new Map<string, string>();
```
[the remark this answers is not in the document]
Zinston: I think @Gorash would scream. No attributes = no map. Better for performances.

packages/utils/src/attributes.ts
```diff
@@ -0,0 +1,9 @@
+function _isElement(node: Node): node is Element {
+    return node.nodeType === Node.ELEMENT_NODE;
+}
+
+export function parseAttributes(node: Node): Map<string, string> {
+    if (_isElement(node)) {
```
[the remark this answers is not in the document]
Zinston: That was to use typescript's type guards. Not doing that, I have to use 'as'.

packages/core/src/Format.ts
```diff
@@ -0,0 +1,16 @@
...
+    render(): Node {
+        const node = document.createElement(this.htmlTag);
+        if (this.attributes) {
+            this.attributes.forEach((value: string, name: string) => {
+                node.setAttribute(name, value);
+            });
+        }
+        return node;
+    }
+}
```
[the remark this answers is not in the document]
Zinston: What is `Inline` in this context?

packages/plugin-devtools/test/devtools.test.ts
Zinston: :imp:

packages/plugin-devtools/test/devtools.test.ts
Zinston: Let's just say I'm not particularly emotional about it :-D

packages/plugin-linebreak/LineBreak.ts
```diff
@@ -17,10 +17,17 @@ export class LineBreak extends JWPlugin {
...
             const domNodes = [node];
+            const attributes = Parser.parseAttributes(context.currentNode);
             if (!node.nextSibling) {
                 // A <br/> with no siblings is there only to make its parent visible.
                 // Consume it since it was just parsed as its parent element node.
                 // TODO: do this less naively to account for formatting space.
+                attributes.forEach((value: string, name: string) => {
+                    if (!context.parentVNode.attributes) {
+                        context.parentVNode.attributes = new Map<string, string>();
+                    }
```
dmo-odoo: I think this instantiation should be done in the node, such that we never have to worry about it, neither for writing attributes nor reading them.

packages/plugin-linebreak/LineBreak.ts
```diff
@@ -17,10 +17,17 @@ export class LineBreak extends JWPlugin {
...
+                    context.parentVNode.attributes.set(name, value);
```
dmo-odoo: That being said, this is very weird. Why move the attributes of the br up to its parent ?

packages/plugin-linebreak/LineBreak.ts
```diff
@@ -29,8 +36,13 @@ export class LineBreak extends JWPlugin {
                 // TODO: do this less naively to account for formatting space.
                 context.currentNode = node.nextSibling;
                 domNodes.push(node.nextSibling);
+                const siblingAttributes = Parser.parseAttributes(node.nextSibling);
+                siblingAttributes.forEach((value: string, name: string) => {
+                    attributes.set(name, value);
+                });
```
dmo-odoo: Same here, why move the attributes from a node to another ? This second BR is somewhat "fake". Remember old versions of the editor where there were the "virtuals" ? You wouldn't want to handle their attributes, would you ?

packages/plugin-linebreak/LineBreak.ts
```diff
@@ -46,6 +58,11 @@ export class LineBreak extends JWPlugin {
         if (context.currentNode.is(LineBreakNode)) {
             const lineBreak = context.currentNode;
             const br = document.createElement('br');
+            if (lineBreak.attributes) {
+                lineBreak.attributes.forEach((value: string, name: string) => {
+                    br.setAttribute(name, value);
+                });
+            }
```
dmo-odoo: It's a bit less weird for the rendering but I'm still wondering why though.

bundles/BasicEditor.ts
```diff
@@ -5,12 +5,27 @@ import { LineBreak } from '../packages/plugin-linebreak/LineBreak';
...
+                Dom,
+                Char,
+                LineBreak,
+                Heading,
+                Paragraph,
+                List,
+                Span,
+                Bold,
+                Italic,
+                Underline,
```
dmo-odoo: It's a bit weird that this list does not include `Format` as a Plugin since it's the one who is suppose to declare the `InlineNode` that `CharNode` is supposed to extend, but maybe that is introduced in a later commit ?

packages/core/src/Format.ts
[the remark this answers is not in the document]
dmo-odoo: `Inline` is the `Format` plugin since formats are not in core.

packages/plugin-char/Char.ts
```diff
@@ -61,14 +62,12 @@ export class Char extends JWPlugin {
...
             const renderedFormats = [];
-            Object.keys(node.format).forEach(type => {
-                if (node.format[type]) {
-                    const formatNode = document.createElement(Format.toTag(type));
-                    renderedFormats.push(formatNode);
-                    parent.appendChild(formatNode);
-                    // Update the parent so the text is inside the format node.
-                    parent = formatNode;
-                }
+            Object.values(node.format || {}).forEach(value => {
```
dmo-odoo: I still don't understand why we have to handle the case where some property might not be defined. Just define them in the class once and be done with it.

packages/core/src/CorePlugin.ts
```diff
@@ -136,4 +142,7 @@ export class CorePlugin extends JWPlugin {
             direction: params.vSelection.direction,
         });
     }
+    applyFormat(): void {
+        return;
```
dmo-odoo: I think the `Inline` plugin should define a `toggleFormat` command rather than `applyFormat`. It would look in the formats of the node to check if the format given in argument is already there, if yes remove them, if no, add it. (This would even allow more complex decisions for later on like the al's cases where we might detect that the formats that are already there have classes on them so we can't remove them so we need to toggle them off in a different way)

packages/plugin-char/Char.ts
```diff
@@ -120,7 +113,7 @@ export class Char extends JWPlugin {
     insertText(params: InsertTextParams): void {
         const range = params.range || this.editor.vDocument.selection.range;
         const text = params.text;
-        const format = this._getCurrentFormat();
+        const format = this.editor.vDocument.getCurrentFormat();
```
dmo-odoo: I think this could be done by the constructor of `InlineNode`.

packages/plugin-char/test/Char.test.ts
```diff
@@ -257,7 +257,7 @@ describe('plugin-char', () => {
                     await testEditor(BasicEditor, {
                         contentBefore: '[]a',
                         stepFunction: (editor: JWEditor) => {
-                            applyFormat(editor, new BoldFormat());
+                            applyFormat(editor, 'bold');
```
dmo-odoo: Well that's sad since I liked the old system better... ^^'

Zinston:
@dmo-odoo This requires discussion. I made a few commits (including a functional but dirty poc that I can clean up if approved) applying some of what you suggested, and answered the rest. I'm not sure about what it is exactly you are suggesting with `InlineNode` and the format sets.

packages/plugin-char/test/Char.test.ts
Zinston: But the old system doesn't make it possible for a separate plugin to implement another version of 'bold'...

Zinston:
@dmo-odoo After discussion I implemented a first version using `InlineNode`. There's _some_ ugly stuff in there but it's mostly dependent on changes brought about by the KeyMap, Parser and Renderer PRs.

Zinston:
Closed in favor of PR #99 .

PR #79, opened 2020-01-24
[IMP] core: Add a keymap mechanism

Goaman:
This keymap allows to:
- specify shortcuts from the config of the editor
- specify default shortcuts for plugins
- specify shortcuts and default shortcuts depending on the platform (mac
  or others)
- specify shortucts with javascript event 'key' or 'code'

### DIFF
--- a/packages/core/src/JWEditor.ts
+++ b/packages/core/src/JWEditor.ts
@@ -1,3 +1,4 @@
+import { BoundCommand, Keymap } from './Keymap';
 import { Dispatcher, CommandIdentifier, CommandArgs } from './Dispatcher';
 import { EventManager } from './EventManager';
 import { JWPlugin } from './JWPlugin';
@@ -6,11 +7,22 @@ import { VDocument } from './VDocument';
 import { CorePlugin } from './CorePlugin';
 import { Parser } from './Parser';
 
+export enum Platform {
+    MAC = 'mac',
+    PC = 'pc',
+}
+
+export interface Shortcut extends BoundCommand {
+    platform?: Platform;
+    pattern: string;
+}
+
 export interface JWEditorConfig {
     autoFocus?: boolean;
     debug?: boolean;
     theme?: string;
     plugins?: Array<typeof JWPlugin>;
+    shortcuts?: Shortcut[];
 }
 
 export class JWEditor {
@@ -24,6 +36,11 @@ export class JWEditor {
     vDocument: VDocument;
     autoFocus = false;
     parser = new Parser();
+    keymaps = {
+        default: new Keymap(),
+        user: new Keymap(),
+    };
+    _platform = navigator.platform.match(/Mac/) ? Platform.MAC : Platform.PC;
 
     constructor(editable?: HTMLElement) {
         this.el = document.createElement('jw-editor');
@@ -80,6 +97,9 @@ export class JWEditor {
         this.el.appendChild(this.editable);
         document.body.appendChild(this.el);
 
+        // Attach the keymaps to the editable.
+        this.editable.addEventListener('keydown', this._onKeydown.bind(this));
+
         // CorePlugin is a special mandatory plugin that handles the matching
         // between the core commands and the VDocument.
         this.addPlugin(CorePlugin);
@@ -135,6 +155,12 @@ export class JWEditor {
             Object.keys(plugin.commandHooks).forEach(key => {
                 this.dispatcher.registerHook(key, plugin.commandHooks[key]);
             });
+            // register the shortcuts for this plugin.
+            if (plugin.shortcuts) {
+                for (const shortcut of plugin.shortcuts) {
+                    this._loadShortcut(shortcut, this.keymaps.default);
+                }
+            }
             // Register the parsing functions of this plugin.
             if (pluginClass.parsingFunctions.length) {
                 this.parser.addParsingFunction(...pluginClass.parsingFunctions);
@@ -152,6 +178,11 @@ export class JWEditor {
         }
         if (config.plugins) {
             config.plugins.forEach(pluginClass => this.addPlugin(pluginClass));
+        }
+        if (config.shortcuts) {
+            for (const shortcut of config.shortcuts) {
+                this._loadShortcut(shortcut, this.keymaps.user);
+            }
         }
     }
 
@@ -176,6 +207,51 @@ export class JWEditor {
         this._originalEditable.id = this.editable.id;
         this._originalEditable.style.display = this.editable.style.display;
         this.el.remove();
+    }
+
+    //--------------------------------------------------------------------------
+    // Private
+    //--------------------------------------------------------------------------
+
+    /**
+     * Load a shortcut in the keymap depending on the platform.
+     *
+     * - If the shortuct has no platform property; load the shortuct in both
+     *   platform ('mac' and 'pc').
+     * - If the shortuct has no platform property and the current platform is
+     *   mac, modify the ctrl key to meta key.
+     * - If the shortuct has a platform property, only load the shortcut for
+     *   that platform.
+     * - If no `mapping.commandId` is declared, it means removing the shortcut.
+     *
+     * @param shortcut The shortuct definition.
+     * @param priority  The highest priority is the one that prevail.
+     */
+    _loadShortcut(shortcut: Shortcut, keymap: Keymap): void {
+        if (!shortcut.platform || shortcut.platform === this._platform) {
+            if (!shortcut.platform && this._platform === Platform.MAC) {
+                shortcut.pattern = shortcut.pattern.replace(/ctrl/gi, 'CMD');
+            }
+            keymap.bindShortcut(shortcut.pattern, shortcut.commandId, shortcut.commandArgs);
+        }
+    }
+
+    /**
+     * Listener added to the DOM that `execCommand` if a shortcut has been found
+     * in one of the keymaps.
+     *
+     * @param event
+     */
+    _onKeydown(event: KeyboardEvent): void {
+        const commandUser = this.keymaps.user.match(event);
+        const commandDefault = !commandUser && this.keymaps.default.match(event);
+        const command = commandUser || commandDefault;
+        if (command && command.commandId) {
+            event.preventDefault();
+            event.stopPropagation();
+            event.stopImmediatePropagation();
+            this.execCommand(command.commandId, command.commandArgs);
+        }
     }
 }
 
--- a/packages/core/src/JWPlugin.ts
+++ b/packages/core/src/JWPlugin.ts
@@ -1,6 +1,7 @@
 import JWEditor from './JWEditor';
 import { ParsingFunction } from './Parser';
 import { CommandIdentifier, CommandDefinition, CommandHandler } from './Dispatcher';
+import { Shortcut } from './JWEditor';
 
 export interface JWPluginConfig {
     name?: string;
@@ -13,6 +14,7 @@ export class JWPlugin {
     editor: JWEditor;
     commands: Record<CommandIdentifier, CommandDefinition> = {};
     commandHooks: Record<CommandIdentifier, CommandHandler> = {};
+    shortcuts: Shortcut[];
 
     constructor(editor: JWEditor, options: JWPluginConfig = {}) {
         this.editor = editor;
--- a/packages/core/src/Keymap.ts
+++ b/packages/core/src/Keymap.ts
@@ -1,0 +1,171 @@
+import { CommandIdentifier, CommandArgs } from './Dispatcher';
+
+export interface BoundCommand {
+    /**
+     * An undefined `commandId` effectively unbinds the shortcut.
+     */
+    readonly commandId: CommandIdentifier | undefined;
+    readonly commandArgs?: CommandArgs;
+}
+
+interface ModifiedShortcut {
+    modifiers: Set<string>;
+}
+
+interface KeyShortcutPattern extends ModifiedShortcut {
+    key: string;
+}
+
+interface CodeShortcutPattern extends ModifiedShortcut {
+    code: string;
+}
+
+type ShortcutPattern = KeyShortcutPattern | CodeShortcutPattern;
+
+interface Shortcut {
+    pattern: ShortcutPattern;
+    boundCommand?: BoundCommand;
+}
+
+/**
+ * Keymap allow to add and remove shortucts and provide a function to match if a
+ * particular pattern is found within registred shortucts.
+ *
+ * ## Adding shortcuts
+ * The expression to describe a shortuct is zero or more `modifiers` and one
+ * `hotkey` joined with the symbol `+`.
+ *
+ * ### Modifiers
+ * - SHIFT
+ * - ALT
+ * - CTRL
+ * - META
+ * - CMD (alias of META)
+ *
+ * Example:
+ * ```typescript
+ * // using a modifier and one key
+ * keymap.bindShortuct('CTRL+A', 'commandIdentifier')
+ * // using multiples modifiers and one key
+ * keymap.bindShortuct('CTRL+ALT+A', 'commandIdentifier')
+ * ```
+ *
+ * ### Hotkeys
+ * A hotkey can be wether a `key` or `code`.
+ *
+ * #### Key
+ * The syntax to describe a `key` is to write the `key` as it is. The convention
+ * is to write it in uppercase.
+ *
+ * Example:
+ * ```typescript
+ * keymap.bindShortuct('CTRL+A', 'commandIdentifier')
+ * ```
+ * The list of possible `key` values are defined in the following link:
+ * https://www.w3.org/TR/uievents/#dom-keyboardevent-key
+ *
+ * #### Code
+ * The syntax to describe a `code` is to write `<code>` (surrounded by "<" and
+ * ">").
+ *
+ * Example:
+ * ```typescript
+ * keymap.bindShortuct('CTRL+<KeyA>', 'commandIdentifier')
+ * ```
+ *
+ * The list of possible `code` values are defined in the following link:
+ * https://www.w3.org/TR/uievents/#dom-keyboardevent-key
+ *
+ * ## Removing shortucts
+ * To remove a shortuct, call `bindShortcut` without commandIdentifier
+ *
+ * Example:
+ * ```typescript
+ * keymap.bindShortuct('CTRL+<KeyA>')
+ * ```
+ */
+export class Keymap {
+    shortcuts: Shortcut[] = [];
+
+    //--------------------------------------------------------------------------
+    // Public
+    //--------------------------------------------------------------------------
+
+    /**
+     * Add or remove a shortuct.
+     *
+     * If `commandId` is set, add the shortcut.
+     *
+     * If there is no `commandId`, it means that we want nothing to execute.
+     * Thus unbinding the command.
+     *
+     * @param patternString
+     * @param commandId
+     * @param commandArgs
+     */
+    bindShortcut(
+        patternString: string,
+        commandId?: CommandIdentifier,
+        commandArgs?: CommandArgs,
+    ): void {
+        this.shortcuts.unshift({
+            pattern: this.parsePattern(patternString),
+            boundCommand: { commandId, commandArgs },
+        });
+    }
+
+    /**
+     * Get first BoundCommand found when providing `keyEvent`.
+     *
+     * @param keyEvent
+     */
+    match(keyEvent: KeyboardEvent): BoundCommand {
+        const matchingMappings = this.shortcuts.find(shortcut => {
+            const modifiers = shortcut.pattern.modifiers;
+
+            let match: boolean;
+            if ('code' in shortcut.pattern) {
+                match = shortcut.pattern.code === keyEvent.code;
+            } else {
+                match = shortcut.pattern.key === keyEvent.key.toUpperCase();
+            }
+
+            match =
+                match &&
+                modifiers.has('CTRL') === keyEvent.ctrlKey &&
+                modifiers.has('SHIFT') === keyEvent.shiftKey &&
+                modifiers.has('META') === keyEvent.metaKey &&
+                modifiers.has('ALT') === keyEvent.altKey;
+
+            return match;
+        });
+        return matchingMappings && matchingMappings.boundCommand;
+    }
+
+    /**
+     * Parse a string that represend a pattern and return a `ShortuctPattern`.
+     * Supported pattern is: [modifier+]*[<code>|key]
+     *
+     * @param pattern
+     */
+    parsePattern(pattern: string): ShortcutPattern {
+        const tokens = pattern
+            .trim()
+            .replace(/cmd/gi, 'META')
+            .split(/[+]/);
+        const keyCode = tokens.pop();
+        const modifiers = new Set(tokens.map(token => token.toUpperCase()));
+        if (!keyCode) {
+            throw new Error('You must have at least one key or code.');
+        }
+        // There are two ways to specify a shortcut hotkey : key or code
+        // - "CTRL+1" is the modifier CTRL with the event.key "1".
+        // - "CTRL+<Key1>" is the modifier CTRL with the event.code "Key1"
+        const codeMatch = keyCode.match(/^<(\w+)>$/);
+        if (codeMatch && codeMatch.length > 1) {
+            return { code: codeMatch[1], modifiers };
+        } else {
+            return { key: keyCode.toUpperCase(), modifiers };
+        }
+    }
+}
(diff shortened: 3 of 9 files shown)


Goaman:
Main work is done. There is todo in the code for the question to be think when reviewing.

nohup.out
```diff
@@ -0,0 +1,4 @@
+Nemo-Share-Message: 13:48:21.967: Called "net usershare info" but it failed: 'net usershare' returned error 255: mkdir failed on directory /var/run/samba/msg.lock: Permission denied
+net usershare: cannot open usershare directory /var/lib/samba/usershares. Error No such file or directory
+Please ask your system administrator to enable user sharing.
```
dmo-odoo: Remember what I said last tuesday ? Do not re-read your code in place in the editor, re-read your *diff*. You would not have been able to miss this.

packages/core/src/CorePlugin.ts
```diff
@@ -61,6 +61,77 @@ export class CorePlugin extends JWPlugin {
             handler: this.toggleList.bind(this),
         },
     };
+    defaultShortcuts = [
+        //
+        // personal computer
+        //
```
dmo-odoo: What is the point of this comment ? Explaining what `pc` means as a `platform` ?

packages/core/src/CorePlugin.ts
```diff
@@ -61,6 +61,77 @@ export class CorePlugin extends JWPlugin {
...
+            pattern: 'ctrl shift <Digit8>',
+            commandId: 'toggleList',
+            commandArgs: { type: 'unordered' } as ListParams,
+        },
+        //
+        // macintosh
+        //
+        //
+        // personal computer
+        //
```
dmo-odoo: It's even weirder here since you have the two of them.

packages/core/src/JWEditor.ts
```diff
@@ -47,6 +49,8 @@ export class JWEditor {
         // The editable property of the editor is the original editable element
         // before start is called, and becomes the clone after start is called.
         this.editable = editable;
+
+        this._loadShortcut = this._loadShortcut.bind(this);
```
dmo-odoo: Wow.

packages/core/src/JWEditor.ts
```diff
@@ -73,13 +77,13 @@ export class JWEditor {
         this.el.appendChild(this.editable);
         document.body.appendChild(this.el);
 
+        // Init the event manager now that the cloned editable is in the DOM.
+        this.eventManager = new EventManager(this);
+
```
dmo-odoo: Why the move ? Just wondering.

packages/core/src/JWEditor.ts
```diff
@@ -122,6 +135,23 @@ export class JWEditor {
             this.debugger = new OwlUI(this);
             this.debugger.addPlugin(DevTools);
         }
+        if (config.shortcuts) {
+            const shortcutWithPriority = config.shortcuts.map(shortcut => ({
+                priority: 1,
+                ...shortcut,
+            }));
```
dmo-odoo: Can't we just have these ones override the others, bypassing the need for setting a `priority` ?

packages/core/src/JWPlugin.ts
```diff
@@ -1,17 +1,26 @@
...
 }
 
+export interface Shortcut {
+    platform?: string;
+    pattern: string;
+    commandId: CommandIdentifier;
+    commandArgs?: CommandArgs;
+    priority?: number;
+}
+
```
dmo-odoo: I'm surprised this interface is not in `KeyMap`.

packages/core/src/JWPlugin.ts
```diff
@@ -1,17 +1,26 @@
...
 export class JWPlugin {
     static readonly parsingFunctions: Array<ParsingFunction> = [];
     name: string;
     editor: JWEditor;
     commands: Record<CommandIdentifier, CommandDefinition> = {};
     commandHooks: Record<CommandIdentifier, CommandHandler> = {};
+    defaultShortcuts: Shortcut[];
```
dmo-odoo: I would have called it `shortcuts`. It is default by construction in the sense that they might be overriden later, but it's still the plugin's shortcut. You can never change the plugin shortcuts, what you can change is the *editor* shortcuts.

packages/core/src/Keymap.ts
```diff
@@ -0,0 +1,101 @@
...
+// todo: should I place this in file Command?
+export interface CommandCall {
+    commandId: string;
+    commandArgs?: CommandArgs;
+}
+
+interface ShortcutItem extends CommandCall {
+    pattern: ShortcutPattern;
+    priority: number;
+}
```
dmo-odoo: I'm not a fan of the two interfaces above. They are suspiciously similar to the `Shortcut` interface in `JWPlugin` that I expected to find here.

packages/core/src/Keymap.ts
```diff
@@ -0,0 +1,101 @@
...
+
+// todo: questions
+// - who preventDefault / stopPropagation / stopImmediatePropagation?
+// - what about when a pattern is already present in the system?
+//   - if the check (static or dynamic) is different?
+export class Keymap {
+    _shortcutItems: ShortcutItem[] = [];
+
+    //--------------------------------------------------------------------------
+    // public
```
dmo-odoo: Capitals. (You did it well for "Private" later in the same file)

packages/core/src/Keymap.ts
```diff
@@ -0,0 +1,101 @@
...
+    ): void {
+        this._shortcutItems.push({
+            pattern: this._getPatternFromString(patternString),
+            commandId,
+            commandArgs,
+            priority,
+        });
+    }
+
+    getCommandCalls(keyEvent: KeyboardEvent): CommandCall[] {
```
dmo-odoo: I would have called this function `match` but I'm 99.99% sure you're going to disagree.

packages/core/src/Keymap.ts
```diff
@@ -0,0 +1,101 @@
...
+    //--------------------------------------------------------------------------
+    // public
+    //--------------------------------------------------------------------------
+
+    addShortuct(
+        patternString: string,
+        commandId: CommandIdentifier,
+        commandArgs?: CommandArgs,
+        priority = 0,
+    ): void {
```
dmo-odoo: Can't we take a `Shortcut` object ? You know, the one from `JWPlugin` that I would like to see here... :p

packages/core/src/Keymap.ts
```diff
@@ -0,0 +1,101 @@
...
+            priority,
+        });
+    }
+
+    getCommandCalls(keyEvent: KeyboardEvent): CommandCall[] {
+        const possibleMappings = this._shortcutItems.filter((listener): boolean => {
+            const key = listener.pattern.key;
+            const modifiers = listener.pattern.modifiers;
+
+            if (key === '*') {
```
dmo-odoo: ?

packages/core/src/Keymap.ts
```diff
@@ -0,0 +1,101 @@
...
+            if (!modifierOk) {
+                return false;
+            }
+            const keyCodeMatch = key.match(/^<(\w+)>$/);
+            const keyCode = keyCodeMatch && keyCodeMatch[1];
+            if (keyCode) {
+                return keyCode === keyEvent.code;
+            } else {
+                return key === keyEvent.key;
+            }
```
dmo-odoo: I'm not sure I understand these `match`, `code` and `key` shenanigans...

packages/core/src/Keymap.ts
```diff
@@ -0,0 +1,101 @@
...
+        });
+        const maxPriority = Math.max.apply(null, possibleMappings.map(x => x.priority));
+        const prioritaryMapping = possibleMappings.filter(x => x.priority === maxPriority);
+        return prioritaryMapping.map(mapping => {
```
dmo-odoo: I'd still be in favor of an override mechanism. You're going to take the higher priority *anyway* so why even bother ? Just override the shortcut.

packages/core/src/JWEditor.ts
```diff
@@ -24,6 +25,7 @@ export class JWEditor {
     vDocument: VDocument;
     debugger: OwlUI;
     parser = new Parser();
+    _platform = navigator.platform.match('/Mac/') ? 'mac' : 'pc';
```
dmo-odoo: I'd make it public, especially considering you're setting it directly in the tests.

packages/core/src/Keymap.ts
```diff
@@ -0,0 +1,101 @@
...
+    //--------------------------------------------------------------------------
+    // Private
+    //--------------------------------------------------------------------------
+
+    _getPatternFromString(pattern: string): ShortcutPattern {
+        const patternArray = pattern
+            .trim()
+            .replace(/\s+/g, ' ')
+            .split(' ');
+        const key = patternArray.slice(-1)[0];
```
dmo-odoo: We might want to discuss the form of the shortcut. I remember @Zinston and I were not agreeing perfectly on the capitalizaion, but we both used a `+` rather than space.

packages/core/src/JWEditor.ts
Goaman: We need the event manager before loading a plugin because `_loadShortcut` need it

packages/core/src/JWEditor.ts
Goaman: I don't like to override the other because for multiples reason:
- We lose information that might be useful later (we could have an interface that shows the default shortcuts as well as the overridden shortcuts)
- If we add a plugin after loading the config, it will not work without extra work

packages/core/src/Keymap.ts
Goaman: The difference lies in the pattern that is not a `string` but a `ShortuctPattern`.
I think too much extends is not necessary. Here is the alternative:
```typescript
interface CommandCall {
    commandId: string;
    commandArgs?: CommandArgs;
}

interface ShortcutItemPriority extends CommandCall {
    priority: number;
}

interface ShortcutItem extends ShortcutItemPriority {
    pattern: ShortcutPattern;
}

interface Shortcut extends ShortcutItemPriority {
    pattern: string;
}
```
Do you prefer this version?

packages/core/src/Keymap.ts
Goaman: The `Shortcut` object has a `platform` property. Keymap should not know about the platform.
And here, the first argument is a pattern as string so it cannot have a `ShortcutItem`.
What do you suggest? Another interface for the function ?

packages/core/src/Keymap.ts
Goaman: Useless, I remove it

packages/core/src/Keymap.ts
Goaman: I added a comment. When we configure the shortcut with:
- "ctrl 1" means the modifier ctrl with the event.key "1" (dependent on the keyboard layout)
- "ctrl< Key1>" means the modifier ctrl with the event.code "Key1" (not dependent of the layout)

packages/core/src/Keymap.ts
Goaman: Overriding = losing information.

packages/core/src/JWEditor.ts
Goaman: 1) I'm setting it in test for mocking purpose.
2) Changing the platform after the editor has been initialized could lead to inconsistencies as it is used only to load shortcuts.
If we need to change the platform, it should be done with the config or a setter.
Changing the platform should reset the shortcuts.

packages/core/src/Keymap.ts
Goaman: I see any separator as equals. The one you decide will be the one used.

Zinston: We had eventually agreed on `Ctrl+a` form.

dmo-odoo: let's uppercase

packages/core/src/Keymap.ts
```diff
@@ -12,7 +12,8 @@ interface ShortcutPattern {
 
 interface ShortcutItem {
     pattern: ShortcutPattern;
-    boundCommand: BoundCommand;
+    // if there is no bound command, it means that we want to remove the ShortcutItem.
```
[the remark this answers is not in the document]
dmo-odoo: I think this might be one of these few cases where explicitely typing `undefined` might be useful. I would make it explicit in the `Shortcut` interface of `JWEditor` (and thus not having it extend `BoundCommand`), like this:
```typescript
export interface Shortcut {
    platform?: Platform;
    pattern: string;
    commandId: CommandIdentifier | undefined;
    commandArgs?: CommandArgs;
}
```
I think it might help as a "documentation" for creating shortcuts in plugins. (Just to be clear, it is fine that it is optional in the interface right here.)

packages/core/src/Keymap.ts
```diff
@@ -39,6 +40,13 @@ export class Keymap {
         });
     }
 
+    removeShortuct(patternString: string, priority = 0): void {
+        this._shortcutItems.push({
+            pattern: this._getPatternFromString(patternString),
+            priority,
```
dmo-odoo: Wait, you can "remove a shortcut" with a "priority" ? What does that even mean ? I think that when someone wants to remove a shortcut by setting its `commandId` to `undefined`, he clearly wants the shortcut free of any command, no matter what.

packages/core/src/Keymap.ts
```diff
@@ -66,7 +74,13 @@ export class Keymap {
...
         const maxPriority = Math.max(...matchingMappings.map(x => x.priority));
         const prioritaryMappings = matchingMappings.filter(x => x.priority === maxPriority);
-        return prioritaryMappings.map(mapping => mapping.boundCommand);
+        // undefined boundCommand means that we want to remove the shortuct.
+        const lastIndexUndefined = prioritaryMappings
+            .map(mapping => mapping.boundCommand)
+            .lastIndexOf(undefined);
+        return prioritaryMappings
+            .slice(lastIndexUndefined + 1)
+            .map(mapping => mapping.boundCommand);
```
dmo-odoo: I don't understand this. Now that I think about it, what is the point of triggering two commands for one shortcut ? I think I would change `this._shortcutItems.filter` above by `this._shortcutItems.find` such that it returns the "latest registered" matching one. This way, if the latest matching one is the `undefined` one ? Fine, don't trigger anything. That could even remove the need for `priority` altogether if we have multiple "layers" of shortcuts like we discussed yesterday. Two keymaps. One for plugins, one for user config. Look for a match in user config and if not found look into the plugins one. Use "last registered first" as tie-breaker, which means you can use `find` rather than `filter` above and stop as soon as you find a match. If the match you found is an "undefined" one, you just found a shortcut that was disabled, do not trigger anything. I think it would be a lot cleaner and easier to understand that it is now.

packages/core/test/JWeditor.test.ts
```diff
@@ -36,6 +36,76 @@ describe('core', () => {
                     );
                     expect(expectedCommands).to.eql(['command-all', 'command-pc']);
                 });
+                it('should transform ctrl to CMD if no platform on mac', () => {
+                    const editor = new JWEditor();
+                    editor._platform = 'mac';
+                    editor.start();
+                    // todo: to remove when the corePlugin will be removed from editor.start().
+                    editor.eventManager.keymap._shortcutItems = [];
```
dmo-odoo: Well actually, all the shortcutable commands from `CorePlugin` are not in `CorePlugin` anymore, so you should be able to remove this after you rebase on master.

packages/core/src/JWEditor.ts
```diff
@@ -186,6 +172,37 @@ export class JWEditor {
...
+
+    /**
+     * Load a shortcut in the keymap depending on the platform.
+     *
+     * - If the shortuct has no platform property; load the shortuct in both platform ('mac' and
+     *   'pc').
+     * - If the shortuct has no platform property and the current platform is mac, modify the ctrl
+     *   key to meta key.
+     * - If the shortuct has a platform property, only load the shortcut for that platform.
+     * - If no `mapping.commandId` is declared, it means removing the shortcut.
```
dmo-odoo: Well that confirms what I suspected about `removeShortcut`. It's never actually used, since removing a shortcut is still adding a shortcut anyway.

packages/core/src/JWEditor.ts
```diff
@@ -186,6 +172,37 @@ export class JWEditor {
...
     }
+
+    //--------------------------------------------------------------------------
+    // Private
+    //--------------------------------------------------------------------------
+
+    /**
+     * Load a shortcut in the keymap depending on the platform.
+     *
+     * - If the shortuct has no platform property; load the shortuct in both platform ('mac' and
```
dmo-odoo: This whole comment is not bound to 80 chars. We discussed this yesterday.

packages/core/src/Keymap.ts
```diff
@@ -17,13 +17,64 @@ interface ShortcutItem {
...
+ * - CTRL
+ * - META
+ * - CMD (alias of META)
+ *
+ * Example:
+ * ```typescript
+ * // using a modifier and one key
+ * keymap.addShortcut('CTRL+A', 'a-command')
+ * // using multiples modifiers and one key
+ * keymap.addShortcut('CTRL+ALT+A', 'a-command')
```
dmo-odoo: I would write "commandIdentifier" rather than "a-command" because I had to go back to the signature of the `addShortcut` function to understand what that was supposed to be.

packages/core/src/Keymap.ts
```diff
@@ -17,13 +17,64 @@ interface ShortcutItem {
...
+ * The list of possible `key` values are defined in the following link:
+ * https://www.w3.org/TR/uievents/#dom-keyboardevent-key
+ *
+ * ### Code
+ * The syntax to describe a `code` is to write `<code>` (surrounded by "<" and
+ * ">").
+ *
+ * Example:
+ * ```typescript
+ * keymap.addShortcut('<KeyA>', 'a-command')
```
dmo-odoo: I suggest adding an example with a modifier, as people might misunderstand and think they need to write `<CTRL+KeyA>` if they want to use a code, which would absolutely not work.

packages/core/src/Keymap.ts
```diff
@@ -17,13 +17,64 @@ interface ShortcutItem {
...
 export class Keymap {
     _shortcutItems: ShortcutItem[] = [];
 
     //--------------------------------------------------------------------------
     // Public
     //--------------------------------------------------------------------------
 
+    /**
+     * Add a shortuct.
+     */
```
dmo-odoo: No params doc ?

packages/core/src/Keymap.ts
```diff
@@ -40,13 +91,29 @@ export class Keymap {
...
 
+    /**
+     * Remove a shortcut.
+     *
+     * Removing a shortcut internally add another shortcut without `commandId`.
+     * The semantic of a shortcut without `commandId` is that it has been
+     * removed.  The reason is that if a lower priority define a shortuct, we
+     * need to have the information of a higher priority shortuct that contain
+     * the information of removal.
+     */
```
dmo-odoo: No params doc ?

packages/core/src/Keymap.ts
```diff
@@ -40,13 +91,29 @@ export class Keymap {
...
         });
     }
 
+    /**
+     * Get all BoundCommand found when providing `keyEvent`.
+     *
+     * 1) Search through all shortucts
+     * 2) Filter to only get shortucts with hihest priority
+     * 3) Filter to only get shortucts that were not removed
+     */
```
dmo-odoo: No params doc ? And as a general rule, it seems you like to describe the function inner workings in the docstring. I'm interested to know why. IMO, these three comment lines would be a lot more useful in the code of the function for these reasons:

- When I'm reading a function docstring, I am interested in what it does and how I can use it, not *how* it does it, I don't really care at that moment.
- When I'm reading the *code* of a function however, I don't want to go back and forth to the docstring trying to match the code I'm reading with one of the steps described up there. It allows you to put the comment at the exact point where it is relevant.

packages/core/src/Keymap.ts
```diff
@@ -87,6 +154,9 @@ export class Keymap {
     // Private
     //--------------------------------------------------------------------------
 
+    /**
+     * Parse a string that represend a pattern and return a `ShortuctPattern`.
+     */
```
dmo-odoo: No params doc ?

packages/core/src/Keymap.ts
```diff
@@ -160,9 +160,7 @@ export class Keymap {
     _getPatternFromString(pattern: string): ShortcutPattern {
         const patternArray = pattern
             .trim()
-            .replace(/\s+/g, ' ')
-            .replace(/ ?\+ ?/g, '+')
-            .replace(/[cC][mM][dD]/g, 'meta')
+            .replace(/cmd/gi, 'META')
             .split(/[+ ]/);
```
dmo-odoo: You're still splitting on either `+` or a space, here.

packages/core/src/JWEditor.ts
```diff
@@ -195,12 +195,16 @@ export class JWEditor {
...
+            if (shortcut.commandId) {
+                this.eventManager.keymap.addShortuct(
+                    shortcut.pattern,
+                    shortcut.commandId,
+                    shortcut.commandArgs,
+                    priority,
+                );
+            } else {
+                this.eventManager.keymap.removeShortuct(shortcut.pattern, priority);
+            }
```
dmo-odoo: I'm really not sure `removeShortcut` adds value. I mean, to remove a shortcut, people *have* to *create* a new shortcut that *does nothing* anyway, they *have* to do it that way, so why pretend in the code ?

packages/core/test/JWeditor.test.ts
```diff
@@ -168,7 +168,66 @@ describe('core', () => {
...
+                                pattern: 'ctrL+a',
+                                commandId: 'command-all',
+                            },
+                        ],
+                    });
+                    const priorityCommands = editor.eventManager.keymap._shortcutItems.map(
+                        item => item.priority,
+                    );
+                    expect(priorityCommands).to.eql([0, 1]);
+                });
```
dmo-odoo: One of your tests (maybe you have one already, I don't know) should call `addPlugin` *after* calling `loadConfig` to check that the plugin shortcuts do not override the config ones. (Though that would not be possible if we use two keymaps like I suggested earlier.)

packages/core/test/JWeditor.test.ts
Goaman: Some are still there

packages/core/src/JWEditor.ts
Goaman: remove `addShortuct` and `removeShortuct`, replace it with `bindShortcut`

packages/core/src/JWEditor.ts
```diff
@@ -24,6 +36,9 @@ export class JWEditor {
     vDocument: VDocument;
     autoFocus = false;
     parser = new Parser();
+    keymapDefault: Keymap;
+    keymapUser: Keymap;
```
dmo-odoo: `defaultKeymap` / `userKeymap` ? Not sure.

packages/core/test/JWeditor.test.ts
```diff
@@ -0,0 +1,310 @@
...
+import { BasicEditor } from '../../../bundles/BasicEditor';
+describe('core', () => {
+    describe('JWEditor', () => {
+        describe('addPlugin', () => {
+            describe('defaultShortcuts & loadConfig', () => {
+                it('should only register default mapping for pc and other', () => {
+                    const editor = new JWEditor();
+                    editor._platform = Platform.PC;
+                    editor.start();
+                    // todo: to remove when the corePlugin will be removed from editor.start().
```
dmo-odoo: ?

packages/core/test/JWeditor.test.ts
```diff
@@ -0,0 +1,310 @@
...
+                    const expectedCommands = editor.keymapDefault._shortcutItems.map(
+                        l => l.boundCommand.commandId,
+                    );
+                    expect(expectedCommands).to.eql(['command-pc', 'command-all']);
+                });
+                it('should transform ctrl to CMD if no platform on mac', () => {
+                    const editor = new JWEditor();
+                    editor._platform = Platform.MAC;
+                    editor.start();
+                    // todo: to remove when the corePlugin will be removed from editor.start().
```
dmo-odoo: ?

packages/core/test/JWeditor.test.ts
```diff
@@ -0,0 +1,310 @@
...
+                    const expectedCommands = editor.keymapDefault._shortcutItems.map(
+                        l => [...l.pattern.modifiers][0],
+                    );
+                    expect(expectedCommands).to.eql(['CTRL', 'META', 'META']);
+                });
+                it('should not transform ctrl to CMD if no platform on pc', () => {
+                    const editor = new JWEditor();
+                    editor._platform = Platform.PC;
+                    editor.start();
+                    // todo: to remove when the corePlugin will be removed from editor.start().
```
dmo-odoo: ? (you got my point)

packages/core/src/JWEditor.ts
dmo-odoo: keymaps: {
    default: Keymap,
    user: Keymap,
}

PR #81, opened 2020-01-27
[IMP] Examples: replace koi example with jabberwocky wikipedia + video

Zinston:


examples/jabberwockipedia/index.html
```diff
@@ -0,0 +1,3259 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>Jabberwocky - Wikipedia</title>
```
dmo-odoo: When it was `koi`, I named it `koi.html`. I suggest naming this file `jabberwockipedia.html`. It's easier to distinguish in the editor when you want to open the file. If all example files are named `index.html` then you have to pay attention to the folder in which they are, and I don't think the name actually matters for other things ?

examples/jabberwockipedia/jabberwockipedia.ts
```diff
@@ -0,0 +1,5 @@
+import { BasicEditor } from '../../bundles/BasicEditor';
+
+const editor = new BasicEditor(document.getElementById('example'));
+editor.loadConfig({ debug: true });
```
dmo-odoo: This does not exist anymore since ... 3 minutes ? :p You now need to `addPlugin(DevTools)`

Zinston: :trollface:

PR #84, opened 2020-01-28
[FIX] Core: deleteSelection with nested blocks

Zinston:


dmo-odoo:
I'm closing this PR since the bug it fixes has been fixed in https://github.com/odoo-dev/jabberwock/pull/87. Feel free to open a new PR for the _repr changes you wanted to introduce, though I suggest to handle the case of atomic nodes in `VNode` rather than introduce `AtomicNode`.

PR #85, opened 2020-01-29
Master indent list nby

Goaman:
We need to talk about in which plugin to store the logic. In indent or List


dmo-odoo:
I would put that in List. The fact that they both indent stuff is irrelevant IMO. This one is very particular to List.

Goaman:
need review

dmo-odoo:
Merged in https://github.com/odoo-dev/jabberwock/pull/104

PR #91, opened 2020-02-04
[ADD] Toolbar: basic static toolbar

Zinston:


Zinston:
@dmo-odoo Ok so it's functional but to avoid infinite loops due to '*' `commandHooks` I tried something a bit strange: command hooks targeting ids with regexes. It's a bit of a dirty commit for the moment but it's exploratory. In my case, it makes it possible for instance to rerender after every command except the 'query commands' (which are just commands prefixed by "query").

Zinston:
Closed in favor of #103

PR #93, opened 2020-02-05
[ADD] Core: apply paragraph (change to paragraph or any kind of heading)

Zinston:


packages/plugin-heading/Heading.ts
```diff
@@ -1,9 +1,47 @@
...
+            pattern: 'CTRL+SHIFT+<Digit5>',
+            commandId: 'applyParagraph',
+            commandArgs: { nodeCreator: (): HeadingNode => new HeadingNode(5) },
+        },
+        {
+            pattern: 'CTRL+SHIFT+<Digit6>',
+            commandId: 'applyParagraph',
+            commandArgs: { nodeCreator: (): HeadingNode => new HeadingNode(6) },
+        },
+    ];
```
dmo-odoo: Why not `applyHeading` then give `level` as argument ? And in `Paragraph` you would have `applyParagraph` with no argument ? Why do they have to share a common api ? That being said, I understand that it has a certain value of unifying the commands, which is great. Then we might want to discuss about the name `applyParagraph`. Google Docs call that tool "heading style" so I suggest `applyHeadingStyle` ? In my opinion, it makes sense that this command would only be provided by the `Heading` plugin, since it makes no sense to have it if you only have `Paragraph`. Of course that would make `Heading` depend on `Paragraph` but I don't see that as an issue. Tell me if you disagree. Then the `applyHeadingStyle` command could instantiate a `Paragraph` when argument `level` is `0`, so no more `nodeCreator` !

packages/plugin-heading/Heading.ts
```diff
@@ -17,4 +55,29 @@ export class Heading extends JWPlugin {
...
+     */
+    applyParagraph(params: ParagraphParams): void {
+        let heading = params.nodeCreator(); // todo: make sure to keep attributes
+        if (heading.is(HeadingNode)) {
+            const range = params.range || this.editor.vDocument.selection.range;
+            let nodesToConvert: VNode[];
+            if (range.isCollapsed()) {
+                nodesToConvert = [range.start.parent];
+            } else {
+                const selectedLeaves = range.selectedNodes(node => !node.children.length);
```
dmo-odoo: We have `isLeaf` in `VNode.ts` specifically for these use cases. (Basically we tend to avoid custom predicates as much as possible.) Though `isLeaf` will be `true` if the node only contains markers, while yours will be `false`. Does it matter ?

packages/plugin-heading/Heading.ts
```diff
@@ -17,4 +55,29 @@ export class Heading extends JWPlugin {
...
+                nodesToConvert = distinct(
+                    selectedLeaves.map(leaf => (leaf.atomic ? leaf.parent : leaf)),
+                );
```
dmo-odoo: It's not the first time you have to do this construction. I have a hard time finding a good name for it but I think we will have to. I'm thinking something like `closestNonAtomic(selectedLeaves)` but it's not pretty. 🆘

packages/plugin-heading/Heading.ts
```diff
@@ -17,4 +55,29 @@ export class Heading extends JWPlugin {
...
+            }
+            for (const node of nodesToConvert) {
+                node.before(heading);
+                node.mergeWith(heading);
+                heading = params.nodeCreator();
```
dmo-odoo: If the parameter is `level` rather `nodeCreator`, you can just create the new node at the top of the `for`. Actually, you could have done that even now, since you are not using `heading` before the `for`. It would have been easier to read than creating the "next" heading node at the end of the "current" loop iteration.

packages/plugin-paragraph/Paragraph.ts
```diff
@@ -17,4 +30,29 @@ export class Paragraph extends JWPlugin {
...
+                    selectedLeaves.map(leaf => (leaf.atomic ? leaf.parent : leaf)),
+                );
+            }
+            for (const node of nodesToConvert) {
+                node.before(paragraph);
+                node.mergeWith(paragraph);
+                paragraph = params.nodeCreator();
+            }
+        }
+    }
```
dmo-odoo: If we consider that `applyHeadingStyle` cannot exist without `Heading` installed, then this function can be removed. I mean, what would the toolbar look like if you only have Paragraph in the select and not Heading ? ^^'

PR #94, opened 2020-02-06
[REF] VNode: replace shallowDuplicate by clone

dmo-odoo:


dmo-odoo:
@Zinston I found a way to properly type the `constructor` of `VNode` without having to resort to another property name. Let me know is the end result suits your needs ;-)

PR #99, opened 2020-02-10
[ADD] Format: handle formats and attributes

Zinston:
See for PR #77 for prior discussion.


packages/plugin-inline/Inline.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * Apply the `format` to the selection.
+     *
+     * If the selection is collapsed, set the format on the selection so we know
+     * in the next insert which format should be used.
+     */
+    toggleFormat(params: FormatParams): void {
+        const format = params.format;
+        const range = this.editor.vDocument.selection.range;
```
dmo-odoo: params.range || this.editor.vDocument.selection.range;

packages/plugin-inline/Format.ts
```diff
@@ -0,0 +1,45 @@
...
+    render(): Node {
+        const node = document.createElement(this.htmlTag);
+        if (this.attributes) {
+            this.attributes.forEach((value: string, name: string) => {
+                node.setAttribute(name, value);
+            });
+        }
+        return node;
+    }
+    shallowDuplicate(): Format {
```
dmo-odoo: rename into `clone` to match `VNode`

packages/plugin-inline/Inline.ts
```diff
@@ -0,0 +1,106 @@
...
+        if (range.isCollapsed()) {
+            if (!this.formatCache) {
+                this.formatCache = this.getCurrentFormats();
+            }
+            const index = this.formatCache.findIndex(f => f.name === format.name);
+            if (index !== -1) {
+                this.formatCache.splice(index, 1);
+            } else {
+                this.formatCache.push(format.shallowDuplicate());
```
dmo-odoo: Why not push it directly ? Just wondering. (Sidenote: clone on VNode should clone the attributes)

packages/plugin-inline/Inline.ts
```diff
@@ -0,0 +1,106 @@
...
+            }
+        } else {
+            const selectedInlines = range.selectedNodes(InlineNode);
+
+            // If every char in the range has the format `format.name`, remove
+            // the format for all of them.
+            const allHaveFormat = selectedInlines.every(inline => {
+                return !!inline.formats.find(f => f.name === format.name);
```
dmo-odoo: Let's reignite the `instanceof` vs `name` discussion ! 🔥

packages/plugin-inline/Inline.ts
```diff
@@ -0,0 +1,106 @@
...
+            });
+            if (allHaveFormat) {
+                selectedInlines.forEach(inline => {
+                    const index = inline.formats.findIndex(f => f.name === format.name);
```
dmo-odoo: What if there are multiple formats of the same name ?

packages/plugin-inline/Inline.ts
```diff
@@ -0,0 +1,106 @@
...
+                    const matchingFormat = inline.formats[index];
+                    if (matchingFormat.attributes && matchingFormat.attributes.size) {
+                        matchingFormat.attributes.forEach((value: string, key: string) => {
+                            if (!inline.attributes) {
+                                inline.attributes = new Map<string, string>();
```
dmo-odoo: That's the magic which is going to keep the `b` classes on the `span` when rendered, right ?

packages/plugin-inline/Inline.ts
```diff
@@ -0,0 +1,106 @@
...
+        } else {
+            const selectedInlines = range.selectedNodes(InlineNode);
+
+            // If every char in the range has the format `format.name`, remove
+            // the format for all of them.
+            const allHaveFormat = selectedInlines.every(inline => {
+                return !!inline.formats.find(f => f.name === format.name);
+            });
+            if (allHaveFormat) {
+                selectedInlines.forEach(inline => {
```
dmo-odoo: forEach

packages/plugin-inline/Inline.ts
```diff
@@ -0,0 +1,106 @@
...
+                            }
+                            inline.attributes.set(key, value);
+                        });
+                    }
+                    inline.formats.splice(index, 1);
+                });
+                // If there is at least one char in the range without the format
+                // `format.name`, set the format for all nodes.
+            } else {
+                selectedInlines.forEach(inline => {
```
dmo-odoo: forEach

packages/plugin-inline/Inline.ts
```diff
@@ -0,0 +1,106 @@
...
+                    if (!inline.formats.find(f => f.isSameAs(format))) {
```
dmo-odoo: It's a bit weird because different nodes might have `format` in different places in their `formats` array.
Consider this:
`<b><i>ab</i></b>[c]<b><i>de</i></b>`

Toggle format bold should give (I know we're not handling this yet but bear with me):
`<b><i>ab</i>[c]<i>de</i></b>`

While this:
`<i><b>ab</b></i>[c]<i><b>de</b></i>`
Toggle format bold should give:
`<i><b>ab</b></i><b>c</b><i><b>de</b></i>`

What I mean is that the order in the `formats` array matters, doesn't it ?

packages/plugin-inline/Inline.ts
```diff
@@ -0,0 +1,106 @@
...
+        if (this.formatCache) {
+            return this.formatCache;
+        } else if (range.isCollapsed()) {
+            const inlineToCopyFormat = range.start.previousSibling() || range.start.nextSibling();
+            if (inlineToCopyFormat && inlineToCopyFormat.is(InlineNode)) {
+                formats = [...inlineToCopyFormat.formats];
+            }
+        } else {
+            const selectedInlines = range.selectedNodes(InlineNode);
+            selectedInlines.forEach(inline => (formats = [...formats, ...inline.formats]));
```
dmo-odoo: I'm not sure this is right.
Consider this:
`<b>ab[cdef</b>ghijkl<i>mnopq</i>rstu<u>vwx]yz</u>`

And now you insert 'k'.

If I understood the code correctly, it's going to take all the formats of the inlines in the selection, that is `b`, `i,` and `u`. I *think* I would have expected to only write in `b` since `range.start` is only affected by bold.

packages/plugin-inline/InlineFormatDomRenderer.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export class InlineFormatDomRenderer extends AbstractRenderer<Node[]> {
+    static id = 'dom';
+    predicate: Predicate<boolean | VNode> = InlineNode;
+    async render(node: InlineNode): Promise<Node[]> {
+        // If the node has a format, render the format nodes first.
+        const fragment = document.createDocumentFragment();
+        let parent: Node = fragment;
+        const renderedFormats = [];
+        Object.values(node.formats || {}).forEach(value => {
```
dmo-odoo: Can't we use `for...of` for objects ?

packages/plugin-inline/InlineFormatDomRenderer.ts
```diff
@@ -0,0 +1,38 @@
...
+            const formatNode = value.render();
+            renderedFormats.push(formatNode);
+            parent.appendChild(formatNode);
+            // Update the parent so the text is inside the format node.
+            parent = formatNode;
+        });
+        // If the node has attributes, wrap it inside a span with those
+        // attributes.
+        // TODO: This is specific to Char nodes because text nodes cannot have
+        // attributes so this should move there.
```
dmo-odoo: Can't this be moved yet ?

packages/plugin-inline/InlineFormatDomRenderer.ts
```diff
@@ -0,0 +1,38 @@
...
+        if (node.attributes && node.attributes.size) {
+            const span = document.createElement('span');
+            node.attributes.forEach((value: string, name: string) => {
```
dmo-odoo: forEach

packages/plugin-inline/Inline.ts
Zinston: Out of scope but we should handle order.

packages/plugin-inline/Inline.ts
Zinston: We only need take the format of the first char in the range.

packages/plugin-inline/InlineFormatDomRenderer.ts
Zinston:
```typescript
for (const value of Object.values(node.formats || {})) {
```

packages/plugin-inline/InlineNode.ts
```diff
@@ -0,0 +1,16 @@
...
+export class InlineNode extends VNode {
+    formats: Format[] = [];
+    get name(): string {
+        let formatNames = '';
+        if (this.formats.length) {
+            formatNames = '.' + this.formatNames.join('.');
+        }
+        return this.constructor.name + formatNames;
+    }
+    get formatNames(): string[] {
```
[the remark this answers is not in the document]
Zinston: image

packages/plugin-italic/Italic.ts
```diff
@@ -0,0 +1,15 @@
...
+import { FormatParams } from '../plugin-inline/Inline';
+import { ItalicDomParser } from './ItalicDomParser';
+
+export class Italic extends JWPlugin {
+    readonly parsers = [ItalicDomParser];
+    shortcuts = [
+        {
+            pattern: 'CTRL+I',
+            commandId: 'toggleFormat',
+            commandArgs: { format: new ItalicFormat() } as FormatParams,
```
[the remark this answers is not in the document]
Zinston: None.

packages/utils/src/utils.ts
```diff
@@ -16,6 +16,23 @@ export function isBlock(node: Node): boolean {
...
 }
 
+/**
+ * Return true if Map object a is deep equal to Map object b, false otherwise.
+ * 
+ * @param a
+ * @param b
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function deepEqualMaps(a: Map<any, any>, b: Map<any, any>): boolean {
```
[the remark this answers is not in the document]
Zinston: image

packages/core/src/VNodes/VNode.ts
```diff
@@ -30,12 +30,12 @@ export class VNode {
     readonly type: VNodeType;
     readonly id = id;
     parent: VNode;
+    attributes = new Map<string, string>();
```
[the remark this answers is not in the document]
Zinston: Problem: then we have
```typescript
// eslint-disable-next-line @typescript-eslint/no-unused-vars
parseAttributes(item: T): Record<string, string> {
    return {};
}
```
in `ParsingEngine`?

packages/plugin-char/CharNode.ts
```diff
@@ -84,9 +65,16 @@ export class CharNode extends VNode {
...
+            // Char VNodes are the same text node if they have the same format,
+            // attributes and format attributes.
+            const attributes = [...this.attributes.keys(), ...node.attributes.keys()];
+            return (
+                this.formats.length === node.formats.length &&
+                this.formats.every(aFormat => {
+                    return !!node.formats.find(bFormat => bFormat.isSameAs(aFormat));
+                }) &&
+                attributes.every(k => this.attributes.get(k) === node.attributes.get(k))
+            );
```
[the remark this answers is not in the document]
Zinston: Comments below address this order question. Seems like this would be out of scope but good to think about...

packages/plugin-inline/Inline.ts
Zinston: I don't think I changed the general logic of this, just the fact that formats are now classes instead of strings...

packages/plugin-inline/Inline.ts
Zinston:
```typescript
/**
 * Get the format for the next insertion.
 */
getCurrentFormats(range = this.editor.vDocument.selection.range): Format[] {
    let formats: Format[] = [];
    if (this.formatCache) {
        return this.formatCache;
    } else {
        // If the range has a previous sibling, take its formats.
        if (previousSibling && previousSibling.is(InlineNode)) {
            formats = [...previousSibling.formats];
        }
    }
    return formats;
}
```

Zinston: So actually this is right because it is the current format. It's in `insertText` that we need to make sure to take the current format _after_ emptying the range.

packages/plugin-char/Char.ts
```diff
@@ -53,82 +32,28 @@ export class Char extends JWPlugin {
...
      */
     insertText(params: InsertTextParams): void {
         const range = params.range || this.editor.vDocument.selection.range;
         const text = params.text;
-        const format = this._getCurrentFormat();
+        const inlinePlugin = this.editor.plugins.find(plugin => plugin instanceof Inline) as Inline; // todo: some form of hook...
+        let formats = [];
+        if (inlinePlugin) {
+            formats = inlinePlugin.getCurrentFormats();
+        }
```
dmo-odoo: Move the `getCurrentFormats` and `formatCache` logic into `Char` since it is actually very specific to text insertion. You can even call it something like `insertTextFormat` or something similar.

packages/plugin-dom/Dom.ts
```diff
@@ -221,8 +222,12 @@ export class Dom extends JWPlugin {
 
         let node = this.editor.vDocument.root.lastLeaf();
         while (node) {
-            const renderedNode = (await this.editor.renderers.dom.render(node)) as Node[];
-            for (const domNode of renderedNode) {
+            const domEngine = this.editor.renderers.dom as RenderingEngine<Node[]>;
+            await domEngine.render(node);
+            const renderings = domEngine.renderings.get(node);
+            const rendering = renderings[renderings.length - 1];
+            const renderedNodes = await rendering[1];
```
dmo-odoo: You don't have to `await domEngine.render(node)` if you already `await rendering[1]` anyway. Just call render to be sure that the rendering is triggered if it was not already, but you don't have to wait for it. We'll gain one microtask tick per node of the tree. Not much, I agree, but we tend to have a lot of nodes.

packages/plugin-inline/Format.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+    clone(): Format {
+        const clone = new (this.constructor as typeof Format)();
+        clone.htmlTag = this.htmlTag;
+        clone.attributes = { ...this.attributes };
+        return clone;
+    }
+    isSameAs(otherFormat: Format): boolean {
+        return (
+            this.constructor.name === otherFormat.constructor.name &&
```
dmo-odoo: `instanceof` ? I wanted to remove `name` altogether but you're using it in DevTools. Though when we merge (back) formats and attributes into a single concept, the `charName.formatName` feature of the DevTools won't make sense anymore so we'll remove it at that point.

packages/plugin-inline/Inline.ts
```diff
@@ -0,0 +1,102 @@
...
+            // the format for all of them.
+            const allHaveFormat = selectedInlines.every(inline => {
+                return !!inline.formats.find(f => f instanceof FormatClass);
+            });
+            if (allHaveFormat) {
+                for (const inline of selectedInlines) {
+                    const index = inline.formats.findIndex(f => f instanceof FormatClass);
+                    const matchingFormat = inline.formats[index];
+                    for (const key of Object.keys(matchingFormat.attributes)) {
+                        inline.attributes[key] = matchingFormat.attributes[key];
```
dmo-odoo: This is "extending" an existing b with classes, right ?
So:
```<b class"dmo">text</b>[ more text]```
Becomes:
```<b class"dmo">text[ more text]</b>```
Instead of:
```<b class"dmo">text</b><b>[ more text]</b>```

Is this what that is doing ?

If yes then what about:
```<b class"dmo">text</b>[ more ]<b class="chm">text</b>```

Re-reading the code, I'm not sure this is what it does... I think a comment would be great.

File: packages/plugin-inline/Inline.ts [hunk 23]
```diff
@@ -0,0 +1,102 @@
...
+                    }
+                    inline.formats.splice(index, 1);
+                }
+                // If there is at least one char in the range without the format
+                // `FormatClass`, set the format for all nodes.
```
dmo-odoo: You meant to put this comment in the `else` clause, right ? Otherwise it is not commenting any code.

packages/plugin-inline/Inline.ts
dmo-odoo: I think I would change the `else` clause of `getCurrentFormats` to:
```typescript
const inlineToCopyFormat = range.start.nextSibling();
if (inlineToCopyFormat && inlineToCopyFormat.is(InlineNode)) {
    formats = [...inlineToCopyFormat.formats];
}
```

I think it would solve the issue I was pointing and would work for your test as well.

Added bonus: the `if` and `else` clauses would become so similar that the entire function could be simplified to:
```typescript
if (this.formatCache) {
    return this.formatCache;
}

let inlineToCopyFormat: VNode;
if (range.isCollapsed()) {
    inlineToCopyFormat = range.start.previousSibling() || range.start.nextSibling();
} else {
    inlineToCopyFormat = range.start.nextSibling();
}
if (inlineToCopyFormat && inlineToCopyFormat.is(InlineNode)) {
    return [...inlineToCopyFormat.formats];
}

return [];
```

And at this point I'm not even sure we would need a format cache since computing this would not be so expensive (but we still want inserting text to be quite fast so we probably somewhat fear to remove it, though.)

packages/plugin-inline/InlineNode.ts
```diff
@@ -0,0 +1,13 @@
+import { VNode } from '../core/src/VNodes/VNode';
+import { Format } from './Format';
+
+export class InlineNode extends VNode {
+    formats: Format[] = [];
+    get name(): string {
+        let name = this.constructor.name;
+        for (const format of this.formats) {
+            name += '.' + format.name;
```
dmo-odoo: Wait, don't you have the same logic there ?

https://github.com/odoo-dev/jabberwock/pull/99/commits/e57e8ef3ce4a178b58a890d18500cb3010dc48b9#diff-00e933b8cab4fdbb08b9bc11caf03ceeR10

Isn't that going to end up appending the format names twice ?

packages/plugin-inline/FormatDomParser.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export abstract class FormatDomParser extends AbstractParser<Node> {
+    static id = 'dom';
+
+    /**
+     * Parse a span node.
+     *
+     * @param nodes
+     */
+    applyFormat(format: Format, nodes: VNode[]): void {
```
dmo-odoo: Well now that I think about it, this function might go into an abstract `FormatParser` rather than `FormatDomParser`, because it has nothing to do with DOM, it operates only on our virtual representation. So basically, I'd rename this class into `FormatParser` and put back the `static id` in the different parsers.

packages/plugin-linebreak/LineBreakDomParser.ts
```diff
@@ -5,16 +5,18 @@ export class LineBreakDomParser extends AbstractParser<Node> {
...
+        return item instanceof Element && item.nodeName === 'BR';
     };
 
-    async parse(item: Node): Promise<LineBreakNode[]> {
+    async parse(item: Element): Promise<LineBreakNode[]> {
         if (!item.nextSibling) {
             // A <br/> with no siblings is there only to make its parent visible.
             // Consume it since it was just parsed as its parent element node.
             // TODO: do this less naively to account for formatting space.
             return [];
```
dmo-odoo: You used to handle attributes on the second `BR`. I don't think you have to, but if you want to, you could use the same "hack" as the attributes on `LI`. (With the same shortcomings that it should be removed if the situation around the LineBreak changes.)

packages/plugin-linebreak/LineBreakDomRenderer.ts
```diff
@@ -10,6 +10,9 @@ export class LineBreakDomRenderer extends AbstractRenderer<Node[]> {
      */
     async render(node: LineBreakNode): Promise<Node[]> {
         const rendering: Node[] = [document.createElement('br')];
+        for (const name of Object.keys(node.attributes)) {
+            (rendering[0] as Element).setAttribute(name, node.attributes[name]);
+        }
```
dmo-odoo: `(rendering[0] as Element)`, seriously ?
```typescript
const br = document.createElement('br');
for (const name of Object.keys(node.attributes)) {
    br.setAttribute(name, node.attributes[name]);
}
const rendering = [br];
```

packages/plugin-inline/Format.ts
```diff
@@ -0,0 +1,43 @@
...
+            deepEqualObjects(this.attributes, otherFormat.attributes)
```
dmo-odoo: Sad but I think we'll be able to remove it when we merge (back) formats and attributes. (Since it will be replaceable by recursive calss to `isSameAs`.)

packages/plugin-inline/Inline.ts
```diff
@@ -0,0 +1,102 @@
...
+            } else {
+                for (const inline of selectedInlines) {
+                    if (!inline.formats.find(f => f instanceof FormatClass)) {
+                        inline.formats.push(new FormatClass());
```
dmo-odoo: Wait so `formatCache` gets a `new FormatClass()` but `inline.formats` gets another `new FormatClass()` ? Shouldn't they share the same instance to be coherent with the behavior of `getCurrentFormats` ?

packages/plugin-inline/test/Inline.test.ts
```diff
@@ -0,0 +1,161 @@
...
+                    contentAfter: '<b>b[]</b>a',
+                });
+            });
+            it('should not make bold the next insertion when toggleFormat 2 times', async () => {
+                await testEditor(BasicEditor, {
+                    contentBefore: '[]a',
+                    stepFunction: async (editor: JWEditor) => {
+                        await toggleFormat(editor, BoldFormat);
+                        await toggleFormat(editor, BoldFormat);
+                        await insertText(editor, 'b');
```
dmo-odoo: I don't think moving these tests made sense. Now you're calling `insertText` in `Inline` which does not depend on `Char`, while previously you were calling `toggleFormat` in `Char`, which *does* depend on `Inline` so it was fine, now it's not.

packages/plugin-char/CharFormatDomRenderer.ts
```diff
@@ -19,7 +19,10 @@ export class CharFormatDomRenderer extends InlineFormatDomRenderer {
         if (attributeNames.length) {
             const span = document.createElement('span');
             for (const name of attributeNames) {
-                span.setAttribute(name, node.attributes[name]);
+                const value = node.attributes[name];
+                if (typeof value === 'string') {
+                    span.setAttribute(name, value);
+                }
```
dmo-odoo: And now *all* renderers are defiled by this `typeof`... :( Hopefully it will go away when merging (back) formats and attributes into a single concept.

packages/plugin-dom/DomParsingEngine.ts
```diff
@@ -10,9 +10,13 @@ export class DomParsingEngine extends ParsingEngine<Node> {
...
-            return attributes;
-        }, {});
+        if (node.attributes) {
+            return Array.from(node.attributes).reduce((attributes, attribute) => {
+                attributes[attribute.name] = attribute.value;
+                return attributes;
+            }, {});
+        } else {
+            return {};
+        }
```
dmo-odoo: What is the difference ?

packages/plugin-linebreak/LineBreakDomParser.ts
```diff
@@ -9,14 +10,38 @@ export class LineBreakDomParser extends AbstractParser<Node> {
...
+            return item.parentNode && this._nextLeafInSameBlock(item.parentNode);
+        } else if (isBlock(next)) {
+            return;
+        } else {
+            while (next.firstChild) {
+                next = next.firstChild;
+            }
+            return next;
+        }
+    }
```
dmo-odoo: Could we try this ?

```typescript
_isLastLeafOfBlock(node: Node): Node {
    let next = item.nextSibling;
    if (next) {
        return isBlock(next);
    } else {
        let result = next;
        while (!result.nextSibling && result.parentNode && isBlock(result.parentNode)) {
            result = result.parentNode;
        }
        return !result.nextSibling;
}
```

packages/plugin-char/CharFormatDomRenderer.ts
Zinston: I know... 
image

packages/plugin-dom/DomParsingEngine.ts
Zinston: We only read the node's attributes if it has any.
In retrospect I could have done `Array.from(node.attributes || []).reduce` instead. Would you have preferred that?

packages/plugin-linebreak/LineBreakDomParser.ts
Zinston: More like
```typescript
_isLastLeafOfBlock(node: Node): boolean {
    const next = node.nextSibling;
    if (next) {
        return isBlock(next);
    } else {
        let result: Node = next;
        while (!result.nextSibling && result.parentNode && isBlock(result.parentNode)) {
            result = result.parentNode;
        }
        return !result.nextSibling;
    }
}
```

dmo-odoo:
🎉

Zinston:
image

PR #104, opened 2020-02-16
Generalize some valuable code from List plugin

dmo-odoo:
This PR generalizes the splitting mechanism of `List` into a `split` method on `VRange`.

This PR also introduces the following concepts which are generically useful and used in `List`:
- `node.closest`
- `node.wrap`
- `node.unwrap`
- `range.targetedNodes`
- `range.extend`

This PR also adds the mechanism that appends a `FragmentNode`'s children instead of the `FragmentNode` itself when appended to another node, even though it actually is of no use to this PR. (To be fair, it used to, and it was missing from master anyway.)


packages/core/src/VNodes/VNode.ts
```diff
@@ -544,14 +544,12 @@ export class VNode {
...
         const descendants = [];
-        if (this.children.length) {
-            let currentDescendant = this.firstChild();
-            do {
-                if (currentDescendant.test(predicate)) {
-                    descendants.push(currentDescendant);
-                }
-                currentDescendant = this._descendantAfter(currentDescendant);
-            } while (currentDescendant);
+        let currentDescendant = this.firstChild();
```
Zinston: Isn't this a REF instead of a FIX?

packages/core/src/VNodes/VNode.ts
```diff
@@ -511,6 +511,18 @@ export class VNode {
         }
         return nextSiblings;
     }
+    /**
+     * Return the closest node from this node that matches the given predicate.
+     * Start with this node then go up the ancestors tree until finding a match.
+     *
+     * @param predicate
+     */
+    closest(predicate?: Predicate): VNode;
```
Zinston: `predicate` shouldn't be optional.

packages/core/src/VRange.ts
```diff
@@ -123,6 +106,31 @@ export class VRange {
...
+     * range is collapsed, the targeted node is the container of the range.
+     * Otherwise, the targeted nodes are the ones encountered by traversing the
+     * tree from the start to end of this range.
+     */
+    targetedNodes(predicate?: Predicate): VNode[];
+    targetedNodes<T extends VNode>(predicate?: Constructor<T>): T[];
+    targetedNodes<T>(predicate?: Predicate<T>): VNode[];
+    targetedNodes<T>(predicate?: Predicate<T>): VNode[] {
+        if (this.isCollapsed()) {
+            const target = this.start.parent;
```
Zinston: This should target _all_ ancestors, not just the parent. Otherwise quid of nested blocks?

packages/core/src/VRange.ts
```diff
@@ -233,6 +233,51 @@ export class VRange {
...
+     * Split the range containers up to their common ancestor. Return all
+     * children of the common ancestor that are targeted by the range after the
+     * split. If a predicate is given, splitting continues up to and including
+     * the node closest to the common ancestor that matches the predicate.
+     *
+     * @param predicate
+     */
+    split<T>(predicate?: Predicate<T>): VNode[] {
+        const ancestor = this.startContainer.commonAncestor(this.endContainer);
+        const closest = ancestor.closest(predicate);
```
Zinston: What if no predicate was given?

packages/core/src/VRange.ts
```diff
@@ -233,6 +233,51 @@ export class VRange {
...
+            }
+            start = startAncestor;
+        } while (start.parent !== container);
+
+        // Split the end ancestors.
+        let end = this.end;
+        do {
+            const endAncestor = end.parent;
+            const children = endAncestor.children;
+            if (children.indexOf(end) < children.length - 1) {
```
Zinston: Comment would be nice:
`// Don't split at the end edge of a node.`
But shouldn't you also check that we're not at the _start_ edge?

packages/core/src/VRange.ts
```diff
@@ -233,6 +233,51 @@ export class VRange {
...
+    split<T>(predicate?: Predicate<T>): VNode[] {
+        const ancestor = this.startContainer.commonAncestor(this.endContainer);
+        const closest = ancestor.closest(predicate);
+        const container = closest ? closest.parent : ancestor;
+
+        // Split the start ancestors.
+        let start = this.start;
+        do {
+            let startAncestor = start.parent;
+            if (startAncestor.children.indexOf(start) > 0) {
```
Zinston: Comment would be nice:
`// Don't split at the start edge of a node.`
But shouldn't you also check that we're not at the _end_ edge?

packages/core/src/VRange.ts
```diff
@@ -233,6 +233,51 @@ export class VRange {
...
+            }
+            end = endAncestor;
+        } while (end.parent !== container);
+
+        // Return all top-most split nodes between and including start and end.
+        const nodes = [];
+        let node = start;
+        while (node !== end) {
+            nodes.push(node);
+            node = node.nextSibling();
```
Zinston: Shouldn't it be `next`? What if the nodes have different parents?
_Brotha from an otha motha_ :call_me_hand:

packages/plugin-list/List.ts
```diff
@@ -84,57 +85,23 @@ export class List extends JWPlugin {
      * @param nodes
      * @param type
      */
-    _convertToList(nodes: Array<VNode>, type: ListType): void {
+    _convertToList(range, type: ListType): void {
```
Zinston: Wow nice :+1:

packages/plugin-list/List.ts
```diff
@@ -117,34 +113,18 @@ export class List extends JWPlugin {
...
-        // Get the direct children of each list to unlist.
-        const listItems = distinct(
-            nodes.map(node => {
-                if (node.parent && node.parent.is(ListNode)) {
-                    return node;
-                } else {
-                    return node.ancestor(ancestor => {
-                        return ancestor.parent && ancestor.parent.is(ListNode);
-                    });
+    _unlist(range): void {
```
Zinston: Re-nice :+1:

packages/plugin-list/List.ts
```diff
@@ -117,34 +113,18 @@ export class List extends JWPlugin {
...
+        withRange(VRange.extending(range), extendedRange => {
+            const nodes = extendedRange.split(ListNode);
+            for (const list of nodes) {
+                for (const nestedList of list.descendants(ListNode)) {
+                    // TODO: automatically invalidate `li-attributes`.
```
Zinston: :+1:

packages/core/src/VNodes/VNode.ts
dmo-odoo: Well, I just switched from a `do...while` to a `while` and removed the `if` that was useless. Somewhat a bit of both in some way but the reason I did it was to fix a bug. (and the true "refactor" part is just removing a useless if)

packages/core/src/VRange.ts
dmo-odoo: I don't think so because we return only the top-most containers that were split, so they are all consecutive siblings of the container we did not split, which is eithier their common ancestor or an ancestor of their common ancestor.

PR #106, opened 2020-02-21
[𝔸𝔻𝔻] 𝕍ℕ𝕠𝕕𝕖: 𝕋𝕒𝕓𝕝𝕖

Zinston:
Handle general table behavior.

Limitations:
- /!\ Does not include `unbreakable` so selection/backspace/delete/enter are momentarily broken in a table.
- Does not handle `<colgroup>`, `<tfoot>` and `<caption>`
- Add row/column only collapsed

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -115,7 +115,7 @@ module.exports = function(config) {
         singleRun: false,
 
         // Concurrency level
-        // how many browser should be started simultaneous
+        // how many browsers should be started simultaneously
         concurrency: Infinity,
         mine: {
             'text/x-typescript': ['ts'],
--- a/packages/plugin-table/src/Table.ts
+++ b/packages/plugin-table/src/Table.ts
@@ -8,10 +8,340 @@ import { TableDomRenderer } from './TableDomRenderer';
 import { TableRowDomRenderer } from './TableRowDomRenderer';
 import { TableCellDomRenderer } from './TableCellDomRenderer';
 import { Renderer } from '../../plugin-renderer/src/Renderer';
+import { TableCellNode } from './TableCellNode';
+import { CommandParams } from '../../core/src/Dispatcher';
+import { TableNode } from './TableNode';
+import { distinct } from '../../utils/src/utils';
+import { TableRowNode } from './TableRowNode';
+import { RelativePosition } from '../../core/src/VNodes/VNode';
+import { Keymap } from '../../plugin-keymap/src/Keymap';
 
 export class Table<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T> {
-    readonly loadables: Loadables<Parser & Renderer> = {
+    readonly loadables: Loadables<Parser & Renderer & Keymap> = {
         parsers: [TableDomParser, TableRowDomParser, TableCellDomParser],
         renderers: [TableDomRenderer, TableRowDomRenderer, TableCellDomRenderer],
+        shortcuts: [
+            {
+                pattern: 'CTRL+<Slash>',
+                commandId: 'addRowBelow',
+            },
+            {
+                pattern: 'CTRL+SHIFT+<Slash>',
+                commandId: 'addRowAbove',
+            },
+            {
+                pattern: 'CTRL+<Equal>',
+                commandId: 'addColumnAfter',
+            },
+            {
+                pattern: 'CTRL+SHIFT+<Equal>',
+                commandId: 'addColumnBefore',
+            },
+            {
+                pattern: 'CTRL+SHIFT+DELETE',
+                commandId: 'deleteRow',
+            },
+            {
+                pattern: 'CTRL+SHIFT+BACKSPACE',
+                commandId: 'deleteColumn',
+            },
+            {
+                pattern: 'CTRL+ALT+SHIFT+DELETE',
+                commandId: 'deleteTable',
+            },
+            {
+                pattern: 'CTRL+<Period>',
+                commandId: 'mergeCells',
+            },
+            {
+                pattern: 'CTRL+SHIFT+<Period>',
+                commandId: 'unmergeCells',
+            },
+        ],
     };
+    commands = {
+        addRowAbove: {
+            handler: this.addRowAbove.bind(this),
+        },
+        addRowBelow: {
+            handler: this.addRowBelow.bind(this),
+        },
+        addColumnBefore: {
+            handler: this.addColumnBefore.bind(this),
+        },
+        addColumnAfter: {
+            handler: this.addColumnAfter.bind(this),
+        },
+        deleteRow: {
+            handler: this.deleteRow.bind(this),
+        },
+        deleteColumn: {
+            handler: this.deleteColumn.bind(this),
+        },
+        deleteTable: {
+            handler: this.deleteTable.bind(this),
+        },
+        mergeCells: {
+            handler: this.mergeCells.bind(this),
+        },
+        unmergeCells: {
+            handler: this.unmergeCells.bind(this),
+        },
+    };
+
+    //--------------------------------------------------------------------------
+    // Public
+    //--------------------------------------------------------------------------
+
+    /**
+     * Add a row above the cell at given range start.
+     *
+     * @param params
+     */
+    addRowAbove(params: CommandParams): void {
+        const range = params.context.range;
+        const cell = range.start.ancestor(TableCellNode);
+        if (!cell) return;
+
+        cell.ancestor(TableNode).addRowAbove(cell);
+    }
+    /**
+     * Add a row below the cell at given range start.
+     *
+     * @param params
+     */
+    addRowBelow(params: CommandParams): void {
+        const range = params.context.range;
+        const row = range.start.ancestor(TableCellNode);
+        if (!row) return;
+
+        row.ancestor(TableNode).addRowBelow(row);
+    }
+    /**
+     * Add a column before the cell at given range start.
+     *
+     * @param params
+     */
+    addColumnBefore(params: CommandParams): void {
+        const range = params.context.range;
+        const cell = range.start.ancestor(TableCellNode);
+        if (!cell) return;
+
+        cell.ancestor(TableNode).addColumnBefore(cell);
+    }
+    /**
+     * Add a column after the cell at given range start.
+     *
+     * @param params
+     */
+    addColumnAfter(params: CommandParams): void {
+        const range = params.context.range;
+        const cell = range.start.ancestor(TableCellNode);
+        if (!cell) return;
+
+        cell.ancestor(TableNode).addColumnAfter(cell);
+    }
+    /**
+     * Delete the row at given range start.
+     *
+     * @param params
+     */
+    deleteRow(params: CommandParams): void {
+        const range = params.context.range;
+        const cell = range.start.ancestor(TableCellNode);
+        if (!cell) return;
+
+        const row = cell.ancestor(TableRowNode);
+        const nextRow = row.nextSibling(TableRowNode) || row.previousSibling(TableRowNode);
+        const nextCell = nextRow && nextRow.children(TableCellNode)[cell.columnIndex];
+
+        if (nextCell) {
+            const nextRowIndex = nextCell.rowIndex;
+
+            // Handle rowspans.
+            const cells = row.children(TableCellNode);
+            for (const cell of cells) {
+                if (cell.rowspan > 1) {
+                    // Cells managed by this cell will now be managed by the
+                    // cell below (or above if there is none below) instead.
+                    const belowCell = Array.from(cell.managedCells).find(
+                        managedCell => managedCell.rowIndex === nextRowIndex,
+                    );
+                    if (belowCell) {
+                        belowCell.unmerge();
+                        for (const managedCell of cell.managedCells) {
+                            if (managedCell !== belowCell) {
+                                managedCell.mergeWith(belowCell);
+                            }
+                        }
+                    }
+                } else if (!cell.isActive()) {
+                    // If this cell is inactive, unmerge it so its manager
+                    // doesn't believe it still manages it.
+                    cell.unmerge();
+                }
+            }
+            // Remove the row.
+            row.remove();
+
+            // The place where the range used to live was just demolished. Give
+            // it shelter within the next active cell.
+            const nextActiveCell = nextCell.managerCell || nextCell;
+            range.setStart(nextActiveCell.firstLeaf(), RelativePosition.BEFORE);
+            range.collapse();
+        } else {
+            // If there is no `nextCell`, we're trying to delete the only row in
+            // this table so just remove the table.
+            this.deleteTable(params);
+        }
+    }
+    /**
+     * Delete the column at given range start.
+     *
+     * @param params
+     */
+    deleteColumn(params: CommandParams): void {
+        const range = params.context.range;
+        const cell = range.start.ancestor(TableCellNode);
+        if (!cell) return;
+
+        const column = cell.column;
+        const nextCell = cell.nextSibling(TableCellNode) || cell.previousSibling(TableCellNode);
+
+        if (nextCell) {
+            const nextColumnIndex = nextCell.columnIndex;
+
+            // Handle colspans and cell removal.
+            for (const cell of column) {
+                if (cell.colspan > 1) {
+                    // Cells managed by this cell will now be managed by the
+                    // cell after (or before if there is none after) instead.
+                    const afterCell = Array.from(cell.managedCells).find(
+                        managedCell => managedCell.columnIndex === nextColumnIndex,
+                    );
+                    if (afterCell) {
+                        afterCell.unmerge();
+                        for (const managedCell of cell.managedCells) {
+                            if (managedCell !== afterCell) {
+                                managedCell.mergeWith(afterCell);
+                            }
+                        }
+                    }
+                } else if (!cell.isActive()) {
+                    // If this cell is inactive, unmerge it so its manager
+                    // doesn't believe it still manages it.
+                    cell.unmerge();
+                }
+                // Remove the cell.
+                cell.remove();
+
+                // The place where the range used to live was just demolished.
+                // Give it shelter within the next active cell.
+                const nextManagerCell = nextCell.managerCell || nextCell;
+                range.setStart(nextManagerCell.firstLeaf(), RelativePosition.BEFORE);
+                range.collapse();
+            }
+        } else {
+            // If there is no `nextCell`, we're trying to delete the only column
+            // in this table so just remove the table.
+            this.deleteTable(params);
+        }
+    }
+    /**
+     * Delete the table at given range start.
+     *
+     * @param params
+     */
+    deleteTable(params: CommandParams): void {
+        const range = params.context.range;
+        const table = range.start.ancestor(TableNode);
+        if (!table) return;
+
+        const nextSibling = table.nextSibling();
+        const previousSibling = table.previousSibling();
+        if (nextSibling) {
+            range.setStart(nextSibling.firstLeaf(), RelativePosition.BEFORE);
+            range.collapse();
+        } else if (previousSibling) {
+            range.setStart(previousSibling.lastLeaf(), RelativePosition.AFTER);
+            range.collapse();
+        }
+        table.remove();
+    }
+    /**
+     * Merge the cells at given range into the first cell at given range.
+     *
+     * @param params
+     */
+    mergeCells(params: CommandParams): void {
+        const range = params.context.range;
+        const cells = range.targetedNodes(TableCellNode);
+        if (this._isRectangle(cells)) {
+            // Only merge the cells if they would not imply to merge
+            // unrelated cells, ie. the selected cells form a rectangle.
+            const managerCell = cells.shift();
+            for (const cell of cells) {
+                cell.mergeWith(managerCell);
+            }
+        }
+    }
+    /**
+     * Unmerge previously merged cells at given range.
+     *
+     * @param params
+     */
+    unmergeCells(params: CommandParams): void {
+        const range = params.context.range;
+        const cells = range.targetedNodes(TableCellNode);
+        for (const cell of cells) {
+            for (const managedCell of cell.managedCells) {
+                managedCell.unmerge();
+            }
+        }
+    }
+
+    //--------------------------------------------------------------------------
+    // Private
+    //--------------------------------------------------------------------------
+
+    /**
+     * Return true if the given array of cells forms a rectangle in the table
+     * grid.
+     *
+     * @param cells
+     */
+    _isRectangle(cells: TableCellNode[]): boolean {
+        cells = [...cells];
+
+        // Add managed cells to the list.
+        for (const cell of [...cells]) {
+            if (cell.managedCells.size) {
+                cells.push(...cell.managedCells);
+            }
+        }
+        cells = distinct(cells);
+
+        // Compute the row/column index extrema.
+        const rowIndices = cells.map(cell => cell.rowIndex);
+        const columnIndices = cells.map(cell => cell.columnIndex);
+        const minRowIndex = Math.min(...rowIndices);
+        const minColumnIndex = Math.min(...columnIndices);
+        const maxRowIndex = Math.max(...rowIndices);
+        const maxColumnIndex = Math.max(...columnIndices);
+
+        // If a cell between the extrema cannot be found in the list, the
+        // selected cells do not form a rectangle.
+        for (let rowIndex = minRowIndex; rowIndex <= maxRowIndex; rowIndex++) {
+            for (let columnIndex = minColumnIndex; columnIndex <= maxColumnIndex; columnIndex++) {
+                const cell = cells.find(cell => {
+                    return cell.rowIndex === rowIndex && cell.columnIndex === columnIndex;
+                });
+                if (!cell) {
+                    return false;
+                }
+            }
+        }
+
+        return true;
+    }
 }
--- a/packages/plugin-table/src/Table.ts
+++ b/packages/plugin-table/src/Table.ts
@@ -2,9 +2,16 @@ import { JWPlugin, JWPluginConfig } from '../../core/src/JWPlugin';
 import { TableDomParser } from './TableDomParser';
 import { Loadables } from '../../core/src/JWEditor';
 import { Parser } from '../../plugin-parser/src/Parser';
+import { TableRowDomParser } from './TableRowDomParser';
+import { TableCellDomParser } from './TableCellDomParser';
+import { TableDomRenderer } from './TableDomRenderer';
+import { TableRowDomRenderer } from './TableRowDomRenderer';
+import { TableCellDomRenderer } from './TableCellDomRenderer';
+import { Renderer } from '../../plugin-renderer/src/Renderer';
 
 export class Table<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T> {
-    readonly loadables: Loadables<Parser> = {
-        parsers: [TableDomParser],
+    readonly loadables: Loadables<Parser & Renderer> = {
+        parsers: [TableDomParser, TableRowDomParser, TableCellDomParser],
+        renderers: [TableDomRenderer, TableRowDomRenderer, TableCellDomRenderer],
     };
 }
--- a/packages/plugin-table/src/TableCellDomParser.ts
+++ b/packages/plugin-table/src/TableCellDomParser.ts
@@ -1,0 +1,26 @@
+import { AbstractParser } from '../../plugin-parser/src/AbstractParser';
+import { DomParsingEngine } from '../../plugin-dom/src/DomParsingEngine';
+import { TableCellNode } from './TableCellNode';
+
+export class TableCellDomParser extends AbstractParser<Node> {
+    static id = 'dom';
+    engine: DomParsingEngine;
+
+    predicate = (item: Node): item is HTMLTableCellElement => {
+        return item.nodeName === 'TD' || item.nodeName === 'TH';
+    };
+
+    /**
+     * Parse a table cell node.
+     *
+     * @param item
+     */
+    async parse(item: HTMLTableCellElement): Promise<TableCellNode[]> {
+        const cell = new TableCellNode(item.nodeName === 'TH');
+        cell.attributes = this.engine.parseAttributes(item);
+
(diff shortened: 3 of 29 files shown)


packages/plugin-table/src/TableCellNode.ts
```diff
@@ -0,0 +1,170 @@
...
+
+            // Hand the managed cells over to the manager.
+            for (const managedCell of this.managedCells) {
+                managedCell.managerCell = newManager;
+                this._managedCells.delete(managedCell);
+            }
+        } else if (oldManager) {
+            // If we just removed this cell's manager, also remove this cell
+            // from the old manager's managed cells.
+            oldManager._managedCells.delete(this);
```
dmo-odoo: Now that I think about it, how would the system react if I set a manager on non-contiguous cells ?

packages/plugin-table/src/TableCellNode.ts
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+    /**
+     * Return the set of cells that this cell managed.
```
dmo-odoo: Shouldn't it be "manage**s**" ?

packages/plugin-table/src/TableCellNode.ts
```diff
@@ -0,0 +1,170 @@
...
+    }
+    /**
+     * Return the computed column span of this cell, in function of its managed
+     * cells.
+     */
+    get colspan(): number {
+        let colspan = 1;
+        for (const managedCell of this.managedCells) {
+            colspan = Math.max(colspan, managedCell.columnIndex - this.columnIndex + 1);
+        }
```
dmo-odoo: Wait, what ? Why isn't it `1 + this.managedCells.length` ?

packages/plugin-table/src/TableCellNode.ts
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * Return the index of the row to which this cell belongs.
+     */
+    get rowIndex(): number {
+        return this.row.rowIndex;
+    }
+    /**
+     * Return the index of the column to which this cell belongs.
+     */
+    get columnIndex(): number {
```
dmo-odoo: You sure do love getters, don't you ?

packages/plugin-table/src/TableDomParser.ts
```diff
@@ -1,20 +1,129 @@
...
+import { TableCellNode } from './TableCellNode';
 
 export class TableDomParser extends AbstractParser<Node> {
     static id = 'dom';
     engine: DomParsingEngine;
 
-    predicate = (item: Node): boolean => {
-        return item instanceof Element && item.tagName === 'TABLE';
+    predicate = (item: Node): item is HTMLTableElement => {
+        return item.nodeName === 'TABLE';
```
dmo-odoo: Why did you change this line ? The only problem it had was that `Element` should have been changed to `HTMLTableElement`.

packages/plugin-table/src/TableDomParser.ts
```diff
@@ -1,20 +1,129 @@
...
         table.attributes = this.engine.parseAttributes(item);
-        const nodes = await this.engine.parse(...item.childNodes);
-        table.append(...nodes);
+
+        // Parse the contents of the table.
+        const children = await this.engine.parse(...item.childNodes);
+
+        // Build the grid.
+        const dimensions = this._getTableDimensions(item);
+        const parsedRows = children.filter(row => row.is(TableRowNode)) as TableRowNode[];
```
dmo-odoo: What happens to the nodes that are not `TR` ? And compared to the DOM ?

packages/plugin-table/src/TableDomParser.ts
```diff
@@ -1,20 +1,129 @@
...
+            const cells = grid[rowIndex];
+            let row = rows[rowIndex];
+            if (!row) {
+                row = new TableRowNode();
+            }
+            row.append(...cells);
+        }
+
+        // Append the rows to the table.
+        table.append(...rows);
```
dmo-odoo: It's a bit sad that we lose the grid from this point onward.

packages/plugin-table/src/TableDomParser.ts
```diff
@@ -1,20 +1,129 @@
...
+    /**
+     * Return a tuple with the row length and the column length of the given DOM
+     * table element.
+     *
+     * @param domTable
+     */
+    _getTableDimensions(domTable: HTMLTableElement): [number, number] {
+        const domRows: HTMLTableRowElement[] = [];
+        for (const domChild of domTable.querySelectorAll('tr')) {
+            domRows.push(domChild);
```
dmo-odoo: Can't the result of `querySelectorAll` be transformed into an array rather than pushing it manually ?

packages/plugin-table/src/TableDomParser.ts
```diff
@@ -1,20 +1,129 @@
...
+                while (grid[rowIndex][columnIndex]) {
+                    columnIndex += 1;
+                }
+
+                // Check traversing colspan and rowspan to insert placeholder
+                // cells where necessary. Consume these technical attributes.
+                const cellAttributes = cell.attributes['table-cell-attributes'];
+                delete cell.attributes['table-cell-attributes'];
+                const colspan = parseInt(cellAttributes.parsedColspan);
+                const rowspan = parseInt(cellAttributes.parsedRowspan);
```
dmo-odoo: Can't this function throw an error ?

packages/plugin-table/src/TableDomRenderer.ts
```diff
@@ -0,0 +1,58 @@
...
+        const renderedChildren = await this.renderChildren(table);
+        for (const renderedChild of renderedChildren) {
+            for (const domChild of renderedChild) {
+                // `TableRowDomRenderer` returns a tbody or thead with a tr in
+                // it. Append the tr to `domHead` or `domBody` in function of
+                // what the children are. If we didn't receive a thead or a
+                // tbody, then the child was not parsed by `TableRowDomRenderer`
+                // (probably a technical node, like a script or style) and we
+                // need to simply append the child to `domTable`.
+                if (this._isTableSection(domChild)) {
```
dmo-odoo: This function is duplicated in the parser and renderer. I would have defined a `const tableSectionTagNames` or something like that in the Plugin itself and imported it here to use `includes` in the `if` rather than create a functon just for that.

packages/plugin-table/src/TableDomRenderer.ts
```diff
@@ -0,0 +1,58 @@
...
+                    const section = domChild.nodeName === 'THEAD' ? domHead : domBody;
```
dmo-odoo: Especially if it's to re-test `THEAD` here...

packages/plugin-table/src/TableDomRenderer.ts
```diff
@@ -0,0 +1,58 @@
...
+                    // Transfer the attributes for the table section.
+                    for (const attribute of domChild.attributes) {
+                        section.setAttribute(attribute.name, attribute.value);
+                    }
```
dmo-odoo: Wait what ?

packages/plugin-table/src/TableNode.ts
```diff
@@ -1,7 +1,81 @@
...
+     */
+    get name(): string {
+        return super.name + ': ' + this.rowCount + 'x' + this.columnCount;
+    }
+    /**
+     * Return an array of row nodes in this table.
+     */
+    get rowNodes(): TableRowNode[] {
+        return this.children(TableRowNode);
+    }
```
dmo-odoo: I would have named this one `rows`.

packages/plugin-table/src/TableNode.ts
```diff
@@ -1,7 +1,81 @@
...
+    /**
+     * Return an array of rows in this table, as arrays of cells.
+     */
+    get rows(): TableCellNode[][] {
```
dmo-odoo: I would have named this one `grid` or `cells`.

packages/plugin-table/src/TableNode.ts
```diff
@@ -1,7 +1,81 @@
...
+        const rows = [];
+        for (const row of this.rowNodes) {
+            const rowGrid = [];
+            for (let columnIndex = 0; columnIndex < this.columnCount; columnIndex += 1) {
+                rowGrid.push(this.rowNodes[row.rowIndex].cells[columnIndex]);
+            }
```
dmo-odoo: Why something like `rowGrid.push([...this.rowNodes[row.rowIndex].cells])` wouldn't work ?

packages/plugin-table/src/TableNode.ts
```diff
@@ -1,7 +1,81 @@
...
+    }
+    /**
+     * Return an array of columns in this table, as arrays of cells.
+     */
+    get columns(): TableCellNode[][] {
+        const columns = [];
+        for (let columnIndex = 0; columnIndex < this.columnCount; columnIndex += 1) {
+            const columnGrid = [];
+            for (let rowIndex = 0; rowIndex < this.rowCount; rowIndex += 1) {
+                columnGrid.push(this.rowNodes[rowIndex].cells[columnIndex]);
```
dmo-odoo: Same here.

packages/plugin-table/src/TableRowDomParser.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Parse a row node or a table section node.
+     *
+     * @param item
+     */
+    async parse(item: Element): Promise<TableRowNode[]> {
+        if (this._isTableSection(item)) {
+            return this.parseTableSection(item);
+        } else if (item.nodeName === 'TR') {
```
dmo-odoo: It's funny because you're re-playing the predicate but separately this time... ^^'

packages/plugin-table/src/TableRowDomRenderer.ts
```diff
@@ -0,0 +1,30 @@
...
+        this.engine.renderAttributes(
+            row.attributes['table-section-attributes'] as Record<string, string>,
+            tableSection,
+        );
+        const renderedChildren = await this.renderChildren(row);
+        for (const renderedChild of renderedChildren) {
+            for (const domChild of renderedChild) {
+                domRow.appendChild(domChild);
+            }
+        }
```
dmo-odoo: I'm already tired of this `for/for` construction. Nothing related to your branch, though.

packages/plugin-table/src/Table.ts
```diff
@@ -5,8 +5,136 @@ import { TableCellDomParser } from './TableCellDomParser';
...
+        },
+        {
+            pattern: 'CTRL+<Period>',
+            commandId: 'mergeCells',
+        },
+        {
+            pattern: 'CTRL+SHIFT+<Period>',
+            commandId: 'unmergeCells',
+        },
+    ];
```
dmo-odoo: Do we do something for these shortcuts ?

packages/plugin-table/src/Table.ts
```diff
@@ -5,8 +5,136 @@ import { TableCellDomParser } from './TableCellDomParser';
...
+     *
+     * @param params
+     */
+    mergeCells(params: CommandParams): void {
+        const range = params.context.range;
+        const cells = range.targetedNodes(TableCellNode);
+        const managerCell = cells.shift();
+        for (const cell of cells) {
+            cell.managerCell = managerCell;
+        }
```
dmo-odoo: I think a proper function would be more obvious, something like `setManagerCell` ? Or maybe override `mergeWith` to have a different behavior if the given node is a table cell ?

packages/plugin-table/src/TableNode.ts
```diff
@@ -78,4 +78,130 @@ export class TableNode extends VElement {
...
+        referenceRow.before(newRow);
+        for (const cell of referenceRow.cells) {
+            const clone = cell.clone();
+            newRow.append(clone);
+
+            // Handle managers.
+            if (cell.managerCell) {
+                const manager = cell.managerCell;
+                if (manager.rowIndex === rowIndex + 1) {
+                    clone.managerCell = this.getCellAt(manager.rowIndex - 1, manager.columnIndex);
```
dmo-odoo: ?

packages/plugin-table/src/TableNode.ts
```diff
@@ -78,4 +78,130 @@ export class TableNode extends VElement {
...
+                } else if (managerEndRow !== rowIndex) {
+                    // Take the manager cell of the above cell (extend rowspan),
+                    // only if said manager's rowspan is not ending with the
+                    // above cell.
+                    clone.managerCell = manager;
+                }
+            } else if (cell.rowspan > 1) {
+                // If the cell has a rowspan, extend it.
+                clone.managerCell = cell;
+            }
```
dmo-odoo: I think I don't get these `if/else if/else`...

Zinston: OK 
image

Zinston:
@dmo-odoo It's ready to be re-reviewed. I kept the most disruptive changes in separate commits appended to the PR to ease the review/adapt loop but of course the idea is to rearrange the commits later.

packages/plugin-table/src/TableCellDomRenderer.ts
```diff
@@ -0,0 +1,51 @@
...
+            for (const domChild of renderedChild) {
+                td.append(domChild);
+            }
+        }
+
+        // If a cell is empty, fill it to make it visible.
+        if (!cell.hasChildren()) {
+            const placeholderBr = document.createElement('BR');
+            td.append(placeholderBr);
+        }
```
dmo-odoo: I would have done:
```typescript
if (cell.hasChildren()) {
    // the code at "Render the cell and its contents".
} else {
    // the br stuff
}
```

packages/plugin-table/src/TableCellNode.ts
```diff
@@ -0,0 +1,170 @@
...
+            coordinatesRepr += ':(' + endRow + ', ' + endColumn + ')';
+        }
+        coordinatesRepr += '>';
+        return (
+            super.name +
+            coordinatesRepr +
+            (this.header ? ': header' : '') +
+            (this.isActive ? '' : ' (inactive)')
+        );
+    }
```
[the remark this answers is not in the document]
dmo-odoo: This looks very complicated.

packages/plugin-table/src/TableCellNode.ts
```diff
@@ -0,0 +1,170 @@
...
+    set managerCell(newManager: TableCellNode | null) {
+        const oldManager = this._managerCell;
+        this._managerCell = newManager;
+
+        if (newManager) {
+            // If we just set a new manager for this cell, also add this cell to
+            // the new manager's managed cells.
+            newManager._managedCells.add(this);
+
+            // Copy the manager.
```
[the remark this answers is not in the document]
dmo-odoo: Well, you're not copying the manager as in "duplicating" him. You're copying the manager *attributes*.

packages/plugin-table/src/TableCellNode.ts
```diff
@@ -0,0 +1,159 @@
...
+
+            // Hand the managed cells over to the manager.
+            for (const managedCell of this.managedCells) {
+                managedCell.managerCell = newManager;
+                this.__managedCells.delete(managedCell);
+            }
+        } else if (oldManager) {
+            // If we just removed this cell's manager, also remove this cell
+            // from the old manager's managed cells.
+            oldManager.__managedCells.delete(this);
```
dmo-odoo: Why is this in a `else if` ? I mean, you had an `oldManager`, I'm assignining you a `newManager`, don't you have to update `oldManager` anyway ?

packages/plugin-table/src/TableCellNode.ts
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+    /**
+     * Return the set of cells that this cell manages.
+     */
+    get managedCells(): Set<TableCellNode> {
+        return new Set(this.__managedCells);
```
dmo-odoo: Because of this getter and the hiding of the property, you are accessing the private property from the outside here: https://github.com/odoo-dev/jabberwock/pull/106/commits/c50eafebc0093984ccd685931b70485e0ddda044#diff-8696de8087e2f8c00c3b0a80ad497247R74

packages/plugin-table/src/TableCellNode.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    //--------------------------------------------------------------------------
+    // Public
+    //--------------------------------------------------------------------------
+
+    /**
+     * Return true if this cell is active (ie not managed by another cell).
+     */
+    isActive(): boolean {
+        return !this.managerCell;
```
dmo-odoo: Wait, if no cell is merged, none of them are `active` ?

packages/plugin-table/src/TableNode.ts
dmo-odoo: I just realized you did not force push the branch, did you ?

packages/plugin-table/src/TableRowDomParser.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export class TableRowDomParser extends AbstractParser<Node> {
+    static id = 'dom';
+    engine: DomParsingEngine;
+
+    predicate = (item: Node): boolean => {
+        return (
+            item instanceof Element &&
+            (item.nodeName === 'THEAD' || item.nodeName === 'TBODY' || item.nodeName === 'TR')
+        );
```
dmo-odoo: Now that we were able to check that together, the predicate return type could be `item is Element` and remove `item instanceof Element` in the function so that the double parenthesis can be removed.

packages/plugin-table/src/TableRowDomParser.ts
dmo-odoo: What I meant is that it could be split into two different parsers, but honestly I'm not sure I would like that. Just something I noticed.

packages/plugin-table/src/Table.ts
dmo-odoo: The toolbar *is* merged.

packages/plugin-table/src/Table.ts
```diff
@@ -139,8 +144,51 @@ export class Table<T extends JWPluginConfig> extends JWPlugin<T> {
...
+     * Return true if the given array of cells forms a rectangle in the table
+     * grid.
+     *
+     * @param cells
+     */
+    _isRectangle(cells: TableCellNode[]): boolean {
+        // Add managed cells to the list.
+        for (const cell of cells.slice()) {
+            if (cell.managedCells.size) {
+                cells.push(...cell.managedCells);
```
dmo-odoo: Wait, you're modifying the array you were given. Rather than do `.slice` in the `for`, you can do `cells = [...cells]`

packages/plugin-table/src/Table.ts
```diff
@@ -139,8 +144,51 @@ export class Table<T extends JWPluginConfig> extends JWPlugin<T> {
...
+        // Compute the row/column index extrema.
+        const rowIndices = cells.map(cell => cell.rowIndex);
+        const columnIndices = cells.map(cell => cell.columnIndex);
+        const minRowIndex = Math.min(...rowIndices);
+        const minColumnIndex = Math.min(...columnIndices);
+        const maxRowIndex = Math.max(...rowIndices);
+        const maxColumnIndex = Math.max(...columnIndices);
+
+        // If a cell between the extrema cannot be found in the list, the
+        // selected cells do not form a quadrilateral.
```
dmo-odoo: quadrilateral

packages/plugin-table/src/TableCellNode.ts
```diff
@@ -150,4 +120,58 @@ export class TableCellNode extends VNode {
...
+            managedCell.mergeWith(newManager);
+            this.__managedCells.delete(managedCell);
+        }
+
+        // Copy the manager's row if an entire row was merged
+        const row = this.ancestor(TableRowNode);
+        if (row) {
+            const areAllManagedByManager = row
+                .children(TableCellNode)
+                .every(cell => cell.managerCell === newManager);
```
dmo-odoo: Why not:
```typescript
const cells = row.children(TableCellNode);
const rowIsMerged = cells.every(cell => cell.managerCell === newManager);
```

packages/plugin-table/src/TableCellNode.ts
```diff
@@ -150,4 +120,58 @@ export class TableCellNode extends VNode {
...
+            if (newManager.rowIndex !== this.rowIndex && areAllManagedByManager) {
```
dmo-odoo: I don't understand the `rowIndex` check in this condition. How can `areAllManagedByManager` be true and the `rowIndex` condition not be true ?

packages/plugin-table/src/TableCellNode.ts
```diff
@@ -150,4 +120,58 @@ export class TableCellNode extends VNode {
...
+                const managerRow = this.managerCell.ancestor(TableRowNode);
+                row.header = managerRow.header;
+                row.attributes = { ...managerRow.attributes };
+            }
+        }
+    }
+    unmerge(): void {
```
dmo-odoo: doc

packages/plugin-table/src/TableCellNode.ts
```diff
@@ -150,4 +120,58 @@ export class TableCellNode extends VNode {
...
+        if (this.__managerCell) {
+            // If we just removed this cell's manager, also remove this cell
+            // from the old manager's managed cells.
+            this.__managerCell.__managedCells.delete(this);
```
dmo-odoo: Accessing private properties here as well.

packages/plugin-table/test/Table.test.ts
```diff
@@ -3667,22 +3667,203 @@ describePlugin(Table, async testEditor => {
     describe('mergeCells', async () => {
         it('should merge two adjacent cells together', async () => {
             await testEditor(BasicEditor, {
-                contentBefore: '<table><tbody><tr><td>a[b</td><td>c]d</td></tr></tbody></table>',
+                /* eslint-disable prettier/prettier */
```
dmo-odoo: boooooh

packages/plugin-table/src/Table.ts
```diff
@@ -117,6 +140,132 @@ export class Table<T extends JWPluginConfig> extends JWPlugin<T> {
...
+     */
+    deleteRow(params: CommandParams): void {
+        const range = params.context.range;
+        const cell = range.start.ancestor(TableCellNode);
+        if (!cell) return;
+
+        const row = cell.ancestor(TableRowNode);
+        const nextRow = row.nextSibling(TableRowNode) || row.previousSibling(TableRowNode);
+        const nextCell = nextRow && nextRow.children(TableCellNode)[cell.columnIndex];
+        if (!nextCell) {
```
dmo-odoo: Please switch these conditions to avoid negative conditions.

packages/plugin-table/src/Table.ts
```diff
@@ -117,6 +140,132 @@ export class Table<T extends JWPluginConfig> extends JWPlugin<T> {
...
+     */
+    deleteColumn(params: CommandParams): void {
+        const range = params.context.range;
+        const cell = range.start.ancestor(TableCellNode);
+        if (!cell) return;
+
+        const column = cell.column;
+        const nextCell = cell.nextSibling(TableCellNode) || cell.previousSibling(TableCellNode);
+
+        if (!nextCell) {
```
dmo-odoo: Same here obviously.

packages/plugin-table/src/Table.ts
```diff
@@ -117,6 +140,132 @@ export class Table<T extends JWPluginConfig> extends JWPlugin<T> {
...
+                } else if (!cell.isActive()) {
+                    // If this cell is inactive, unmerge it so its manager
+                    // doesn't believe it still manages it.
+                    cell.unmerge();
+                }
+                // Remove the cell.
+                cell.remove();
+
+                // The place where the range used to live was just demolished.
+                // Give it shelter within the next active cell.
```
dmo-odoo: funny comment 👍 🍊

packages/plugin-table/src/TableCellNode.ts
Zinston: Same comment as above. Those private properties are built to be handled locally by merge/unmerge.

packages/plugin-table/test/Table.test.ts
Zinston: What? You prefer a gross and intrusive regular expression that removes the space? Or the whole html inline perhaps? :trollface:

packages/plugin-table/src/Table.ts
Zinston: :sunglasses:

karma.conf.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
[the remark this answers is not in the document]
dmo-odoo: Manifestement, oui... ;-) cc @Goaman

packages/plugin-table/src/TableCellDomParser.ts
```diff
@@ -0,0 +1,34 @@
+import { AbstractParser } from '../../core/src/AbstractParser';
+import { DomParsingEngine } from '../../plugin-dom/src/DomParsingEngine';
+import { TableCellNode } from './TableCellNode';
+
+export class TableCellDomParser extends AbstractParser<Node> {
+    static id = 'dom';
+    engine: DomParsingEngine;
+
+    predicate = (item: Node): item is HTMLTableCellElement => {
+        return item.nodeName === 'TD' || item.nodeName === 'TH';
```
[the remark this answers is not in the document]
dmo-odoo: Yes we checked it together, this was an old comment.

packages/plugin-table/src/TableCellNode.ts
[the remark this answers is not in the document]
dmo-odoo: Ok. However, the `this.isActive` bit cannot work, as `isActive` is a function and the property always exists. I suggest you re-test this function. :trollface:

packages/plugin-table/src/TableCellNode.ts
dmo-odoo: Well, if you do `<anything_other_than_"this">.__<private_property>` then yes, for me it is the outside. I'm wondering if it would be better or worse if instead of doing:
```typescript
managedCell.managerCell = manager;
```

We did:
```typescript
managerCell.manage(managedCell);
```

But I guess we're a bit screwed on that since we decided to use `mergeWith` as it would also need to write on the manager's private property from the outside...

packages/plugin-table/src/TableCellNode.ts
dmo-odoo: I guess I misread the code, my bad.

packages/plugin-table/src/TableCellNode.ts
dmo-odoo: But if we selected only part of a line, then `areAllManagedByManager` is going to be `false`, right ? So the entire condition will be `false` anyway without the need for the `rowIndex` condition, or did I miss something ?

packages/plugin-table/src/TableCellNode.ts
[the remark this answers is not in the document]
dmo-odoo: No, it is not the same. When you copy/paste on your computer, the original is duplicated. So when you say you "copy the manager", I expect that we end up with a copy *of the manager* somewhere, which is not the case. You never create a *copy* of the manager. It's possible it's mostly my understanding which is biased by computing and "copy" might actually mean what you're saying it means (you know english better than I do), but I still think writing "Copy the manager's attributes and properties." instead of "Copy the manager." would not hurt anything and would improve the understandability for people like me whose english is highly influenced by computing, which is likely to be the case of most people who might stumble upon this comment.

packages/plugin-table/src/Table.ts
dmo-odoo: Then I'd do `cells = [...cells]` above then `for (const cell of [...cells])` for the loop.

packages/plugin-table/test/Table.test.ts
dmo-odoo: Sincerely ? Yes I prefer the regex because at least the html in the test is easy to update. In your version, I have to manually manage the quotes and commas, while they do not even exist with the regex. With the regex, it's just like if you were in a proper html file, nothing else to worry about. You wrote a lot of tests though and it would be a waste of precious resources to have you update them (and to be fair, this regex is not even merged yet !) so we'll leave it at that for this PR.

packages/plugin-table/src/TableCellNode.ts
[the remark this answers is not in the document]
Zinston: Fine, no problem.

packages/plugin-table/src/TableCellNode.ts
Zinston: Exactly...

You:
> if you do <anything_other_than_"this">.__<private_property> then yes, for me it is the outside

Code:
> `this.__managedCells`

Problem:
> ...

packages/plugin-table/src/Table.ts
Zinston: ... which is what I did :-)

packages/plugin-table/src/TableCellNode.ts
Zinston: Ok so here `__managedCells` is not accessed by `this`. but I believe my above comment still apply though. Whaddayathink?

packages/plugin-table/test/Table.test.ts
Zinston: Ok. Agree to disagree ^^

packages/plugin-table/src/TableCellNode.ts
dmo-odoo: Re-read my comment ! 🚎  The problem is not here, it is there: c50eafe#diff-8696de8087e2f8c00c3b0a80ad497247R74
I put the comment here because it is the existence of this getter that is somewhat the *cause* of the issue... ;-)

dmo-odoo: I think we need to discuss these private properties by audio. I still need to work with christophe tomorrow but maybe we can squeeze like 10 minutes to discuss this.

Zinston:
@dmo-odoo Done and reorganized.

Zinston:
@dmo-odoo Ready.

dmo-odoo:
@Zinston Je peux te demander d'ajouter un warning sur la page demo au dessus de la table pour expliquer que les suppressions ne sont pas supportées pour l'instant parce qu'on n'a pas encore fait unbreakable ?

PR #109, opened 2020-02-24
Master ui chm

Gorash:


packages/utils/src/isBlock.ts
```diff
@@ -51,6 +51,15 @@ const computedStyles = new WeakMap<Node, CSSStyleDeclaration>();
...
     if (node instanceof Element) {
+        const tagName = node.tagName.toUpperCase();
+        // every custom jw-* node will be considered as blocks
+        if (tagName.startsWith('JW-') || tagName === 'T') {
+            return true;
+        }
+        // every custom td, tr node will be considered as blocks
+        if (tagName === 'TD' || tagName === 'TR') {
+            return true;
+        }
```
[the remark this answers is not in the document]
dmo-odoo: @Zinston Christophe tells me that formatting spaces are not correctly removed around TD and TR without this. Since you worked a lot on `isBlock` and its related issues, could you please enlighten me/us ? Thanks !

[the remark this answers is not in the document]
Zinston: I adapted `blockTagNames` to ensure that so normally it should work without this extra code... Maybe he just didn't try since tables have been merged?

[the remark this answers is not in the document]
dmo-odoo: I doubt this PR has been rebased on master since tables were merged, yes.

packages/owl-ui/src/OwlPlugin.ts
```diff
@@ -0,0 +1,25 @@
...
+    readonly loadables: Loadables<Renderer> = {
+        renderers: [OwlDomRenderer],
+    };
+
+    owlEnv: OwlUIEnv = {
+        qweb: new QWeb(),
+        editor: this.editor,
+    };
+
+    addTemplates(xmlstr: string | Document): void {
```
Goaman: To add use ressource from another plugin we use the loadable mechanism.

packages/plugin-toolbar/src/Toolbar.ts
```diff
@@ -21,18 +27,19 @@ export interface ToolbarConfig extends JWPluginConfig {
...
+                id: 'toolbar',
+                zone: 'tools',
+                VNode: OwlNode,
+                args: [ToolbarComponent as typeof OwlUIComponent, {}],
+            },
+        ],
+    };
+    constructor(editor: JWEditor, configuration: T) {
+        super(editor, configuration);
+        this.editor.plugins.get(Owl).addTemplates(toolbarTemplates);
```
Goaman: Use the loadable mechanism to add the templates.

packages/plugin-dom-editable/src/DomEditable.ts
```diff
@@ -0,0 +1,202 @@
...
+    readonly loadables: Loadables<Renderer & Layout> = {
+        renderers: [EditableDomRenderer],
+        layoutComponents: [],
+    };
+
+    commandHooks = {
+        '*': this._renderInEditable,
+    };
+
+    node: VElement;
```
Goaman: root?

packages/plugin-layout/src/Layout.ts
```diff
@@ -0,0 +1,111 @@
...
+    };
+    commands = {
+        'UI-show': {
+            title: 'Show a layout component',
+            handler: this.show.bind(this),
+        },
+        'UI-hide': {
+            title: 'Hide a layout component',
+            handler: this.hide.bind(this),
+        },
```
Goaman: Does the commands are intended to be used by a user from a "command bar" or are they intended to be use by the system with execCommand?

packages/plugin-layout/src/LayoutEngine.ts
```diff
@@ -0,0 +1,277 @@
...
+            }
+        }
+        return zones;
+    }
+    /**
+     *
+     * Show the components corresponding to given ref. Return the updated zones.
+     */
+    async show(ref: componentReference): Promise<VNode[]> {
+        ref = (ref instanceof VNode && this.references.get(ref)) || ref;
```
Goaman: Do not write on the attribute but create another const variable instead

packages/plugin-layout/src/Layout.ts
```diff
@@ -0,0 +1,111 @@
...
+    };
+
+    async start(): Promise<void> {
+        for (const layoutEngine of this.engines) {
```
Goaman: Promise.all?

packages/plugin-layout/src/Layout.ts
```diff
@@ -0,0 +1,111 @@
...
+            await layoutEngine[1].start();
+        }
+    }
+    async stop(): Promise<void> {
+        for (const layoutEngine of this.engines) {
```
Goaman: Promise.all?

packages/plugin-layout/src/Layout.ts
```diff
@@ -0,0 +1,111 @@
...
+        const promises = [];
+        for (const layoutEngine of this.engines) {
+            promises.push(layoutEngine[1].remove(ref));
+        }
+        await Promise.all(promises);
+    }
+    // to show component (instance or clonse) inside the zone
+    async show(params: { ref: componentReference }): Promise<void>;
+    async show(ref: componentReference): Promise<void>;
+    async show(params: { ref: componentReference } | componentReference): Promise<void> {
```
Goaman: This syntactic sugar is not useful.
This is not a pattern that we use for defining command.

The pattern used for all command would look like this:
```typescript
async show(params: UIShowParams): Promise<void> {}
```

packages/plugin-dom-layout/src/DomLayout.ts
```diff
@@ -0,0 +1,109 @@
...
+import { Dom } from '../../plugin-dom/src/Dom';
+import { VSelectionDescription, VSelection } from '../../core/src/VSelection';
+import { DomSelectionDescription } from '../../plugin-dom-editable/src/EventNormalizer';
+import { ZoneDomRenderer } from './ui/ZoneDomRenderer';
+import { ZoneDomParser } from './ui/ZoneDomParser';
+import { VElement } from '../../core/src/VNodes/VElement';
+import { VNode } from '../../core/src/VNodes/VNode';
+import { ZoneIdentifier, ZoneNode } from '../../plugin-layout/src/ZoneNode';
+
+export interface DomLayoutConfig extends JWPluginConfig {
```
Goaman: I don't see the use of having 4 differents ways to configure the layout.

packages/plugin-layout/src/ZoneNode.ts
```diff
@@ -0,0 +1,36 @@
+import { VNode } from '../../core/src/VNodes/VNode';
+
+export type ZoneIdentifier = string;
+
+export class ZoneNode extends VNode {
+    hidden: Map<VNode, boolean> = new Map();
+
+    constructor(public managedZones: ZoneIdentifier[]) {
```
Goaman: Why one zone can manages multiples zone?
I could see one zone needing to manage multiples components but no other nodes.

packages/plugin-layout/src/ZoneNode.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+    show(child: VNode): void {
+        this.hidden.set(child, false);
+        const parentZone: ZoneNode = this.ancestor(ZoneNode);
+        if (parentZone) {
+            parentZone.show(this);
+        }
+    }
+    _insertAtIndex(child: VNode, index: number): void {
+        this.hidden.set(child, true);
```
Goaman: Strange behavior, why whould you change the visibility of a node when you move it?
It's like having div.display = 'block' when moving the div no?

packages/plugin-layout/src/LayoutEngine.ts
```diff
@@ -0,0 +1,277 @@
...
+     */
+    load(Component: C): void {
+        this.Components.push(Component);
+    }
+    async start(): Promise<void> {
+        this.templates = [];
+        const zoneNodes: ZoneNode[] = [];
+        for (const template of this.configuration.templates) {
+            this.templates.push(template);
+            zoneNodes.push(...template.node.descendants(ZoneNode));
```
Goaman: What if `template.node` is empty as the node could be not set as written in the interface?

packages/plugin-dom-editable/src/DomEditable.ts
```diff
@@ -0,0 +1,202 @@
...
+            },
+        );
+    }
+    /**
+     * KeyboardEvent listener to be added to the DOM that calls `execCommand` if
+     * the keys pressed match one of the shortcut registered in the keymap.
+     *
+     * @param event
+     */
+    private async _processKeydown(event: KeyboardEvent): Promise<CommandIdentifier> {
```
Goaman: The logic to handle keydown should be on the layout manager as shortcuts could be done on any part of the layout.
However, this is true that some of the logic of the current function should stay here (the EventNormalizer).

packages/owl-ui/src/OwlPlugin.ts
Gorash: :+1:

packages/plugin-dom-editable/src/DomEditable.ts
Gorash: It is not necessarily root, I think. Nothing prevents to choose another element.

packages/plugin-layout/src/Layout.ts
Gorash: by buttons, toolbar, close dialog...

packages/plugin-layout/src/Layout.ts
Gorash: There are two syntaxes, I put the two definitions. I don't see the use of your UIShowParams

packages/plugin-dom-layout/src/DomLayout.ts
Gorash: ?

packages/plugin-layout/src/ZoneNode.ts
Gorash: dialog who manage float, default for example.

packages/plugin-layout/src/ZoneNode.ts
Gorash: This is the area that knows what to do with the elements to hide (put them in an onlget (eg: a tabbed panel), not add them to the dom (ex: dialog), show them anyway so as not to have a empty area (ex: hand)

packages/plugin-layout/src/LayoutEngine.ts
Gorash: It is indeed impossible. there are always vNodes, even if the user gives a template because it is parsed.

packages/plugin-dom-editable/src/DomEditable.ts
Gorash: It's can be move on the DomLayout, and changed to use all interface.
But, when I tried to do, and it's strange to call the domLayout in '_onNormalizedEvent' method

packages/plugin-dialog/src/ui/DialogZoneNode.ts
```diff
@@ -0,0 +1,3 @@
+import { ZoneNode } from '../../../plugin-layout/src/ZoneNode';
+
+export class DialogZoneNode extends ZoneNode {}
```
Goaman: See the comment on ZoneNode.
I don't see why a DialogZone need to inherit from a ZoneNode. It could however have a flag that tells that it manages zones in order to receive "signals".

packages/plugin-layout/src/Layout.ts
Goaman: Yes but why do we need two syntax.
For syntactic sugar ? That only add an unnecessary code style.
Let's prefer consistency.

packages/plugin-layout/src/LayoutEngine.ts
```diff
@@ -0,0 +1,277 @@
...
+import { ZoneIdentifier, ZoneNode } from './ZoneNode';
+import { FragmentNode } from '../../core/src/VNodes/FragmentNode';
+
+export type componentReference = string | VNode;
+export type DomZonePosition = 'after' | 'prepend' | 'append' | 'replace';
+
+export interface Component<N extends Constructor<VNode> = Constructor<VNode>> {
+    zone: ZoneIdentifier;
+    VNode?: N | InstanceType<N>;
+    args?: ConstructorParameters<N>;
```
Goaman: The only thing that we need when defining a component is a way to get it's VNode. 
These two properties generate clutter in the code because we have to make branching to handle them.
This could simply be avoided by providing a function that retrieve the VNode describing the Component
```typescript
getVnode() => VNode
```

packages/utils/src/isBlock.ts
```diff
@@ -51,6 +51,15 @@ const computedStyles = new WeakMap<Node, CSSStyleDeclaration>();
 export function isBlock(node: Node): boolean {
     let result: boolean;
     if (node instanceof Element) {
+        const tagName = node.tagName.toUpperCase();
+        // every custom jw-* node will be considered as blocks
+        if (tagName.startsWith('JW-') || tagName === 'T') {
+            return true;
+        }
```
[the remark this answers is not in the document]
Gorash: In the template, to parse and remove spaces

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -719,32 +742,10 @@ describe('Plugin: DevTools', () => {
...
-                'mergeCells',
-                'unmergeCells',
-                'align',
-                'applyPreStyle',
-                'colorText',
-                'uncolorText',
-                'colorBackground',
-                'uncolorBackground',
+                'UI-show',
+                'UI-hide',
```
dmo-odoo, line 749: I would rename these commands to `show` and `hide`.

packages/plugin-keymap/test/Keymap.test.ts
```diff
@@ -26,7 +29,7 @@ describe('Keymap', () => {
             expect(keymap.mappings[LEVEL.DEFAULT].length === 1);
         });
     });
-    describe('parsePattern', () => {
+    describe('parsePattern', async () => {
```
dmo-odoo: I thought you realized we didn't need the `async` describe at the end ?

packages/plugin-linebreak/test/LineBreakDomRenderer.test.ts
```diff
@@ -11,7 +10,7 @@ import { nodeName } from '../../utils/src/utils';
 describe('LineBreakDomRenderer', () => {
     describe('render', () => {
         let editor: JWEditor;
-        let root: FragmentNode;
+        let root: VElement;
```
dmo-odoo, line 13: I don't see why root couldn't be a `VNode`.

packages/utils/src/testUtils.ts
```diff
@@ -80,7 +81,14 @@ export async function testEditor(
...
     const editor = initSpec(Editor, spec, container);
-    await testSpec(editor, spec);
+    try {
+        await testSpec(editor, spec);
+    } catch (e) {
+        if (!spec.debug) {
+            container.remove();
+        }
+        throw e;
+    }
```
dmo-odoo, line 91: 👍

packages/utils/src/testUtils.ts
```diff
@@ -182,13 +182,16 @@ async function testSpec(editor: JWEditor, spec: TestEditorSpec): Promise<void> {
...
         const domPlugin = editor.plugins.get(DomEditable);
-        expect(domPlugin.editable.innerHTML).to.deep.equal(spec.contentAfter);
-    }
-    if (!spec.debug) {
+        const value = domPlugin.editable.innerHTML;
+        if (!spec.debug) {
+            await editor.stop();
+        }
+        expect(value).to.equal(spec.contentAfter);
+    } else if (!spec.debug) {
```
dmo-odoo, line 194: I don't understand this change.

packages/core/src/Core.ts
```diff
@@ -78,19 +78,25 @@ export class Core<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
             const previousSibling = range.start.previousSibling();
             if (previousSibling) {
                 previousSibling.removeBackward();
-            } else if (range.startContainer.breakable) {
+            } else if (range.startContainer.breakable && !range.startContainer.independent) {
```
dmo-odoo: I feel this is outside of the scope of this PR so I'm not reviewing this commit, especially considering it uses the `independent` concept that we are still unsure of. (Yesterday we talked about `editable`.)

packages/plugin-owl/src/ui/OwlDomRenderer.ts
```diff
@@ -1,18 +1,28 @@
...
 export class OwlDomRenderer extends AbstractRenderer<Node[]> {
     static id = 'dom';
     engine: DomRenderingEngine;
     predicate = OwlNode;
+    env: OwlUIEnv;
+
+    constructor(engine: DomRenderingEngine, superRenderer: Renderer<Node[]>) {
+        super(engine, superRenderer);
+        this.engine = engine;
+        this.super = superRenderer;
```
dmo-odoo, line 17: This is already done by the call to `super`.

packages/plugin-owl/src/ui/OwlDomRenderer.ts
```diff
@@ -1,18 +1,28 @@
...
+        const owlPlugin = this.engine.editor.plugins.get(Owl);
+        this.env = owlPlugin.owlEnv;
```
dmo-odoo, line 19: At this point why not do `this.engine.editor.plugins.get(Owl).env` ? (So we would also rename `owlEnv`  to `env` since it is already in the plugin `Owl` anyway.)

packages/plugin-blockquote/src/BlockquoteDomParser.ts
```diff
@@ -1,11 +1,11 @@
 import { AbstractParser } from '../../plugin-parser/src/AbstractParser';
-import { DomParsingEngine } from '../../plugin-dom/src/DomParsingEngine';
+import { XmlParsingEngine } from '../../plugin-xml/src/XmlParsingEngine';
 import { BlockquoteNode } from './BlockquoteNode';
 import { nodeName } from '../../utils/src/utils';
 
 export class BlockquoteDomParser extends AbstractParser<Node> {
```
dmo-odoo, line 6: All these `<...>Domparser` should be renamed `<...>XmlParser` if they are supposed to go into the `XmlParsingEngine`.

packages/plugin-blockquote/src/BlockquoteDomParser.ts
```diff
@@ -1,11 +1,11 @@
...
-    static id = 'dom';
-    engine: DomParsingEngine;
+    static id = 'xml';
+    engine: XmlParsingEngine;
 
     predicate = (item: Node): boolean => {
```
dmo-odoo, line 10: If they are supposed to parse `xml` then they should be an `AbstractParser<string>` I guess, and the predicate should take a `string` rather than a `Node`. Same with `parse`, if it's an "xml parser" then it should parse XML, that is `string`, not `Element`.

packages/plugin-dom/src/DomParsingEngine.ts
```diff
@@ -1,18 +1,9 @@
...
-        return Array.from(node.attributes || []).reduce((attributes, attribute) => {
-            attributes[attribute.name] = attribute.value;
-            return attributes;
-        }, {});
-    }
+export type DomNodeType = HTMLElement | CharacterData;
+
+export class DomParsingEngine<T extends DomNodeType = DomNodeType> extends XmlParsingEngine<T> {
+    static readonly id: ParsingIdentifier = 'dom';
+    static fallbacks = ['xml'];
```
dmo-odoo, line 8: I suggest to use the `T1` and `T2` notation we created yesterday. `T1` would be `Node` (for Dom) and `T2` would be `string` (for Xml). This way, the `DomParsingEngine` will be able to parse both Dom nodes and xml string, which I think is the objective. Then I suggest to rename `fallbacks` to `extends` since at this point they are not "fallbacks" per se anymore.

packages/plugin-parser/src/Parser.ts
```diff
@@ -22,10 +23,21 @@ export class Parser<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<
...
+            for (const id in this.engines) {
+                const parsingEngine = this.engines[id];
+                if (id === ParserClass.id) {
+                    parsingEngine.register(ParserClass);
+                } else {
+                    const fallbacks = parsingEngine.constructor.fallbacks;
+                    if (fallbacks && fallbacks.includes(ParserClass.id)) {
+                        parsingEngine.register(ParserClass);
+                    }
+                }
```
dmo-odoo, line 35: Instead of this `if/else/if` construction, I suggest this:
```typescript
const supportedTypes = [id, ...parsingEngine.constructor.extends];
if (supportedTypes.includes(ParserClass.id) {
    parsingEngine.register(ParserClass);
}
```

packages/plugin-parser/src/Parser.ts
```diff
@@ -22,10 +23,21 @@ export class Parser<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<
...
             }
         }
     }
+
+    async parse(engineId: string, ...items): Promise<VNode[]> {
```
dmo-odoo, line 40: Copy the doc of `render` of `Renderer` here and adapt it for `parse`.

packages/plugin-parser/src/ParsingEngine.ts
```diff
@@ -11,18 +11,21 @@ export interface Parser<T = {}> {
 
 export class ParsingEngine<T = {}> {
     static readonly id: ParsingIdentifier;
-    static defaultParser: Constructor<Parser>;
+    static readonly defaultParser: ParserConstructor;
+    static fallbacks?: ParsingIdentifier[];
```
dmo-odoo, line 15: I would put `fallbacks` (or `extends`) right after `id`. In addition, I would put it `readonly`.

packages/plugin-parser/src/ParsingEngine.ts
```diff
@@ -11,18 +11,21 @@ export interface Parser<T = {}> {
...
     readonly parsingMap = new Map<T, VNode[]>();
 
     constructor(editor: JWEditor) {
         this.editor = editor;
-        const defaultParser = new this.constructor.defaultParser(this);
-        if (defaultParser.predicate) {
-            throw `Default renderer cannot have a predicate.`;
-        } else {
-            this.parsers.push(defaultParser);
+        if (this.constructor.defaultParser) {
```
dmo-odoo, line 22: I don't understand this change. `ParsingEngines` *must* have a `defaultParser`, why the `if` ?

packages/plugin-parser/src/ParsingEngine.ts
```diff
@@ -35,8 +38,26 @@ export class ParsingEngine<T = {}> {
...
+                if (constructorA.id === constructorB.id) {
+                    return 0;
+                }
+                if (constructorA.id === id) {
+                    return -1;
+                }
+                if (constructorB.id === id) {
+                    return 1;
+                }
+                return fallbacks.indexOf(constructorA.id) - fallbacks.indexOf(constructorB.id);
```
dmo-odoo, line 58: You'll need to re-explain to me why we need this code.

packages/plugin-parser/src/ParsingEngine.ts
```diff
@@ -84,6 +105,7 @@ export type ParsingEngineConstructor<T = {}> = {
     new (...args: ConstructorParameters<typeof ParsingEngine>): ParsingEngine;
     id: ParsingIdentifier;
     defaultParser: ParserConstructor<T>;
+    fallbacks?: ParsingIdentifier[];
```
dmo-odoo, line 108: If we rename `fallbacks` to `extends` then this one will need to be renamed as well. And please move it alongside `id` and before `defaultParser`.

packages/plugin-parser/src/ParsingEngine.ts
dmo-odoo: I would also put back the `Constructor<Parser>` type.

dmo-odoo:
🎉

Zinston:
Wow! Good job guys!

PR #113, opened 2020-02-27
[IMP] mocha: do not timeout when debugging

Goaman:


karma.conf.js
```diff
@@ -58,9 +58,12 @@ module.exports = function(config) {
     }
 
     let port = 0;
+    let mochaTimeout = 2000;
     if (config.debug) {
         port = 9876;
         webpackConfig.devtool = 'inline-source-map';
+        // When debuging in the browser, do not timeout
+        mochaTimeout = 1000000;
```
dmo-odoo: Why using an arbitrary high value when using `0` would simply disable the timeout altogether ?

Goaman: I didn't know, you're right.

PR #130, opened 2020-03-18
[ADD] Align: align text left/center/right/justify

Zinston:
image

### DIFF
--- a/packages/plugin-align/src/Align.ts
+++ b/packages/plugin-align/src/Align.ts
@@ -1,0 +1,111 @@
+import JWEditor, { Loadables } from '../../core/src/JWEditor';
+import { JWPlugin, JWPluginConfig } from '../../core/src/JWPlugin';
+import { CommandParams } from '../../core/src/Dispatcher';
+import { Keymap } from '../../plugin-keymap/src/Keymap';
+import { VNode } from '../../core/src/VNodes/VNode';
+
+export enum AlignType {
+    LEFT = 'left',
+    CENTER = 'center',
+    RIGHT = 'right',
+    JUSTIFY = 'justify',
+}
+export type AlignParams = CommandParams & {
+    type: AlignType;
+};
+
+export class Align<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T> {
+    static dependencies = [];
+    editor: JWEditor;
+    commands = {
+        align: {
+            handler: this.align.bind(this),
+        },
+    };
+    loadables: Loadables<Keymap> = {
+        shortcuts: [
+            {
+                pattern: 'Ctrl+Shift+L',
+                commandId: 'align',
+                commandArgs: { type: AlignType.LEFT },
+            },
+            {
+                pattern: 'Ctrl+Shift+E',
+                commandId: 'align',
+                commandArgs: { type: AlignType.CENTER },
+            },
+            {
+                pattern: 'Ctrl+Shift+R',
+                commandId: 'align',
+                commandArgs: { type: AlignType.RIGHT },
+            },
+            {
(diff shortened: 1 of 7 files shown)


packages/plugin-align/src/Align.ts
```diff
@@ -0,0 +1,86 @@
...
+                } else {
+                    // Add an alignment to the existing styles.
+                    node.attributes.style = [textAlign, node.attributes.style].join('; ');
+                }
+            } else {
+                // Create a new style attribute, with an alignment.
+                node.attributes.style = textAlign;
+            }
+            // Remove unnecessary space from the style attribute.
+            node.attributes.style = node.attributes.style.replace(/^\s*|\s*$/g, '');
```
dmo-odoo: Unnecessary spaces ? Like what ?

packages/plugin-align/src/Align.ts
```diff
@@ -0,0 +1,86 @@
...
+            if (!node.attributes.style.length) {
+                // If the style attribute is empty, remove it.
+                delete node.attributes.style;
```
dmo-odoo: How can it be empty ? You just set text-align on it.

packages/plugin-align/src/Align.ts
```diff
@@ -0,0 +1,86 @@
...
+            } else if (!node.attributes.style.endsWith(';')) {
+                // Ensure the style attribute ends with a semicolon.
+                node.attributes.style += ';';
```
dmo-odoo: Just curious, why ?

packages/plugin-align/src/AlignButtons.ts
```diff
@@ -0,0 +1,55 @@
...
+import JWEditor from '../../core/src/JWEditor';
+import { VNode } from '../../core/src/VNodes/VNode';
+
+function isAligned(node: VNode, type: AlignType): boolean {
+    if (typeof node.attributes.style !== 'string') {
+        return false;
+    } else {
+        const styles = node.attributes.style.split(';');
+        const align = styles.find(style => style.includes('text-align'));
+        return align && align.includes(type);
```
dmo-odoo: Vivement les modifiers et le `CSSStyle` modifier... ;-)

packages/plugin-align/src/Align.ts
Zinston: Like 'color: white; ' or '  '. Don't know why they're there but they can be. Potentially from template.

packages/plugin-align/src/Align.ts
Zinston: Oops. Indeed. Originally I _removed_ the `text-align` if it was to be set to `left` because that's the default. But then I realized the node could have a parent that has a `text-align` so I would need to force it. Then I forgot to remove these three lines.

packages/plugin-align/src/Align.ts
Zinston: Standards... I'm used to seeing styles ending with ';' in html and it bothers me if it doesn't. Is that wrong doctor?

packages/plugin-align/src/AlignButtons.ts
Zinston: Ah carrément!

packages/plugin-align/src/Align.ts
```diff
@@ -0,0 +1,140 @@
...
+        }
+    }
+    /**
+     * Return the first ancestor that has an alignment style.
+     *
+     * @param node
+     */
+    static alignedAncestor(node: VNode): VNode {
+        return node.ancestor(ancestor => this.isAligned(ancestor));
+    }
```
dmo-odoo: Seriously ?

PR #136, opened 2020-03-19
[FIX] EventNormalizer: paste with context menu

Zinston:
Concerning the test, I see no reason why we should expect a `setSelection`. If @Goaman (who wrote the test) confirms then I believe we can merge as is.


dmo-odoo:
C'est rouge 🔴 chez moi :(

PR #137, opened 2020-03-19
[ADD] Plugins: Fontawesome

Zinston:


webpack.config.js
```diff
@@ -39,6 +39,10 @@ module.exports = {
                 test: /\.xml$/i,
                 use: ['text-loader'],
             },
+            {
+              test: /\.(eot|svg|ttf|woff|woff2)$/,
+              loader: 'file-loader?name=/fonts/[name].[ext]'
```
Zinston: This is actually unused because we force file-loader at every import.
@Goaman This is your area of expertise... If you find a better solution please be my guest ;-)

examples/demo/demo.ts
```diff
@@ -19,6 +19,16 @@ import { UnorderedListButton } from '../../packages/plugin-list/src/ListButtons'
...
+import '!style-loader!css-loader!@fortawesome/fontawesome-free/css/all.css';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-brands-400.woff';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-brands-400.woff2';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-brands-400.ttf';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-regular-400.woff';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-regular-400.woff2';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-regular-400.ttf';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-solid-900.woff';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-solid-900.woff2';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-solid-900.ttf';
```
Zinston: I tried moving this to `Fontawesome.ts` later and it works. Should I do it in a separate commit?

packages/plugin-fontawesome/src/FontawesomeNode.ts
```diff
@@ -1,8 +1,53 @@
...
+            // We are targetting one of the invisible characters surrounding the
+            // fontawesome node.
+            if (FontawesomeDomParser.isFontawesome(domNode.previousSibling)) {
+                // We are targetting the invisible character AFTER the
+                // fontawesome node.
+                if (offset === 0) {
+                    // Moving from after the fontawesome node to before it.
+                    // (DOM is `<invisible/><fontawesome/>[]<invisible/>` but
+                    // should be `[]<invisible/><fontawesome/><invisible/>`).
+                    return [this, RelativePosition.BEFORE];
```
dmo-odoo: We need to move this `locate` stuff. It is DOM stuff so it should be on the renderer. If it was ideed on the renderer, you could simply read the `engine.renderings` object rather than having to "observe" your DOM surroundings. Something like this:
```typescript
class FontAwesomeDomRenderer {
    locate(domNode: Node, offset: number, node: VNode): [this, RelativePosition] {
        const rendering = this.engine.renderings.get(node);
        if (domNode === rendering[rendering.length - 1]) {
            // We are targetting the invisible character AFTER the
                // fontawesome node.
            if (offset === 0) {
                // Moving from after the fontawesome node to before it.
                // (DOM is `<invisible/><fontawesome/>[]<invisible/>` but
                // should be `[]<invisible/><fontawesome/><invisible/>`).
                return [this, RelativePosition.BEFORE];
            } else {
                // Stay after the fontawesome node.
                return [this, RelativePosition.AFTER];
            }
        } else if (domNode === rendering[0]) {
            // We are targetting the invisible character BEFORE the
            // fontawesome node.
            if (offset === 1) {
                 // Moving from before the fontawesome node to after it.
                // (DOM is `<invisible/>[]<fontawesome/><invisible/>` but
                // should be `<invisible/><fontawesome/><invisible/>[]`).
                return [this, RelativePosition.AFTER];
            } else {
                // Stay before the fontawesome node.
                return [this, RelativePosition.BEFORE];
            }
        }  else {
            // We should never reach this. It's a fallback to prevent a crash in
            // the eventuality that it would.
            return [this, RelativePosition.BEFORE];
        }
    }
}
```

In particular, your code will fail in the event that some node that is not a `FontAwesomeNode` ends up rendering itself using a node that your regex would recognize as being a fontawesome. If it does, the first invisible character would be mistaken for the second one because its "previousSibling" would be recognized as a fontawesome even though it is not the one that this invisible character is "attached" to. Hence my suggestion of using the `renderings` object to specifically work on the nodes rendered by that VNode and only that.

packages/core/test/todo.md
```diff
@@ -9,9 +9,8 @@ When implementing the tests for the delete, backspace and enter keys, certain te
         1. [Unbreakable](#f-t-unbreakable)
         2. [PRE](#f-t-pre)
         3. [Tables](#f-t-tables)
-        4. [Pictograms](#f-t-pictograms)
-        5. [HR](#f-t-hr)
-        6. [Integration](#f-t-integration)
+        4. [HR](#f-t-hr)
+        5. [Integration](#f-t-integration)
```
dmo-odoo: Markdown is bad if you truly have to manually re-order every item of the list when you remove one in the middle.

packages/core/src/EventNormalizer.ts
```diff
@@ -993,7 +994,7 @@ export class EventNormalizer {
             if (element) {
                 if (
                     !dataTransfer.draggingFromEditable &&
-                    element.nodeName === 'A' &&
+                    nodeName(element) === 'A' &&
```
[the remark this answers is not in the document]
Zinston: Yes it is indeed when I tried parsing XML that I noticed it could be lowercase and I made this commit as a response...

examples/demo/demo.ts
Zinston: I tried moving these imports to `Fontawesome.ts` which would be cool as then the fonts are loaded automatically when you install the plugin. But... Then for some reason the tests don't work. To me this is a question for... @Goaman ! Is it a plane? Is it a bird? No! It's GOAMAN!

packages/plugin-toolbar/src/Toolbar.ts
```diff
@@ -3,6 +3,16 @@ import { OwlUI } from '../../owl-ui/src/OwlUI';
...
+import '!style-loader!css-loader!@fortawesome/fontawesome-free/css/all.css';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-brands-400.woff';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-brands-400.woff2';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-brands-400.ttf';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-regular-400.woff';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-regular-400.woff2';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-regular-400.ttf';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-solid-900.woff';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-solid-900.woff2';
+import '!file-loader?name=./fonts/[name].[ext]!@fortawesome/fontawesome-free/webfonts/fa-solid-900.ttf';
```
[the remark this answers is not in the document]
Zinston: See above comment. Unclear.

examples/demo/demo.xml
```diff
@@ -11,14 +12,14 @@ within the back-to-front world of Looking-Glass Land.
 <p>
 "Jabberwocky" is considered one of the greatest nonsense poems written in
 English. Its playful, whimsical language has given English nonsense words and
-neologisms such as " galumphing" and "chortle".
+neologisms such as "galumphing" and "chortle".
```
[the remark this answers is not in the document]
Zinston: While I was modifying the template anyway I corrected that extra space that was bothering me...

examples/demo/demo.xml
```diff
@@ -11,14 +12,14 @@ within the back-to-front world of Looking-Glass Land.
...
 </p>
 
 
 <h1>Table examples</h1>
 
 <p style="color: red;"><u><b>Disclaimer:</b></u> Tables will only be fully
-implemented once <i>unbreakables nodes</i> are implemented.<br/>In the meantime
+implemented once <i>unbreakable nodes</i> are implemented.<br/>In the meantime
```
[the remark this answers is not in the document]
Zinston: I saw :tada:

packages/plugin-fontawesome/src/FontawesomeDomParser.ts
```diff
@@ -0,0 +1,34 @@
+import { FontawesomeNode } from './FontawesomeNode';
+import { AbstractParser } from '../../plugin-parser/src/AbstractParser';
+import { DomParsingEngine } from '../../plugin-dom/src/DomParsingEngine';
+
+export class FontawesomeDomParser extends AbstractParser<Node> {
+    static id = 'dom';
+    engine: DomParsingEngine;
+
+    predicate = (item: Node): boolean => {
+        return this._isFontawesome(item);
```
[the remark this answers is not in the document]
Zinston: Well it becomes a static function later so we can use it in the buttons. But I guess I _could_ make the regex public instead.

packages/plugin-fontawesome/src/FontawesomeDomParser.ts
```diff
@@ -0,0 +1,34 @@
...
+    };
+
+    async parse(item: Element): Promise<FontawesomeNode[]> {
+        const fontawesome = new FontawesomeNode(item.nodeName);
```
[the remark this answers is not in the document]
Zinston: I don't know if that in particular would work as it's a block. But when writing tests I realized it was a tad disturbing to have `contentBefore = <span class="fa-whatever"/>` become `contentAfter = <i class="fa-whatever"/>` or something like that so I figured I'd preserve the `nodeName` from the given template.

packages/plugin-fontawesome/src/FontawesomeNode.ts
```diff
@@ -0,0 +1,8 @@
+import { InlineNode } from '../../plugin-inline/src/InlineNode';
+
+export class FontawesomeNode extends InlineNode {
+    static readonly atomic = true;
```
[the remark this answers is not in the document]
Zinston: Well not in the DOM at least but it's true that in our abstraction it might be that they are: non-atomic inline nodes are parsed as formats.

packages/plugin-dom/src/Dom.ts
```diff
@@ -222,7 +222,7 @@ export class Dom<T extends DomConfig = DomConfig> extends JWPlugin<T> {
         if (forceAfter) {
             return [reference, RelativePosition.AFTER];
         }
-        if (forcePrepend) {
+        if (forcePrepend && !reference.atomic) {
```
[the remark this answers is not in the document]
Zinston: It was because it undoes some of the things that are done in the previous commit. But I don't mind keeping it separate.

packages/plugin-fontawesome/src/FontawesomeNode.ts
Zinston: Right but if I do that I have to move `locate` for everything else, don't I? I think it's a little bit beyond the scope of this PR.

dmo-odoo: Well yes you'd have to move `locate` for everything else. It's out of the scode of this PR but it will have to be done eventually, as the code is soemwhat wrong here.

packages/plugin-fontawesome/src/FontawesomeDomParser.ts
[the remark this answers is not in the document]
Zinston: @dmo-odoo Want me to do it?

packages/plugin-fontawesome/src/FontawesomeNode.ts
Zinston: OK I'm opening an issue for it.

packages/plugin-fontawesome/src/FontawesomeDomParser.ts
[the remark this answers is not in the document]
dmo-odoo: I guess we can leave it like that for now, I just ask you to remember to back here when `locate` is moved in the renderer, making this function used only in one place again.

examples/demo/demo.ts
Goaman: Which tests don't work? I tried to remove theses from demo.ts and everything works fine.

webpack.config.js
Goaman: True that for now, we don't need them as we do not import those format without "inline-import".
We can investigate in the future for having those files automatically configured in one place. But right now it seems that the css-loader does not inform webpack properly about the files it tries to access. Thus webpack do not try to load them.

examples/demo/demo.ts
Zinston: Did you put them in `Fontawesome.ts`, remove them from `demo.ts` and then run all the tests?

webpack.config.js
Zinston: Ok so should I remove these lines for the time being?

examples/demo/demo.ts
Zinston: @Goaman @dmo-odoo

webpack.config.js
Zinston: @Goaman ?

Goaman: Yep, fine with me.

packages/plugin-fontawesome/src/FontawesomeDomParser.ts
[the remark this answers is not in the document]
Zinston: Issue #144

packages/plugin-fontawesome/src/FontawesomeNode.ts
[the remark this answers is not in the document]
Zinston: So I checked and `InlineNode` is not by default atomic. But it probably could be.

PR #139, opened 2020-03-20
[FIX] Table: properly parse nested tables

Zinston:
Prior to this fix, a table within a table could not be parsed, and that would make the whole editor crash (this happened on _Jabberwockipedia_).


packages/plugin-table/src/TableDomParser.ts
```diff
@@ -68,14 +68,28 @@ export class TableDomParser extends AbstractParser<Node> {
...
     _getTableDimensions(domTable: HTMLTableElement): [number, number] {
-        const domRows = Array.from(domTable.querySelectorAll('tr'));
+        const domRows = Array.from(domTable.querySelectorAll('tr'))
+        const domTableRows = domRows.filter(row => this._domTableAncestor(row) === domTable);
         let columnCount = 0;
-        if (domRows.length) {
-            for (const domChild of domRows[0].querySelectorAll('td, th')) {
+        if (domTableRows.length) {
+            const domCells = Array.from(domTableRows[0].querySelectorAll('td, th'));
+            const domTableCells = domCells.filter(cell => this._domTableAncestor(cell) === domTable);
```
dmo-odoo: The dom have `closest`, doesn't it ? I suggest trying with `closest` in order to reove this auxiliary function.

Zinston: Man good job, I didn't know this existed, I thought that was only in jQuery...

PR #142, opened 2020-03-24
[ADD] Plugins: Pre

Zinston:
Todo: tests.
Question: is it important to render the contents of the `<pre>` as is ? - should we change rendering of chars in `<pre>` so that multiple spaces don't insert `\u00A0`s?

### DIFF
--- a/examples/demo/demo.xml
+++ b/examples/demo/demo.xml
@@ -15,6 +15,60 @@ English. Its playful, whimsical language has given English nonsense words and
 neologisms such as "galumphing" and "chortle".
 </p>
 
+<h2>The Mouse's Tale</h2>
+<h3>by Lewis Carroll</h3>
+<pre>
+"Fury said to
+      a mouse, That
+          he met in the
+              house, 'Let
+                 us both go
+                   to law: <i>I</i>
+                     will prose-
+                      cute you.—
+                     Come, I'll
+                  take no de-
+                nial: We
+             must have
+          the trial ;
+        For really
+      this morn-
+    ing I've
+   nothing
+  to do.'
+   Said the
+    mouse to
+     the cur,
+      'Such a
+        trial, dear
+           Sir, With
+             no jury
+               or judge,
+                 would
+                be was-
+              ting our
+            breath.'
+         'I'll be
+       judge,
+     I'll be
+    jury,'
+   Said
+  cun-
+ ning
+   old
+    Fury;
+     'I'll
+       try
+         the
+          whole
+           cause,
+            and
+           con-
+          demn
+      you to
+    death.'"
+</pre>
+
 
 <h1>Table examples</h1>
 
--- a/packages/plugin-pre/src/Pre.ts
+++ b/packages/plugin-pre/src/Pre.ts
@@ -1,0 +1,50 @@
+import { JWPlugin, JWPluginConfig } from '../../core/src/JWPlugin';
+import { Loadables } from '../../core/src/JWEditor';
+import { Parser } from '../../plugin-parser/src/Parser';
+import { Renderer } from '../../plugin-renderer/src/Renderer';
+import { PreDomParser } from './PreDomParser';
+import { PreDomRenderer } from './PreDomRenderer';
+import { CommandParams } from '../../core/src/Dispatcher';
+import { VNode, isLeaf } from '../../core/src/VNodes/VNode';
+import { distinct } from '../../utils/src/utils';
+import { PreNode } from './PreNode';
+
+export class Pre<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T> {
+    commands = {
+        applyPreStyle: {
+            handler: this.applyPreStyle,
+        },
+    };
+    readonly loadables: Loadables<Parser & Renderer> = {
+        parsers: [PreDomParser],
+        renderers: [PreDomRenderer],
+    };
+
+    //--------------------------------------------------------------------------
+    // Public
+    //--------------------------------------------------------------------------
+
+    /**
+     * Change the formatting of the nodes in given range to Pre.
+     *
+     * @param params
+     */
+    applyPreStyle(params: CommandParams): void {
+        const range = params.context.range;
+        let nodesToConvert: VNode[];
+        if (range.isCollapsed()) {
+            nodesToConvert = [range.start.parent];
+        } else {
+            const selectedLeaves = range.selectedNodes(isLeaf);
+            nodesToConvert = distinct(
+                selectedLeaves.map(leaf => (leaf.atomic ? leaf.parent : leaf)),
+            );
+        }
+        for (const node of nodesToConvert) {
+            const pre = new PreNode();
+            pre.attributes = { ...node.attributes };
+            node.before(pre);
+            node.mergeWith(pre);
+        }
+    }
+}
--- a/packages/plugin-pre/src/Pre.ts
+++ b/packages/plugin-pre/src/Pre.ts
@@ -5,8 +5,6 @@ import { Renderer } from '../../plugin-renderer/src/Renderer';
 import { PreDomParser } from './PreDomParser';
 import { PreDomRenderer } from './PreDomRenderer';
 import { CommandParams } from '../../core/src/Dispatcher';
-import { VNode, isLeaf } from '../../core/src/VNodes/VNode';
-import { distinct } from '../../utils/src/utils';
 import { PreNode } from './PreNode';
 
 export class Pre<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T> {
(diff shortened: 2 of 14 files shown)


examples/demo/demo.xml
```diff
@@ -14,6 +14,60 @@ English. Its playful, whimsical language has given English nonsense words and
...
+       try
+         the
+          whole
+           cause,
+            and
+           con-
+          demn
+      you to
+    death.'"
+</pre>
```
dmo-odoo: Well that's kind of sad, isn't it ? 🐭

packages/plugin-pre/src/Pre.ts
```diff
@@ -0,0 +1,50 @@
...
+import { PreDomParser } from './PreDomParser';
+import { PreDomRenderer } from './PreDomRenderer';
+import { CommandParams } from '../../core/src/Dispatcher';
+import { VNode, isLeaf } from '../../core/src/VNodes/VNode';
+import { distinct } from '../../utils/src/utils';
+import { PreNode } from './PreNode';
+
+export class Pre<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T> {
+    commands = {
+        applyPre: {
```
dmo-odoo: I'm not convinced by the name of this command but to be fair I don't have a better idea yet. Though if we used `applyHeading` for Headings then it makes sense I guess. I think to remember we used `applyHeadingStyle` so I suggest `applyPreStyle` ? Maybe that's a tad too much ? 🐸

packages/plugin-pre/src/Pre.ts
```diff
@@ -0,0 +1,50 @@
...
+    applyPre(params: CommandParams): void {
+        const range = params.context.range;
+        let nodesToConvert: VNode[];
+        if (range.isCollapsed()) {
+            nodesToConvert = [range.start.parent];
+        } else {
+            const selectedLeaves = range.selectedNodes(isLeaf);
+            nodesToConvert = distinct(
+                selectedLeaves.map(leaf => (leaf.atomic ? leaf.parent : leaf)),
+            );
```
dmo-odoo: Nooo, you did that construction again... 😿 Could you please try `targetedNodes` to see if it could match your need ? If it does not then we need to find another way to avoid this leaves + map + parent/leaf + distinct construction.

packages/plugin-pre/src/Pre.ts
```diff
@@ -0,0 +1,50 @@
...
+        }
+        for (const node of nodesToConvert) {
+            const pre = new PreNode();
+            pre.attributes = { ...node.attributes };
+            node.before(pre);
+            node.mergeWith(pre);
```
dmo-odoo: I see you, `replaceWith` 👁

packages/plugin-pre/src/PreDomParser.ts
```diff
@@ -0,0 +1,21 @@
+import { PreNode } from './PreNode';
+import { AbstractParser } from '../../plugin-parser/src/AbstractParser';
+import { DomParsingEngine } from '../../plugin-dom/src/DomParsingEngine';
+
+export class PreDomParser extends AbstractParser<Node> {
+    static id = 'dom';
+    engine: DomParsingEngine;
+
+    predicate = (item: Node): boolean => {
+        return item instanceof Element && item.tagName === 'PRE';
```
dmo-odoo: I guess this would be updated to use the `nodeName` utils of the other branch ?

packages/plugin-pre/src/PreDomParser.ts
```diff
@@ -0,0 +1,21 @@
...
+    };
+
+    async parse(item: Element): Promise<PreNode[]> {
+        const pre = new PreNode();
+        pre.attributes = this.engine.parseAttributes(item);
+        const children = await this.engine.parse(...item.childNodes);
+        // TODO: parse contained \n as LineBreaks.
```
dmo-odoo: So, in my branch, you could use `new this.engine.editor.configuration.defaults.Separator()` to create a new "separator" using the editor configuration.

packages/plugin-pre/src/PreDomRenderer.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Render the VNode to the given format.
+     */
+    async render(node: PreNode): Promise<Node[]> {
+        const pre = document.createElement('pre');
+        this.engine.renderAttributes(node.attributes, pre);
+        for (const child of node.children()) {
+            let domChildren = await this.renderChild(child);
+            // TODO: prevent CharNode from rendering spaces as nbsp.
```
dmo-odoo: A possible way would to load a special parser which predicate recognizes CharNode inside PreNode but I fear that would lead to quite the duplication I think...

packages/plugin-pre/src/PreDomRenderer.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Replace every <br> in a list of DOM Nodes with newlines.
+     *
+     * @param domNodes
+     */
+    _brToNewline(domNodes: Node[]): Node[] {
+        if (domNodes.length > 1) {
+            // Remove the last <br> if it's a placeholder (a trailing BR after
+            // another <br>, added just to make the first one visible).
```
dmo-odoo: This is bad design on our part (somewhat) because we're manually undoing the work that we manually did in the `LineBreakDomRenderer`. I guess we should have a `PreSeparatorDomRenderer` which predicate is:
```node.ancestor(PreNode) && node instanceof this.engine.editor.configuration.defaults.Separator```
and never let `LineBreakDomRenderer` intervene here.

File: packages/plugin-pre/src/PreNode.ts [hunk 9]
```diff
@@ -0,0 +1,7 @@
+import { VElement } from '../../core/src/VNodes/VElement';
+
+export class PreNode extends VElement {
```
dmo-odoo: You can extend `VNode` rather than `VElement` because you're not using the `htmlTag` feature of `VElement` anyway. (And we will eventually remove `VElement` from the core anyway)

packages/utils/src/formattingSpace.ts
```diff
@@ -12,6 +12,9 @@ import { isBlock } from './isBlock';
 export function removeFormattingSpace(node: Node): string {
     // TODO: check the value of the `white-space` property
     const text: string = node.textContent;
+    if ((node.parentNode as Element)?.closest('PRE')) {
```
dmo-odoo: I think there is a `parent` attribute which is an Element rather than a node.

packages/plugin-pre/src/Pre.ts
dmo-odoo: Fail on me for merging `Heading` with this construction then. Would you care to try with `targetedNodes` in `Heading` and update it here ?

packages/plugin-pre/src/Pre.ts
dmo-odoo: Let's do that then.

packages/plugin-pre/src/PreDomRenderer.ts
dmo-odoo: If it's parsed as the character `\n` then it's not a br and does not need to be converted now, does it ?

packages/plugin-pre/src/Pre.ts
Zinston: Separate commit?

packages/plugin-pre/src/PreNode.ts
dmo-odoo: It's because our rendering and parsing logic is outside of the node itself.

packages/utils/src/formattingSpace.ts
dmo-odoo: Minimum effort here, huh ? 🚎 See `Node.parentElement`.

packages/plugin-pre/src/PreDomRenderer.ts
Zinston: Right :duck:

Zinston:
@dmo-odoo Ready.

packages/plugin-pre/src/PreDomParser.ts
Zinston: https://github.com/odoo-dev/jabberwock/pull/162/commits/54c05ed308fde232bf092c3be2d0b1f7d8ed85fa

packages/plugin-pre/src/PreDomRenderer.ts
Zinston: https://github.com/odoo-dev/jabberwock/pull/162/commits/05495ed7e1359a8b8f5fd12c85e4da15b23f2e5e

packages/plugin-pre/src/PreDomRenderer.ts
Zinston: https://github.com/odoo-dev/jabberwock/pull/162/commits/35c8ac3640c40b884d8189a8b1ee9026a060ada5

PR #143, opened 2020-03-26
Unbreakable

dmo-odoo:


packages/core/src/VNodes/ChildNode.ts
```diff
@@ -0,0 +1,484 @@
+import { Constructor, nodeLength } from '../../../utils/src/utils';
+import { VNode, RelativePosition, isLeaf, Predicate, Typeguard } from './VNode';
+
+export abstract class ChildNode extends VNode {
```
Zinston: I don't really get the added value of separating these methods from other `VNode` methods. Plus, why is it called `ChildNode`? It's confusing.

packages/core/src/VNodes/VNode.ts
```diff
@@ -618,261 +316,90 @@ export class VNode {
...
      */
-    append(...children: VNode[]): void {
-        for (const child of children) {
-            this._insertAtIndex(child, this.children.length);
-        }
-    }
+    abstract append(...children: VNode[]): void;
     /**
-     * Insert the given node before the given reference (which is a child of
+     abstract * Insert the given node before the given reference (which is a child of
```
Zinston: fail

packages/core/src/VNodes/VNode.ts
```diff
@@ -618,261 +316,90 @@ export class VNode {
...
-        const index = this.children.indexOf(reference);
-        if (index < 0) {
-            throw new Error('The given VNode is not a child of this VNode');
-        }
-        this._insertAtIndex(node, index);
-    }
+    abstract insertBefore(node: VNode, reference: VNode): void;
     /**
-     * Insert the given node after the given reference (a child of this VNode).
+     abstract * Insert the given node after the given reference (a child of this VNode).
```
Zinston: fail

packages/core/src/VNodes/VNode.ts
```diff
@@ -618,261 +316,90 @@ export class VNode {
...
-        const index = this.children.indexOf(child);
-        if (index < 0) {
-            throw new Error('The given VNode is not a child of this VNode');
-        }
-        this._removeAtIndex(index);
-    }
+    abstract removeChild(child: VNode): void;
     /**
-     * Remove this node in forward direction. (e.g. `Delete` key)
+     abstract * Remove this node in forward direction. (e.g. `Delete` key)
```
Zinston: gnnn

packages/core/src/VNodes/VNode.ts
```diff
@@ -618,261 +316,90 @@ export class VNode {
...
      */
-    removeForward(): void {
-        this.remove();
-    }
+    abstract removeForward(): void;
     /**
-     * Remove this node in backward direction. (e.g. `Backspace` key)
+     abstract * Remove this node in backward direction. (e.g. `Backspace` key)
```
Zinston: idem

packages/plugin-char/src/CharNode.ts
```diff
@@ -41,22 +41,6 @@ export class CharNode extends InlineNode {
...
-    get length(): number {
-        return 1;
-    }
-    /**
-     * Return this VNode's inner text (concatenation of all descendent
-     * char nodes values).
-     *
-     * @param __current
-     */
-    text(__current = ''): string {
```
Zinston: Why did you remove this? Was it moved somewhere else instead?

packages/plugin-linebreak/src/LineBreakNode.ts
```diff
@@ -1,11 +1,8 @@
-import { VElement } from '../../core/src/VNodes/VElement';
 import { VNode, RelativePosition } from '../../core/src/VNodes/VNode';
+import { AtomicNode } from '../../core/src/VNodes/AtomicNode';
 
-export class LineBreakNode extends VElement {
+export class LineBreakNode extends AtomicNode {
     static readonly atomic = true;
```
Zinston: Is this still needed if it extends `AtomicNode`?

packages/core/src/JWEditor.ts
```diff
@@ -34,8 +36,12 @@ type CommandParams<T extends JWPlugin, K extends string> = K extends Commands<T>
     : never;
 
 export interface JWEditorConfig {
+    defaults?: {
+        Container?: new () => VNode & { atomic: false };
+        Atomic?: new () => VNode & { atomic: true };
+        Separator?: new () => VNode;
```
Zinston: Why isn't it `new AtomicNode()` etc.?

packages/core/src/Core.ts
```diff
@@ -47,7 +47,18 @@ export class Core<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
      * Insert a paragraph break.
      */
     insertParagraphBreak(params: InsertParagraphBreakParams): void {
-        this.editor.vDocument.insertParagraphBreak(params.context.range);
+        const range = params.context.range;
```
Zinston: Typo in the name of this commit.

packages/core/src/Core.ts
```diff
@@ -47,7 +47,18 @@ export class Core<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
...
+        // Remove the contents of the range if needed.
+        if (!range.isCollapsed()) {
+            range.empty();
+        }
+        if (range.startContainer.breakable) {
+            range.startContainer.splitAt(range.start);
+        } else {
+            // Use a separator to break paragraphs in an unbreakable.
```
Zinston: :+1:

examples/demo/demo.ts
```diff
@@ -18,7 +18,10 @@ import { OrderedListButton } from '../../packages/plugin-list/src/ListButtons';
 import { UnorderedListButton } from '../../packages/plugin-list/src/ListButtons';
 import { IndentButton } from '../../packages/plugin-indent/src/IndentButtons';
 import { OutdentButton } from '../../packages/plugin-indent/src/IndentButtons';
-import { AlignLeftButton, AlignCenterButton, AlignRightButton, AlignJustifyButton } from '../../packages/plugin-align/src/AlignButtons';
+import { AlignLeftButton } from '../../packages/plugin-align/src/AlignButtons';
+import { AlignCenterButton } from '../../packages/plugin-align/src/AlignButtons';
+import { AlignRightButton } from '../../packages/plugin-align/src/AlignButtons';
+import { AlignJustifyButton } from '../../packages/plugin-align/src/AlignButtons';
```
Zinston: What was the issue? I wasn't aware it was not allowed to have several imports on the same line.

examples/demo/demo.xml
```diff
@@ -17,11 +17,6 @@ neologisms such as " galumphing" and "chortle".
 
 <h1>Table examples</h1>
 
-<p style="color: red;"><u><b>Disclaimer:</b></u> Tables will only be fully
-implemented once <i>unbreakables nodes</i> are implemented.<br/>In the meantime
```
Zinston: Yeah :+1: :tada:

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -97,7 +97,9 @@ describe('Plugin: DevTools', () => {
     describe('InspectorComponent', () => {
         it('should change panel to "Inspector" with top button', async () => {
             await openDevTools(devtools);
-            const buttonCommand = devtools.querySelector('devtools-navbar devtools-button:nth-child(2)');
+            const buttonCommand = devtools.querySelector(
+                'devtools-navbar devtools-button:nth-child(2)',
+            );
```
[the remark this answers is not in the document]
dmo-odoo: Well, eslint+prettier I guess... =/ You want me to undo it ? (it might come back on save though)

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -205,15 +209,17 @@ describe('Plugin: DevTools', () => {
...
                 '<devtools-id>' +
                 editor.vDocument.root.children()[1].id +
                 '</devtools-id>' +
                 '</devtools-about>';
             expect(about.outerHTML).to.equal(aResult);
 
-            const aboutMe = devtools.querySelector('devtools-info devtools-properties devtools-table');
+            const aboutMe = devtools.querySelector(
+                'devtools-info devtools-properties devtools-table',
+            );
```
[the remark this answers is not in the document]
dmo-odoo: same here

packages/plugin-devtools/test/devtools.test.ts
```diff
@@ -434,16 +439,18 @@ describe('Plugin: DevTools', () => {
                 '<devtools-type>' +
                 vNodeChar.constructor.name +
                 '</devtools-type> ' +
-                vNodeChar.type +
+                vNodeChar.name +
                 ': "b" ' +
                 '<devtools-button class="logger">&gt;_</devtools-button>' +
                 '<devtools-id>' +
-                + vNodeChar.id +
+                +vNodeChar.id +
```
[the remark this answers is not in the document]
dmo-odoo: That's weird, really... =/

packages/plugin-dom/src/DefaultDomRenderer.ts
```diff
@@ -10,29 +10,30 @@ export class DefaultDomRenderer extends AbstractRenderer<Node[]> {
...
-                domNode.appendChild(domChild);
+            const renderedChildren = await this.renderChildren(node);
+            for (const renderedChild of renderedChildren) {
+                for (const domChild of renderedChild) {
+                    domNode.appendChild(domChild);
+                }
             }
+            return [domNode];
+        } else {
+            return [];
```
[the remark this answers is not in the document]
dmo-odoo: It was wrong to begin with. I have been longing to remove since the day I merged it.

packages/core/src/Core.ts
```diff
@@ -67,13 +67,22 @@ export class Core<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
...
+                let previous = range.startContainer.previous();
+                while (previous && !isLeaf(previous)) {
+                    const sibling = previous.previousSibling();
+                    if (sibling) {
+                        previous = sibling.lastLeaf();
+                    } else if (previous.parent?.breakable) {
+                        previous = previous.parent;
+                    } else {
+                        // Stop browsing at the first child of an unbreakable.
+                        previous = undefined;
```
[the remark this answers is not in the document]
dmo-odoo: I don't like to use `return` in weird places because it makes the flow of the code hard to grasp. Though I'll have a second look to this function to see if I can improve it.

packages/core/src/VNodes/AtomicNode.ts
```diff
@@ -0,0 +1,200 @@
...
+import { ChildNode } from './ChildNode';
+import { AtomicityError } from '../../../utils/src/errors';
+
+/**
+ * This class provides typing overrides for multiple VNode methods which are
+ * supposed to take parameters but that are unused in the case of atomic nodes.
+ */
+/* eslint-disable @typescript-eslint/no-unused-vars */
+export class AtomicNode extends ChildNode {
+    static readonly atomic: true = true;
```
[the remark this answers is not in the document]
dmo-odoo: So that TypeScript can warn me if it ever detects that I'm testing if an AtomicNode is not atomic.
Something like that would trigger a warning:
```typescript
const node = new AtomicNode();
if (node.atomic) {
    ...
} else {
    // unreachable
}
```
But of course it's rather hypothetical, because in most cases the node's type will be `VNode` so Typescript won't be able to use the specialized typings I have defined in this file. It might be useful one day if someone makes a mistake at one point though.

packages/core/src/VNodes/AtomicNode.ts
```diff
@@ -0,0 +1,200 @@
...
+    nthChild(n: number): undefined {
+        return undefined;
+    }
+    /**
+     * See {@link VNode.length}.
+     *
+     * @return Returns 1 since an atomic node cannot have children.
+     */
+    get length(): 1 {
+        return 1;
```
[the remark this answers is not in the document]
dmo-odoo: Yes, I think there is something weird with this `length` thing. It was counting children yes, except CharNode which is 1, wait what ? I don't know what is the use of this `length` method so you might be the right person to enlighten me on that.

packages/core/src/VNodes/AtomicNode.ts
```diff
@@ -0,0 +1,200 @@
...
+    // Updating
+    //--------------------------------------------------------------------------
+
+    /**
+     * See {@link VNode.prepend}.
+     *
+     * @throws AtomicityError An atomic node cannot have children.
+     */
+    prepend(...children: VNode[]): void {
+        throw new AtomicityError(this);
```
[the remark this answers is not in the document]
dmo-odoo: Well, I had a previous version of this code which used `before` as an implementaton of `prepend` but I think having "magic fallbacks" like that will make errors very hard to find and understand, while a throw will be found and understood directly. We might later have a system that catches errors, restore the previous memory and do a `console.error` though.

packages/plugin-char/src/CharNode.ts
dmo-odoo: You have quite the keen eye. 🦅 I removed it because it was not used anywhere in production code.

packages/plugin-linebreak/src/LineBreakNode.ts
dmo-odoo: Nope, you're right.

packages/core/src/VNodes/ChildNode.ts
dmo-odoo: I want VNode to be the lightest possible. In my mind, VNode is an interface. It is not in practice for the reasons I explained in the commit message, but that answers your question since interfaces cannot have method implementations. I'm open for name suggestions though.

packages/core/src/VNodes/ChildNode.ts
```diff
@@ -0,0 +1,484 @@
...
+     *
+     * @param __repr
+     * @param level
+     */
+    repr(): string {
+        let repr = this.name + ' (' + this.id + ')' + '\n';
+        for (const child of this.childVNodes) {
+            repr += Array(4 + 2).join(' ') + child.repr();
+        }
+        return repr;
```
dmo-odoo: TODO: I'm pretty sure I broke this function, I forgot to test it.

packages/core/src/JWEditor.ts
dmo-odoo: Because people can do a node that is atomic without using our AtomicNode class.

examples/demo/demo.ts
dmo-odoo: It is, but the line was too long and my linter complained about it so I broke it into multiple lines to fix it.

packages/plugin-devtools/test/devtools.test.ts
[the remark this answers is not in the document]
Zinston: No no this is good :-) I was just laughing with how it's slipped in there while it has nothing to do with the commit ^^

packages/core/src/JWEditor.ts
Zinston: Why?

packages/core/src/Core.ts
Zinston: :trolleybus:

packages/core/src/VNodes/VNode.ts
```diff
@@ -552,6 +476,126 @@ export abstract class VNode {
...
+     * Return the last descendant of this VNode that satisfies the predicate.
+     * If no predicate is given, return the last descendant of this VNode.
+     *
+     * @param [predicate]
+     */
+    abstract lastDescendant<T extends VNode>(predicate?: Predicate<T>): T;
+    abstract lastDescendant<T>(predicate?: Predicate<T>): VNode;
+
+    //--------------------------------------------------------------------------
+    // Updating children. To be implemented by the concrete subclass.
```
Zinston: Now we're talking :+1:

packages/core/src/VRange.ts
```diff
@@ -188,7 +189,7 @@ export class VRange {
         } else {
             reference = reference.lastLeaf();
         }
-        if (!reference.hasChildren() && !reference.atomic) {
+        if (reference.is(ContainerNode) && !reference.hasChildren()) {
```
Zinston: :+1:

packages/core/src/VNodes/VNode.ts
```diff
@@ -28,14 +28,6 @@ export abstract class VNode {
     parent: VNode;
     attributes: Record<string, string | Record<string, string>> = {};
     childVNodes: VNode[];
-    /**
-     * A node is atomic when it is not allowed to have children in the
-     * abstraction. Its real-life children should be ignored in the abstraction.
-     * This getter allows us to call `this.atomic` on the extensions of `VNode`
-     * while declaring the `atomic` property in a static way.
-     */
-    readonly atomic: boolean;
```
Zinston: Have you made sure that devtools don't break?

packages/core/src/Core.ts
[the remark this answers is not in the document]
dmo-odoo: I could not find a better way of doing it. And using `return` doesn't cut it for me. This `undefined` means something, it symbolizes the fact that the "previous" of an unbreakable, in the context of this command, is `undefined`. It's what an hypothetical `previousButDontBreakUnbreakables` would return and for me the while *is* that hypothetical function. The fact that we end up not doing anything later on if there is no `previous` is outside of the scope of the `while`.

In other words, the responsibility of the `while` block here is to compute the value of `previous` in this context. It turns out that the computed value of `previous` in this context is `undefined`. I don't want the while to take control of the entire command and decide that the command will stop there just because it was unable to compute a meaningful value for `previous`, it is outside of the scope I designed it for.

packages/core/src/VNodes/AtomicNode.ts
```diff
@@ -102,13 +102,13 @@ export class AtomicNode extends VNode {
...
+     * See {@link AbstractNode.descendants}.
      *
      * @return Returns an empty array since an atomic node cannot have children.
      */
-    descendants<T extends VNode>(predicate?: Predicate<T>): [];
-    descendants<T>(predicate?: Predicate<T>): [];
-    descendants<T>(predicate?: Predicate<T>): [] {
+    descendants<T extends VNode>(predicate?: Predicate<T>): VNode[];
+    descendants<T>(predicate?: Predicate<T>): VNode[];
+    descendants<T>(predicate?: Predicate<T>): VNode[] {
```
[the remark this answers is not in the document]
dmo-odoo: There is a limitation with what you can write as a type. If you write `VNode[]` then it gives no information on the length of the array. If you write `[]` like I used to do, it's an empty tuple, which is technically correct, but now Typescript does not know that these are VNodes, so writing this:
```typescript
const children = atomic.children();
children.indexOf(node):
```
Results in a typeing error because the type of `children` is `[]` so the type of the parameter of `indexOf` is `never`, which is true in some way, since there is no element that is going to fit in there. However, it's kind of sad that I can't just call my `indexOf` and get my `-1` as a result. From an other point of view, if typescript is able to realize that the node is an `AtomicNode` and not merely any `VNode`, it might actually be interesting that I stumble upon this typing warning because there is no point in calling `indexOf` and getting `-1` since the children already will *always* be empty. I'll change it back to `[]` and see what happens.

packages/core/src/VNodes/AbstractNode.ts
```diff
@@ -0,0 +1,649 @@
...
+    /**
+     * Return whether this node is an instance of the given VNode class.
+     *
+     * @param predicate The subclass of VNode to test this node against.
+     */
+    is<T extends VNode>(predicate: Constructor<T> | Typeguard<T>): this is T {
+        if (AbstractNode.isConstructor(predicate)) {
+            return this instanceof predicate;
+        } else {
+            return predicate(this as VNode);
```
dmo-odoo: Yeah... that's the main drawback of this approach. `AbstractNode` itself does not satisfy the constraint of `VNode` since it is neither `ContainerNode` nor `AtomicNode` by itself. If you have ideas on how to solve that I'm open to suggestions.

packages/core/src/VNodes/AtomicNode.ts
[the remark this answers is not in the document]
Zinston: Yes I think it can be a good thing to get that typing error but on the other hand not if it means having to check all the time if the node is atomic where you could just do nothing if `indexOf` returns `-1`.

packages/core/src/VNodes/AbstractNode.ts
Zinston: Why not `T extends AbstractNode` ?

packages/core/src/VNodes/AtomicNode.ts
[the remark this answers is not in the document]
dmo-odoo: If it's a `VNode`, so not narrowed down to `AtomicNode`, `indexOf` will work as expected.

PR #146, opened 2020-03-27
[ADD] Plugins: TextColor, BackgroundColor

Zinston:


packages/plugin-backgroundcolor/src/BackgroundColor.ts
```diff
@@ -0,0 +1,164 @@
...
+
+export class BackgroundColor<T extends BackgroundColorConfig = BackgroundColorConfig> extends JWPlugin<T> {
+    static hasColor(node: VNode): boolean;
+    static hasColor(backgroundColor: string, node: VNode): boolean;
+    static hasColor(backgroundColor: string | VNode, node?: VNode): boolean {
+        if (backgroundColor instanceof VNode) {
+            node = backgroundColor;
+        }
+        const nodeBackgroundColor = getStyle(node, 'background-color');
+        if (!nodeBackgroundColor || backgroundColor instanceof VNode) {
```
dmo-odoo: I'm susprised this condition isn't simply `backgroundColor instanceof VNode` ?

packages/plugin-backgroundcolor/src/BackgroundColor.ts
```diff
@@ -0,0 +1,164 @@
...
+            return !!nodeBackgroundColor;
+        } else {
+            return nodeBackgroundColor === backgroundColor;
+        }
+    }
+    static dependencies = [Inline];
+    configuration = { ...defaultConfiguration, ...this.configuration };
```
dmo-odoo: That's rather out of scope of this PR but what about having something automatic like this:
```typescript
class JWPlugin<T extends JWPluginConfiguration> {
    defaults: T;
    configuration: T = { ...this.defaults, ...this.configuration};
}
```

packages/plugin-backgroundcolor/src/BackgroundColor.ts
```diff
@@ -0,0 +1,164 @@
...
+}
+
+const defaultConfiguration = {
+    defaultBackgroundColor: 'white',
+};
+
+export class BackgroundColor<T extends BackgroundColorConfig = BackgroundColorConfig> extends JWPlugin<T> {
+    static hasColor(node: VNode): boolean;
+    static hasColor(backgroundColor: string, node: VNode): boolean;
+    static hasColor(backgroundColor: string | VNode, node?: VNode): boolean {
```
dmo-odoo: If you renamed `hasBackgroundColor` into `hasColor` then why not also renaming the parameter ? It's the static method of `BackgroundColor`, of course `color` is going to refer to the background color... ;-)

packages/plugin-backgroundcolor/src/BackgroundColor.ts
```diff
@@ -0,0 +1,164 @@
...
+            node = backgroundColor;
+        }
+        const nodeBackgroundColor = getStyle(node, 'background-color');
+        if (!nodeBackgroundColor || backgroundColor instanceof VNode) {
+            return !!nodeBackgroundColor;
+        } else {
+            return nodeBackgroundColor === backgroundColor;
+        }
+    }
+    static dependencies = [Inline];
```
dmo-odoo: I would put the dependencies first and foremost.

packages/plugin-backgroundcolor/src/BackgroundColor.ts
```diff
@@ -0,0 +1,164 @@
...
+    configuration = { ...defaultConfiguration, ...this.configuration };
+    commands = {
+        colorBackground: {
+            handler: this.colorBackground,
+        },
+        uncolorBackground: {
+            handler: this.uncolorBackground,
+        },
```
dmo-odoo: I'm not 100% convinced by the command names. The `contenteditable` calls it `backColor` and DOM InputEvents call it `formatBackColor`. That does not convince me either. What about `setBackgroundColor` and `resetBackgroundColor` ? I think what bothers me is `uncolor`. Not sure... What do you think ?

packages/plugin-backgroundcolor/src/BackgroundColor.ts
```diff
@@ -0,0 +1,164 @@
...
+        const color = params.color;
+
+        if (params.context.range.isCollapsed()) {
+            // Set the style cache.
+            const inline = this.dependencies.get(Inline);
+            const currentCache = inline.styleCache || {};
+            inline.styleCache = { ...currentCache, 'background-color': color };
+        } else {
+            for (const node of this._highestSelectedNodes(params.context.range)) {
+                // Apply the style to the node or its first format.
```
dmo-odoo: ... so much for not having proper `modifiers` yet...

packages/plugin-backgroundcolor/src/BackgroundColor.ts
```diff
@@ -0,0 +1,164 @@
...
+                // this style, style these ancestors instead of their
+                // descendants.
+                let parent = node.parent;
+                while (parent?.children().every(child => BackgroundColor.hasColor(color, child))) {
+                    setStyle(parent, 'background-color', color);
+                    for (const child of parent.children()) {
+                        removeStyle(child, 'background-color');
+                    }
+                    parent = parent.parent;
+                }
```
dmo-odoo: Actually, the gspec acts weirdly. It sets the color on the container if you set the color its last character. It seems to be the only rule. You don't even need to select everything. Coloring the last character of a container colors the container. Not saying it's great though, but I have the feeling that going futher than the direct parent is overengineering here.

packages/plugin-backgroundcolor/src/BackgroundColor.ts
```diff
@@ -0,0 +1,164 @@
...
+        }
+    }
+    /**
+     * Remove the current background color from the range. If the color was
+     * applied to an ancestor, apply the default color to its relevant inline
+     * descendants.
+     *
+     * @param params
+     */
+    uncolorBackground(params: CommandParams): void {
```
dmo-odoo: Whether it is `uncolorBackground` or `resetBackgroundColor` or whatever, I think it should not set `style=background-color: defaultColor` if it did not originally had a color. I fear that, currently, if i have:
```html
<p>[abcde]</p>
``` 
I do color: red, I get:
```html
<p style="backround-color:red">abcde</p>
```
Alright. Now I select only some chars:
```html
<p  style="backround-color:red">ab[cd]e</p>
```
I do uncolor, I think I'll get:
```html
<p  style="backround-color:red">ab<span style="backround-color:white">[cd]</span>e</p>
```
Isn't that weird ? `background-color: white` ? As a user, I expected `resetBackgroundColor` or `uncolorBackground` to somewhat *undo* the coloration I did on these characters, not *adding* yet another coloring that *happens* to be the one I had before. I'm not sure what to expect in this case though, because I guess I expected this:
```html
<p><span style="backround-color:red">ab</span>[cd]<span style="backround-color:red">e</span></p>
```
Which would also be weird I suppose, since it kinda "undid" the style that was on the `p`. Remind me a bit of what was the point of handling the style on the container ? It ends up making both scenarious looking weird at the end.. :/

packages/utils/src/utils.ts
```diff
@@ -70,3 +73,79 @@ export function distinct<T>(array: Array<T>): Array<T> {
...
 }
+
+/**
+ * Get the target VNode or Format's style attribute as a record of style name to
+ * style value.
+ *
+ * @param target
+ */
+export function getStyles(target: VNode | Format): Record<string, string> {
+    const stylesArray = (target.attributes.style as string || '')
```
dmo-odoo: I'm surprised the type isn't already `string` ?

dmo-odoo:
@Zinston I agree to have a `Color` plugin to handle the mountain of duplicated logic.

packages/plugin-backgroundcolor/src/BackgroundColor.ts
Zinston: Why not indeed. Out of scope indeed :-) Let's put a pin on it.

packages/plugin-backgroundcolor/src/BackgroundColor.ts
Zinston: Yup. Soon...

packages/plugin-backgroundcolor/src/BackgroundColor.ts
Zinston: I don't know. I think this is better than the gspec. It's easy to target the highest and it's safer. Don't forget we don't know in advance what we're going to parse...

packages/plugin-backgroundcolor/src/BackgroundColor.ts
Zinston: Both scenarios _are_ weird I'm afraid... Whatever we do we could parse parents with the style so we have to handle it.

packages/utils/src/utils.ts
Zinston:
```ts
Attributes: Record<string, string | Record<string, string>;
```
It doesn't know that `style` was not a `Record<string, string>`.

packages/plugin-backgroundcolor/src/BackgroundColor.ts
dmo-odoo: My point was that it wouldn't have been weird if the first `color` command had not been applied on the container but on the selected children instead:
```html
<p><span style="backround-color:red">abcde</span></p>
```
Then the second `color` would result in:
```html
<p><span style="backround-color:red">ab</span>[cd]<span style="backround-color:red">e</span></p>
```
Which would not be weird this time since the style was never applied to the parent to begin with. Suboptimal, maybe, but not weird.

I guess you're right with the style on the parent from the original parsing though.

packages/plugin-backgroundcolor/src/BackgroundColor.ts
```diff
@@ -1,36 +1,16 @@
...
-export class BackgroundColor<T extends BackgroundColorConfig = BackgroundColorConfig> extends JWPlugin<T> {
-    static dependencies = [Inline];
-    configuration = { ...defaultConfiguration, ...this.configuration };
+export class BackgroundColor<T extends ColorConfig = ColorConfig> extends Color<T> {
+    styleName = 'background-color';
+    configuration = { defaultColor: 'white', ...this.configuration };
     commands = {
         colorBackground: {
-            handler: this.colorBackground,
+            handler: this.color,
```
dmo-odoo: Here is why the command names are not derived from function names, thank you for demonstrating that.

packages/plugin-color/src/Color.ts
```diff
@@ -0,0 +1,166 @@
...
+    //--------------------------------------------------------------------------
+
+    /**
+     * Return true if the given node has the given color.
+     * If no color is passed, return true if the given node has any color.
+     * Return false otherwise.
+     */
+    hasColor(node: VNode): boolean;
+    hasColor(color: string, node: VNode): boolean;
+    hasColor(color: string | VNode, node?: VNode): boolean {
```
dmo-odoo: I think you'll be srewed when you try to create the color buttons because this function is not static anymore.

packages/plugin-backgroundcolor/src/BackgroundColor.ts
Zinston: Hehe :-)

packages/plugin-color/src/Color.ts
Zinston: Yes I figured as much. But actually this is not the function I should use for the buttons : I think the buttons should check the color of the first ancestor/format that has a color. So we'll implement that when we do have buttons...

PR #152, opened 2020-04-02
dialog

Gorash:


packages/plugin-dialog/src/ui/Dialog.xml
```diff
@@ -0,0 +1,6 @@
+<jw-dialog>
+    <jw-backdrop class="close"></jw-backdrop>
```
dmo-odoo, line 2: Shouldn't the class be something like "jw-close" rather than "close" ?

packages/plugin-dialog/src/ui/DialogZoneDomParser.ts
```diff
@@ -0,0 +1,21 @@
...
+import { DomParsingEngine } from '../../../plugin-dom/src/DomParsingEngine';
+import { AbstractParser } from '../../../plugin-parser/src/AbstractParser';
+import { DialogZoneNode } from './DialogZoneNode';
+
+export class DialogZoneDomParser extends AbstractParser<Node> {
+    static id = 'dom';
+    engine: DomParsingEngine;
+
+    predicate = (item: Node): boolean => {
+        return item instanceof Element && item.tagName.toLowerCase() === 't-dialog';
```
dmo-odoo, line 11: Please use Antoine's `nodeName` util.

packages/plugin-dialog/src/ui/DialogZoneDomRenderer.ts
```diff
@@ -0,0 +1,43 @@
...
+    predicate = DialogZoneNode;
+
+    async render(node: DialogZoneNode): Promise<Node[]> {
+        const float = document.createElement('jw-dialog-container');
+        for (const child of node.children) {
+            if (!node.hidden.get(child)) {
+                float.appendChild(await this._renderDialog(child));
+            }
+        }
+        return float.childNodes.length ? [float] : [document.createDocumentFragment()];
```
dmo-odoo, line 24: You can return `[]` rather then creating an empty fragment.

packages/plugin-dialog/src/ui/DialogZoneDomRenderer.ts
```diff
@@ -0,0 +1,43 @@
+import { DomRenderingEngine } from '../../../plugin-dom/src/DomRenderingEngine';
+import { AbstractRenderer } from '../../../plugin-renderer/src/AbstractRenderer';
+import { DialogZoneNode } from './DialogZoneNode';
+import { VNode } from '../../../core/src/VNodes/VNode';
+import template from './Dialog.xml';
+import './Dialog.css';
+
+const container = document.createElement('jw-container');
+container.innerHTML = template;
+const floatItem = container.firstElementChild;
```
dmo-odoo, line 10: Isn't it possible to assign `outerHTML` and use the element directly rather than having to resort to use its first child ?

packages/plugin-dialog/src/ui/DialogZoneDomRenderer.ts
```diff
@@ -0,0 +1,43 @@
...
+        for (const child of node.children) {
+            if (!node.hidden.get(child)) {
+                float.appendChild(await this._renderDialog(child));
+            }
+        }
+        return float.childNodes.length ? [float] : [document.createDocumentFragment()];
+    }
+
+    private async _renderDialog(node: VNode): Promise<Element> {
+        const clone = floatItem.cloneNode(true) as Element;
```
dmo-odoo, line 28: It might be good to remove the confusing "float" from the code. This `floatItem` *is* a `dialog` now.

packages/plugin-dialog/src/ui/DialogZoneDomRenderer.ts
```diff
@@ -0,0 +1,43 @@
...
+const floatItem = container.firstElementChild;
+
+export class DialogZoneDomRenderer extends AbstractRenderer<Node[]> {
+    static id = 'dom';
+    engine: DomRenderingEngine;
+    predicate = DialogZoneNode;
+
+    async render(node: DialogZoneNode): Promise<Node[]> {
+        const float = document.createElement('jw-dialog-container');
+        for (const child of node.children) {
```
dmo-odoo, line 19: This is not rebased on `master`.

packages/plugin-dialog/src/ui/DialogZoneDomRenderer.ts
```diff
@@ -0,0 +1,43 @@
...
+        const content = clone.querySelector('jw-content');
+        for (const domNode of await this.engine.render(node)) {
+            content.appendChild(domNode);
+        }
+
+        clone.addEventListener('click', (ev: MouseEvent): void => {
+            const target = ev.target as Element;
+            if (target.classList.contains('close')) {
+                this.engine.editor.execCommand('UI-hide', { ref: node });
+            }
```
dmo-odoo, line 38: Shouldn't this be on each element which has `close` in its class rather than having to check it for every click ? And what about unbinding ? Do we sometimes want to unbind them ?

packages/plugin-dialog/src/ui/DialogZoneDomRenderer.ts
Gorash: No, else the parent add à <br> we don't want this

packages/plugin-dialog/src/ui/DialogZoneDomRenderer.ts
Gorash: Nop. We can use xml parser but it's longer

packages/plugin-dialog/src/ui/DialogZoneDomRenderer.ts
Gorash: no need to unbind, because we remove the rendering node

PR #154, opened 2020-04-10
shortcuts devtools poc

Goaman:


packages/plugin-devtools/src/components/CommandsComponent.ts
```diff
@@ -3,29 +3,75 @@ import { CommandIdentifier } from '../../../core/src/Dispatcher';
...
+                this.state.shortcuts.user[mappings.configuredCommand.commandId].push(
+                    this._representShortcutPattern(mappings.pattern),
+                );
+            }
+        }
+    }
+    _representShortcutPattern(pattern: ShortcutPattern) {
+        const key = isKeyPattern(pattern) ? pattern.key : '<' + pattern.code + '>';
+        return [[...pattern.modifiers].join(' '), key].join(' ');
+    }
```
dmo-odoo, line 74: Wouldn't it be way easier to just do something like that ? I'm not saying it works, it's just for the idea.
```typescript
const defaultShortcuts = this.editor.plugins.values().map(plugin => plugin.loadables.shortcuts).flat();
const shortcuts = [this.editor.coniguration.shortcuts, ...defaultShortcuts];
```
This way you already have it in the "pure" form of "a string then a command" and you just have to use the command name to "group" them.

dmo-odoo:
I'm closing this since a Shortcuts was introduced in the DevTools with https://github.com/odoo-dev/jabberwock/pull/160.

PR #157, opened 2020-04-27
Master fix selection without root nby

Goaman:


Goaman:
I need something like this for the integration

dmo-odoo:
@Goaman Does this still apply with the new layout ?

dmo-odoo:
@Goaman ping

Goaman:
Indeed, the problem has disappeared in master.
But the test might still be relevant.

dmo-odoo:
Then if the PR is modified to only contain the test then I could merge it.

Goaman:
done

PR #159, opened 2020-05-05
[IMP] change id reference for parser and renderer

Goaman:


Goaman:
There is still some reference left unchanged in defaultRenderer/Parser because there is a cyclic import. The default class import the engine AND the engine import the default class. Which make the compiled code blow up at runtime.

PR #160, opened 2020-05-08
[ADD] UI: more UI related to past features

Zinston:


packages/plugin-input/src/InputNode.ts
```diff
@@ -0,0 +1,40 @@
...
+     */
+    get inputType(): string {
+        return this.attributes.type as string;
+    }
+    /**
+     * Set the `type` attribute of this Input node.
+     */
+    set inputType(name: string) {
+        this.attributes.type = name;
+    }
```
dmo-odoo, line 39: Attributes are a modifier that is specific to XML. If these values make sense for this node outside of the context of XML (I think it does) then I think they deserve a proper field on the node.

packages/core/src/VNodes/AbstractNode.ts
```diff
@@ -36,6 +36,12 @@ export abstract class AbstractNode {
     toString(): string {
         return this.name;
     }
+    /**
+     * Return the text content of this node.
+     */
+    text(): string {
```
dmo-odoo, line 42: 😢 Why do you need this ?

packages/plugin-char/src/Char.ts
```diff
@@ -48,6 +48,9 @@ export class Char<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
         const text = params.text;
         const inline = this.editor.plugins.get(Inline);
         const formats = inline.getCurrentFormats(range);
+        if (params.formats) {
+            formats.append(...params.formats);
+        }
```
dmo-odoo, line 53: Looks like a bugfix to me.

packages/plugin-dom-layout/test/DomLayout.test.ts
```diff
@@ -549,7 +549,6 @@ describe('DomLayout', () => {
                             '</jw-test>' +
                             '</jw-body>' +
                             '<jw-footer></jw-footer>' +
-                            '<t-dialog></t-dialog>' +
```
dmo-odoo, line 552: Why ? Just wondering.

packages/plugin-link/src/Link.ts
```diff
@@ -36,57 +41,51 @@ export class Link<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
...
     // Public
     //--------------------------------------------------------------------------
 
-    link(params: LinkParams): void {
-        const range = params.context.range;
-        const selectedInlines = range.selectedNodes(InlineNode);
+    async link(params: LinkParams): Promise<void> {
+        // If the url is undefined, ask the user to provide one.
+        if (!params.url) {
+            const layoutPlugin = this.editor.plugins.get(Layout);
```
dmo-odoo, line 76: Same as `KeyMap`. There is no difference between the `Layout` *plugin* and *the* layout. The `Layout` plugin *is* the layout. So I would just call this variable `Layout`.

packages/plugin-link/src/Link.ts
```diff
@@ -36,57 +41,51 @@ export class Link<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
...
+            await layoutPlugin.remove('link');
```
dmo-odoo, line 77: Why do you have to `remove` it preventively ?

packages/plugin-link/src/Link.ts
```diff
@@ -36,57 +41,51 @@ export class Link<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
...
+            await layoutPlugin.add('link', 'float');
```
dmo-odoo, line 78: Normally, if the layout is correctly configured, you should be able to add `link` without specifying a zone. It should then put it in the magic `default` zone which should be `float` in the default configuration. The point is that, in Odoo, we might want to configure the editor such that the `default` zone is in the sidebar for example.

packages/plugin-link/src/components/LinkComponent.ts
```diff
@@ -0,0 +1,49 @@
...
+import { LinkParams } from '../Link';
+import { Layout } from '../../../plugin-layout/src/Layout';
+import { useState } from '@odoo/owl';
+
+export class LinkComponent<T = {}> extends OwlComponent<T> {
+    static components = {};
+    static template = 'link';
+    state = useState({
+        url: this._currentUrl(),
+        label: this._currentLabel(),
```
dmo-odoo, line 13: It's weird. Those are functions but it have no verbs. Shouldn't they be named something like `_getCurrentUrl` instead ?

packages/plugin-link/src/components/LinkComponent.ts
```diff
@@ -0,0 +1,49 @@
...
+    //--------------------------------------------------------------------------
+    // Public
+    //--------------------------------------------------------------------------
+
+    async saveLink(): Promise<void> {
+        await this.env.editor.execCommand('link', {
+            url: this.state.url,
+            label: this.state.label,
+        } as LinkParams);
+        await this.env.editor.plugins.get(Layout).remove('link');
```
dmo-odoo, line 25: You're removing `link` in the plugin already and you're doing it here as well ? At least one of them should disappear. Honestly, I think both should disappear as calling `destroy` should remove it IMO. cc @Gorash

packages/plugin-keymap/src/Keymap.ts
```diff
@@ -211,6 +211,19 @@ export class Keymap<T extends KeymapConfig = KeymapConfig> extends JWPlugin<T> {
...
+     * command ID.
+     */
+    getShortcuts(): ConfiguredCommand[] {
+        return this.mappings[0]
+            .map(shortcut => shortcut.configuredCommand)
+            .sort((a, b) => {
+                if (a.commandId < b.commandId) return -1;
+                if (a.commandId > b.commandId) return 1;
+                return 0;
+            });
```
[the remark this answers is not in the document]
Zinston: `ConfiguredCommand` contains `commandId` and `pattern`.

packages/plugin-input/src/InputNode.ts
Zinston: I did the same here as on `LinkNode` with `url` though. Should I change both then?

packages/core/src/VNodes/AbstractNode.ts
Zinston: In the DOM, `Node` has `textContent` which does the same thing so I don't find this shocking.
I need it because a user might not use `CharNode` and there are enough use cases for finding out what the text content of a node is.

packages/plugin-char/src/Char.ts
Zinston: It is indeed.

packages/plugin-dom-layout/test/DomLayout.test.ts
Zinston: @Gorash says it's because I added the `Dialog` plugin to `BasicEditor` and that somehow has the effect of not including `<t-dialog>` if there is none to display. He says it's normal. I didn't search further.

packages/plugin-link/src/Link.ts
Zinston: So we don't end up with two instances of it. This line will disappear once @Gorash makes `remove` the default instead of `hide`.

packages/plugin-link/src/Link.ts
Zinston: It is an instance though right? So `layout` is more appropriate.

packages/plugin-link/src/Link.ts
Zinston: Aaaaaaaah _that_'s what `componentZones` does!

packages/plugin-keymap/src/Keymap.ts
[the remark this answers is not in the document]
dmo-odoo: In master:
```typescript
export interface ConfiguredCommand extends Contextual {
    /**
     * An undefined `commandId` effectively unbinds the shortcut.
     */
    readonly commandId: CommandIdentifier | undefined;
    readonly commandArgs?: {};
}
export interface Contextual {
    context?: Context;
    selector?: Predicate[];
    // If the selector match and a `check` callback has been provided, call
    // it with the current context to perform an additional checking.
    check?: (context: CheckingContext) => boolean;
}
interface Mapping {
    pattern: ShortcutPattern;
    configuredCommand?: ConfiguredCommand;
}
```

I see pattern on `Mapping`, not on `ConfiguredCommand`. Am I missing something ?

packages/plugin-input/src/InputNode.ts
dmo-odoo: Well work on the modifiers changed the semantics of the attributes, so yes I suggest changing `LinkFormat` as well

packages/core/src/VNodes/AbstractNode.ts
dmo-odoo: Well, what you call "enough uses cases", I call it `LinkComponent` line 46 and that's all... ;-) Maybe keep the name `textContent` at least ? Though it would also be a verb-less function. Maybe a getter if we want to stay closer to the dom ?

packages/plugin-char/src/Char.ts
dmo-odoo: Not really, just wanted to point it out in case I missed something.

packages/plugin-link/src/Link.ts
dmo-odoo: Pff, amateurs... 🙄 🚎

packages/plugin-devtools/src/components/CommandsComponent.ts
```diff
@@ -97,4 +98,27 @@ export class CommandsComponent extends OwlComponent<CommandsProps> {
...
+     * Return the configured commands matching the given command identifier.
+     *
+     * @param commandIdentifier
+     */
+    matchingShortcuts(commandIdentifier: string): ConfiguredCommand[] {
+        return this.env.editor.plugins
+            .get(Keymap)
+            .mappings.flat()
+            .map(shortcut => shortcut.configuredCommand)
+            .filter(shortcut => shortcut.commandId === commandIdentifier);
```
dmo-odoo, line 113: I don't like the "chained transformation on each line" pattern. It's in the devtools so I don't really care, but please don't use it in actual editor code.

For the sake of comparison, my original suggestion was (modified to return `ConfiguredCommand` instead of `Mapping`):
```typescript
const shortcuts = this.editor.plugins.get(KeyMap).mappings.flat();
const commands = shortcuts.map(shortcut => shortcut.configuredCommand);
return commands.filter(command => command.commandId === commandIdentifier);
```

Not sure which one is neater though. Yours looks simpler and more chaotic at the same time, it's very weird. Maybe I'm just not used to see this pattern ? What do you think ?

Zinston: I think not using the chaining pattern allows you to name variables which makes things clearer, but that at the same time it makes the code bulky, which can make it somewhat harder to read. So for me it's on a case by case basis. And I figured in this case that I liked it better like this because the names I could come up with didn't provide much added value.

packages/plugin-input/src/InputNode.ts
Zinston: I should probably wait for `modifiers` though then...

packages/core/src/VNodes/AbstractNode.ts
Zinston: We used to have a `text` field in the description of the `VNode` in DevTools, which I find useful. In many tests we import `CharNode` just so we can do `(node as CharNode).char`... There's plenty of cases in the DOM where I want to use textContent and I don't see why it would be different here.

packages/plugin-input/src/InputNode.ts
dmo-odoo: Alright, as long as you don't forget that you should not just change it to `modifiers.get(Attributes)` at that point and to change `LinkFormat`.

packages/core/src/VNodes/AbstractNode.ts
dmo-odoo: Ok then a `textContent` getter and you update the tests accordingly such that the added feature looks useful ? :)

Issue #147, opened 2020-03-30
Improve Pre


dmo-odoo:
The parsing and rendering of `Pre` could be improved, see comments on https://github.com/odoo-dev/jabberwock/pull/142/files.

Zinston:
Resolved by #162

Zinston:
Resolved by #162.

PR #162, opened 2020-05-11
Improve Pre

Zinston:
This PR addresses issue https://github.com/odoo-dev/jabberwock/issues/147

### DIFF
--- a/packages/plugin-pre/src/PreSeparatorHtmlDomRenderer.ts
+++ b/packages/plugin-pre/src/PreSeparatorHtmlDomRenderer.ts
@@ -1,0 +1,23 @@
+import { PreNode } from './PreNode';
+import { VNode } from '../../core/src/VNodes/VNode';
+import { AbstractRenderer } from '../../plugin-renderer/src/AbstractRenderer';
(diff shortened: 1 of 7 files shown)


packages/plugin-pre/src/PreSeparatorHtmlDomRenderer.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     * Render the VNode.
+     */
+    async render(node: VNode): Promise<Node[]> {
+        const separators = await this.super.render(node);
+        const rendering: Node[] = [];
+        for (let separatorIndex = 0; separatorIndex < separators.length; separatorIndex += 1) {
+            rendering.push(document.createTextNode('\n'));
+        }
```
dmo-odoo, line 23: I'm not sure I understand. Isn't that the equivalent of:
```typescript
const rendering = separators.map(s => document.creatTextNode('\n'));
```

Zinston: It is but I figured you wouldn't like that because it would be using `map` for the side effect of rendering the nodes. :man_shrugging:

PR #168, opened 2020-05-13
[IMP] OdooVideo: parse/render odoo video

Goaman:
In Odoo, video information are saved within a div that has
the class "media_iframe_video" with an attribute "data-oe-expression".

### DIFF
--- a/examples/odoo/index.html
+++ b/examples/odoo/index.html
@@ -1,0 +1,12 @@
+<html>
+    <head>
+        <title>Jabberwock Editor</title>
+        <link rel="shortcut icon" href="#" />
+    </head>
+    <body>
+        <section id="wrapper"></section>
+        <!-- The following file is created by webpack in the config file -->
+        <!-- The path generated is "/<filename>.js" -->
+        <script type="text/javascript" src="odoo.js" charset="UTF-8"></script>
+    </body>
+</html>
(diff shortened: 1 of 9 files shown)


packages/plugin-video/src/OdooVideo.ts
```diff
@@ -0,0 +1,13 @@
+import { JWPlugin, JWPluginConfig } from '../../core/src/JWPlugin';
+import { OdooVideoXmlDomParser } from './OdooVideoXmlDomParser';
+import { OdooVideoHtmlDomRenderer } from './OdooVideoHtmlDomRenderer';
+import { Loadables } from '../../core/src/JWEditor';
+import { Parser } from '../../plugin-parser/src/Parser';
+import { Renderer } from '../../plugin-renderer/src/Renderer';
+
+export class OdooVideo<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T> {
```
dmo-odoo, line 8: I think the folder should be named `plugin-odoo-video`.

packages/plugin-video/test/OdooVideo.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    constructor() {
+        super();
+        this.load(OdooVideo);
+    }
+}
+
+describePlugin(OdooVideo, () => {
+    describe('parser/parser', () => {
+        it('should parse and render an Odoo video', async () => {
+            await testEditor(TestVideoEditor, {
```
dmo-odoo, line 15: Since you are using `describePlugin`, you should not need to create a new editor just to load a custom plugin. Replacing `TestVideoEditor` here by `BasicEditor` should work. If it doesn't then let me know because there would definitely be a bug somewhere.

PR #155, opened 2020-04-20
add checklist feature

Gorash:

PR #169, opened 2020-05-14
[ADD] List: introduce the checklist feature

Zinston:
Refactor of #155.

### DIFF
--- a/packages/plugin-list/src/ListItemHtmlDomRenderer.ts
+++ b/packages/plugin-list/src/ListItemHtmlDomRenderer.ts
@@ -31,7 +31,7 @@ export class ListItemHtmlDomRenderer extends AbstractRenderer<Node[]> {
                 if (ev.offsetX < 0) {
                     ev.stopImmediatePropagation();
                     ev.preventDefault();
-                    withRange(VRange.at(node.firstChild()), async range => {
+                    withRange(VRange.at(node.firstChild() || node), async range => {
                         await this.engine.editor.execCommand('toggleChecked', {
                             context: {
                                 range: range,
--- a/packages/plugin-list/src/ListItemHtmlDomRenderer.ts
+++ b/packages/plugin-list/src/ListItemHtmlDomRenderer.ts
@@ -5,6 +5,7 @@ import { VElement } from '../../core/src/VNodes/VElement';
 import { List } from './List';
 import { HtmlDomRenderingEngine } from '../../plugin-html/src/HtmlDomRenderingEngine';
 import { ListItemAttributes } from './ListItemXmlDomParser';
+import { VRange, withRange } from '../../core/src/VRange';
 
 export class ListItemHtmlDomRenderer extends AbstractRenderer<Node[]> {
     static id = HtmlDomRenderingEngine.id;
@@ -13,10 +14,32 @@ export class ListItemHtmlDomRenderer extends AbstractRenderer<Node[]> {
 
     async render(node: VNode): Promise<Node[]> {
         const domListItem = document.createElement('li');
+
+        // Render the node's attributes that were stored on the technical key
+        // that specifies those attributes belong on the list item.
+        this.engine.renderAttributes(ListItemAttributes, node, domListItem);
+
         if (node.is(ListNode)) {
             if (!domListItem.style.listStyle) {
                 domListItem.style.listStyle = 'none';
             }
+        } else if (ListNode.CHECKLIST(node.parent)) {
+            domListItem.classList.add(ListNode.isChecked(node) ? 'checked' : 'unchecked');
+
+            // Handle click in the checkbox.
+            domListItem.addEventListener('mousedown', async (ev: MouseEvent) => {
+                if (ev.offsetX < 0) {
+                    ev.stopImmediatePropagation();
+                    ev.preventDefault();
+                    withRange(VRange.at(node.firstChild()), async range => {
+                        await this.engine.editor.execCommand('toggleChecked', {
+                            context: {
+                                range: range,
+                            },
+                        });
+                    });
+                }
+            });
         }
 
         // Direct ListNode's VElement children "P" are rendered as "LI"
@@ -29,9 +52,6 @@ export class ListItemHtmlDomRenderer extends AbstractRenderer<Node[]> {
             const renderedNode = await this.super.render(node);
             domListItem.append(...renderedNode);
         }
-        // Render the node's attributes that were stored on the technical key
-        // that specifies those attributes belong on the list item.
-        this.engine.renderAttributes(ListItemAttributes, node, domListItem);
         return [domListItem];
     }
 }
--- a/packages/plugin-list/src/ListNode.ts
+++ b/packages/plugin-list/src/ListNode.ts
@@ -1,10 +1,14 @@
 import { VNode } from '../../core/src/VNodes/VNode';
 import { ContainerNode } from '../../core/src/VNodes/ContainerNode';
+import { Modifier } from '../../core/src/Modifier';
 
 export enum ListType {
     ORDERED = 'ORDERED',
     UNORDERED = 'UNORDERED',
+    CHECKLIST = 'CHECKLIST',
 }
+
+export class IsChecked extends Modifier {}
 
 export class ListNode extends ContainerNode {
     // Typescript currently doesn't support using enum as keys in interfaces.
@@ -13,7 +17,10 @@ export class ListNode extends ContainerNode {
         return node && node.is(ListNode) && node.listType === ListType.ORDERED;
     }
     static UNORDERED(node: VNode): node is ListNode {
-        return node && node.is(ListNode) && node.listType == ListType.UNORDERED;
+        return node && node.is(ListNode) && node.listType === ListType.UNORDERED;
+    }
+    static CHECKLIST(node: VNode): node is ListNode {
+        return node && node.is(ListNode) && node.listType === ListType.CHECKLIST;
     }
     listType: ListType;
     constructor(params: { listType: ListType }) {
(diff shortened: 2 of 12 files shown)


packages/plugin-list/src/ListItemHtmlDomRenderer.ts
```diff
@@ -12,10 +14,39 @@ export class ListItemHtmlDomRenderer extends AbstractRenderer<Node[]> {
...
+
         if (node.is(ListNode)) {
             if (!domListItem.style.listStyle) {
                 domListItem.style.listStyle = 'none';
             }
+        } else if (ListNode.CHECKLIST(node.parent)) {
+            domListItem.classList.add(ListNode.isChecked(node) ? 'checked' : 'unchecked');
+
+            // Handle click in the checkbox.
+            domListItem.addEventListener('mousedown', async (ev: MouseEvent) => {
```
Goaman, line 33: Why "mousedown" ? Isn't a click more appropriate?

packages/plugin-list/src/ListXmlDomParser.ts
```diff
@@ -15,17 +15,52 @@ export class ListXmlDomParser extends AbstractParser<Node> {
...
+        list.append(...children);
+        list.attributes = this.engine.parseAttributes(item);
+
+        // In the case of a checklist, parse their checked/unchecked status and
+        // ensure vertical propagatation.
+        if (type === ListType.CHECKLIST) {
+            const reChecked = /(^| )checked( |$)/i;
+            for (const child of children) {
+                const liAttributes = child.attributes['li-attributes'] as Record<string, string>;
+                if (!liAttributes?.class) continue;
```
dmo-odoo, line 43: I *hate* `continue`s and `break`s when they can be written as proper `filter` or `while` conditions because the latter is always more idiomatic than the former.

In this case:
```typescript
const items = children.filter(child => child.modifiers.find(LiAttributes)?.class);
``` 
If you can somehow remove the `const next = child.nextSibling();` below then you could even iterate over the `LiAttributes` rather than the nodes themselves. (Especially considering I don't understand the case for which this part of the code exist, I don't understand why the last checked item must be treated differently.)

packages/plugin-list/src/ListXmlDomParser.ts
```diff
@@ -15,17 +15,52 @@ export class ListXmlDomParser extends AbstractParser<Node> {
...
+                if (liAttributes.class.match(reChecked) && !ListNode.CHECKLIST(next)) {
+                    // Don't (un-)check the children as they already are
+                    // properly checked/unchecked and propagate information.
+                    list.check(child);
+                }
+
+                // Remove the checklist-related classes from `liAttributes`.
+                const clearClass = liAttributes.class
+                    .replace(/(^| )(checklist|(un)?checked)( |$)/gi, ' ')
+                    .trim();
```
dmo-odoo, line 56: This part will be better when `ClassList` is merged.

packages/plugin-list/src/ListXmlDomParser.ts
Zinston: I need `classList` but I also need `child` so I can check it.
Therefore using filter means I'm going to have to go find the attributes for each child twice:
```ts
const items = children.filter(child => child.modifiers.find(ListItemAttributes));
for (const item of items) {
    const liAttributes = item.modifiers.find(ListItemAttributes);

    // Parse the list item's checked status.
    if (liAttributes.classList.has('checked')) {
        ListNode.check(item);
    }

    // Remove the checklist-related classes from `liAttributes`.
    liAttributes.classList.remove('checklist checked unchecked');
}
```
I don't think we want that. Right now I'm putting an `if` instead.

packages/plugin-list/src/ListNode.ts
```diff
@@ -1,14 +1,16 @@
...
-import { AtomicNode } from '../../core/src/VNodes/AtomicNode';
+import { Modifier } from '../../core/src/Modifier';
 
 export enum ListType {
     ORDERED = 'ORDERED',
     UNORDERED = 'UNORDERED',
     CHECKLIST = 'CHECKLIST',
 }
 
+// Todo: make into regular boolean
```
dmo-odoo, line 11: ?

packages/plugin-list/test/List.test.ts
```diff
@@ -111,7 +111,7 @@ describePlugin(List, testEditor => {
                 '        <li>b.4</li>', // li3_3
                 '    </ol>',
                 '    <ul class="checklist">', // li4
-                '        <li class="unchecked">c.1</li>', // li4_0
+                '        <li>c.1</li>', // li4_0
```
dmo-odoo, line 114: I thought chm went back on his "you can delete it" statement regarding `unchecked` ?

packages/plugin-list/test/Checklist.test.ts
```diff
@@ -223,7 +221,8 @@ describePlugin(List, testEditor => {
             });
             it('should parse a complet checklist', async () => {
                 await testEditor(BasicEditor, {
-                    contentBefore: unformat(`
+                    contentBefore: unformat(
+                        `
```
dmo-odoo, line 225: If you care about that kind of things then why not `'<ul` rather than using an entire line for `'` ? I mean, the variable ends with `</ul>'` already so...

packages/plugin-list/test/Checklist.test.ts
```diff
@@ -297,12 +300,14 @@ describePlugin(List, testEditor => {
                                     <li class="unchecked">3.2</li>
                                 </ul>
                             </li>
-                            <li class="unchecked">4</li>
+                            <li class="checked">4</li>
```
dmo-odoo, line 303: So... the test used to be wrong then ? The comment in `contentBefore` uses the premise that "4 is checked" yet it used to be unchecked ?

packages/plugin-list/test/Checklist.test.ts
```diff
@@ -1269,16 +1279,21 @@ describePlugin(List, testEditor => {
...
                                 '<ul class="checklist"><li class="unchecked">ab[]gh</li></ul>',
                         });
                         await testEditor(BasicEditor, {
                             contentBefore:
                                 '<ul class="checklist"><li>ab]cd</li><li style="list-style: none;"><ul class="checklist"><li class="checked">ef[gh</li></ul></li></ul>',
                             stepFunction: deleteForward,
+                            // The indented list's parent gets rendered as
+                            // checked because its only child is checked. When
+                            // we remove that child, the checklist gets
+                            // unchecked because it becomes independant again.
```
dmo-odoo, line 1294: It's great that you talk about "becoming independant again" as we will be able to grep that to adapt the tests in case fp disagrees.

packages/plugin-list/test/Checklist.test.ts
```diff
@@ -2386,8 +2425,13 @@ describePlugin(List, testEditor => {
                             contentBefore:
                                 '<ul class="checklist"><li class="checked">ab[cd</li><li style="list-style: none;"><ul class="checklist"><li class="unchecked">ef]gh</li></ul></li></ul>',
                             stepFunction: deleteBackward,
+                            // The indented list cannot be unchecked while its
+                            // parent is checked: it gets checked automatically
+                            // as a result. So "efgh" gets rendered as checked.
+                            // Given that the parent list item was explicitely
+                            // set as "checked", that status is preserved.
```
dmo-odoo, line 2432: Well this test was wrong to begin with since the initial situation does not make sense. Nothing to do here, it's a good test to check our "resolution" of this issue, but the original test by chm was wrong IMO.

packages/plugin-list/test/Checklist.test.ts
```diff
@@ -2594,16 +2679,16 @@ describePlugin(List, testEditor => {
                                         '<ul class="checklist"><li class="checked">abc</li></ul><p>[]<br></p>',
                                 });
                             });
-                            it('should add an empty list item at the end of an indented list, then remove it', async () => {
+                            it('should add an empty list item at the end of an indented list, then outdent it', async () => {
```
dmo-odoo, line 2682: Here you changed the test because of `preserve`, right ? Yet you did not duplicate+skip like the others. Any reason ? Just wondering.

packages/plugin-list/test/Checklist.test.ts
```diff
@@ -2624,61 +2709,155 @@ describePlugin(List, testEditor => {
...
-                                        '<ul class="checklist a"><li class="checked">abc[]</li></ul>',
-                                    stepFunction: async (editor: JWEditor) => {
-                                        await insertParagraphBreak(editor);
-                                        await insertText(editor, 'b');
-                                        await insertParagraphBreak(editor);
-                                    },
-                                    contentAfter:
-                                        '<ul class="checklist a"><li class="checked">abc</li><li class="unchecked">b</li><li class="unchecked">[]<br></li></ul>',
+                            describe('after unchecked item', () => {
+                                // TODO: do not clone the `IsChecked` modifier
```
dmo-odoo, line 2713: I thought this TODO would only be present in the skipped tests, not in the duplicated ones ? Did I miss something ?

packages/plugin-list/test/Checklist.test.ts
```diff
@@ -2690,23 +2869,24 @@ describePlugin(List, testEditor => {
...
                                         </ul>`),
-                                    stepFunction: async (editor: JWEditor) => {
-                                        await insertParagraphBreak(editor);
-                                        await insertText(editor, 'b');
-                                        await insertParagraphBreak(editor);
-                                    },
-                                    contentAfter: unformat(`
+                                        stepFunction: async (editor: JWEditor) => {
+                                            await insertParagraphBreak(editor);
+                                            await insertText(editor, '0');
```
dmo-odoo, line 2874: Why change the inserted text ? Just wondering.

packages/plugin-list/test/Checklist.test.ts
```diff
@@ -2624,61 +2709,155 @@ describePlugin(List, testEditor => {
...
-                                    contentBefore:
-                                        '<ul class="checklist"><li class="a checked">abc[]</li></ul>',
-                                    stepFunction: async (editor: JWEditor) => {
-                                        await insertParagraphBreak(editor);
-                                        await insertText(editor, 'b');
-                                        await insertParagraphBreak(editor);
-                                    },
-                                    contentAfter:
-                                        '<ul class="checklist"><li class="a checked">abc</li><li class="a unchecked">b</li><li class="a unchecked">[]<br></li></ul>',
+                                it('should add two list items with a class at the end of a checklist', async () => {
```
dmo-odoo, line 2729: I'm trusting you with the duplicated tests.

packages/plugin-list/test/Checklist.test.ts
```diff
@@ -3042,14 +3222,14 @@ describePlugin(List, testEditor => {
                             </ul>`),
                     });
                 });
-                it('should outdent a checklist and previous ligne as "title"', async () => {
+                it('should outdent a checklist and previous line as "title"', async () => {
```
dmo-odoo, line 3225: 🇫🇷

packages/plugin-list/test/Checklist.test.ts
```diff
@@ -3081,14 +3261,7 @@ describePlugin(List, testEditor => {
...
         });
     });
-    describe('toggleChecked', () => {
-        it('should throw if we try to check an outside item', async () => {
-            const checklist = new ListNode({ listType: ListType.CHECKLIST });
-            const node = new CharNode({ char: 'a' });
-            expect(() => {
-                checklist.check(node);
-            }).to.throw();
-        });
```
dmo-odoo: 🙄

Zinston: As far as stupid tests go... A relic from when `ListNode` had a list of its checked children... Then `check` was called on the list to check one of its children :open_mouth: So this test checked that calling `check` on a list to check a node that wasn't one of its children made the whole editor crash. Because that would be best :shrug:

PR #180, opened 2020-05-20
[IMP] Core: add editor mode

Goaman:


Goaman:
To discuss:
- naming of "mode" need to be discussed
- the class ContextManager could be reorganised/splitted
- performance: the check might be expensive
- there is confusion between the attribute `editable` of the node and the
  editablity it's children.

PR #190, opened 2020-05-28
Master fix link format nby

Goaman:


dmo-odoo:
cc @Zinston

packages/plugin-link/src/LinkFormat.ts
```diff
@@ -17,4 +17,13 @@ export class LinkFormat extends Format {
         element.setAttribute('href', this.url);
         return element;
     }
+
+    /**
+     * @override
+     */
+    clone(): this {
+        const clone = super.clone();
+        clone.url = this.url;
```
Zinston, line 26: Ah oui de fait, avant `url` était un getter vers l'attribut donc quand j'ai changé les attributs ça a peté le clone, bien vu !

packages/core/src/Modifiers.ts
```diff
@@ -215,6 +221,20 @@ export class Modifiers {
...
+    replaceOrAppend(modifiers: Modifiers): void {
+        for (const modifier of modifiers.getContent()) {
+            const foundModifier = this.find(modifier.constructor);
+            if (foundModifier) {
+                this.replace(foundModifier, modifier);
+            } else {
+                this.append(modifier.clone());
+            }
+        }
+    }
```
Zinston, line 237: This is useless: `replace` does exactly this already...

packages/core/src/Modifiers.ts
```diff
@@ -121,6 +121,12 @@ export class Modifiers {
         }
         return found;
     }
+    /**
+     * Return all the modifiers instances that are hold by `this`.
+     */
+    getContent(): Modifier[] {
+        return this._contents;
+    }
```
Zinston, line 129: No! The whole point of not using an array is to not manipulate `_contents` directly! If we wanted to be able to get it directly we wouldn't have made it private...

dmo-odoo:
Les deux premiers commits sont ok, pour le dernier je suis 100% d'accord avec Antoine.

packages/plugin-char/src/Char.ts
```diff
@@ -49,7 +49,7 @@ export class Char<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
         const inline = this.editor.plugins.get(Inline);
         const modifiers = inline.getCurrentModifiers(range);
         if (params.formats) {
-            modifiers.append(...params.formats.map(format => format.clone()));
+            modifiers.replaceOrAppend(params.formats);
```
Zinston, line 52:
```ts
for (const format of params.formats) {
    modifiers.replace(format.constructor, format.clone());
}
```

packages/core/test/modifiers.test.ts
```diff
@@ -22,5 +22,20 @@ describe('core', () => {
                 expect(modifiersMap[3]).to.equal(m4);
             });
         });
+        describe('set()', () => {
+            it('should replace one modifier and append the other', () => {
```
Zinston, line 26: hehe thanks :-)

dmo-odoo:
@Goaman I could not reproduce the bug you were trying to fix in your last commit. In order to merge the PR, I edited the last commit. The end result should be the same so the bug should be fixed, but I can't be sure since I was not able to reproduce it in the first place. Please let me know if the bug still exists.

PR #191, opened 2020-05-28
Add doc & rename & add type information

Goaman:


Goaman:
Note: We should add more doc pages. We actually have none.

dmo-odoo:
ok [IMP] Link: add type information
ko [DOC] Modifiers => [DOC] Core: add Modifiers doc
ko [RENAME] InsertTextParams formats -> modifiers => [REN] Char: InsertTextParams formats -> modifiers

documentation/core/Modifiers.md
```diff
@@ -0,0 +1,20 @@
...
+When parsing and rendering a document, we need to save the attributes that a
+DOM node has and save it in our VNode.
+
+Attributes exists to:
+- retrieve HTML attributes that a DOM node has originally
+- render the HTML attributes that a VNode has
+
+## Format
+The DOM inline nodes (e.g. span, strong, em, b, a, ...) in HTML are difficult to
+manipulate. We therefore we transform them into a Format modifiers that we
```
Zinston, line 16: - "Therefore we" or "We therefore" but not "We therefore we" ^^
- "into `Format` modifiers"

documentation/core/Modifiers.md
```diff
@@ -0,0 +1,20 @@
...
+## Attributes
+When parsing and rendering a document, we need to save the attributes that a
+DOM node has and save it in our VNode.
+
+Attributes exists to:
+- retrieve HTML attributes that a DOM node has originally
+- render the HTML attributes that a VNode has
+
+## Format
+The DOM inline nodes (e.g. span, strong, em, b, a, ...) in HTML are difficult to
```
Zinston, line 15: The existence of a dichotomy between blocks and inline nodes is what causes a level of complexity that we managed to eliminate by introducing `Format`. It's not so much that the inline nodes themselves are hard to manipulate but the fact that there is that dichotomy means that we have to constantly be on our guard and treat different nodes differently, which can end up being complicated.

documentation/core/Modifiers.md
```diff
@@ -0,0 +1,20 @@
...
+manipulate. We therefore we transform them into a Format modifiers that we
+attach to a VNode through the "modifiers" properties.
```
Zinston, line 17: Everywhere in the DOC please don't forget to use \`\` for the references to classes etc.
In this case: `VNode` instead of VNode.

documentation/core/Modifiers.md
```diff
@@ -0,0 +1,20 @@
...
+
+Format themselve have modifiers. For instance, a HTML strong tag migh have
```
Zinston, line 19: "`Format`s may themselves have modifiers. For instance a bold text may be colored red, which would translate into HTML as a strong tag with a style attribute "color: red", and into Jabberwock as (with the help of `Char`, `Attributes` and `Inline` plugins) a group of `Char` nodes with as modifiers: (1) an extension of the `Format` class that renders as a `strong` tag and (2) `Attributes` containing the "color" style set to red."
Too much? :'D

packages/core/src/Modifier.ts
```diff
@@ -8,9 +8,15 @@ export interface Modifier {
     constructor: ModifierConstructor & this;
 }
 export class Modifier {
+    /**
+     * The name of the this Modifier.
```
Zinston, line 12: The this? :-)
"Return the name of this modifier."

packages/core/src/Modifier.ts
```diff
@@ -8,9 +8,15 @@ export interface Modifier {
...
+     */
     get name(): string {
         return '';
     }
+    /**
+     * Return the string representation of this Modifier.
```
Zinston, line 18: Why the capital letter?

packages/core/src/Modifier.ts
```diff
@@ -22,9 +28,17 @@ export class Modifier {
     applyTo(node: VNode): void {
         node.modifiers.prepend(this);
     }
+    /**
+     * Compare two modifiers.
+     *
+     * Meant to be overridden if necessary.
```
Zinston, line 34: "Return true if this modifier is the same as the given modifier, false otherwise.

This can be overridden to change the meaning of equality between modifiers."

packages/core/src/Modifier.ts
```diff
@@ -22,9 +28,17 @@ export class Modifier {
...
+     */
     isSameAs(otherModifier: Modifier): boolean {
         return this === otherModifier;
     }
+    /**
+     * Clone this modifier.
```
Zinston, line 40: "Return a clone of this modifier."

packages/core/src/Modifiers.ts
```diff
@@ -1,6 +1,12 @@
 import { Modifier } from './Modifier';
 import { Constructor, isConstructor } from '../../utils/src/utils';
 
+/**
+ * When parsing dom elements, modifiers can be DOM inline nodes (a, b, strong).
```
Zinston, line 5: Don't write this here.
1. No mention of the DOM in `core`
2. No mention of a plugin in `core` (`Format` in this case)
3. Start your description of the class by talking about the general case, the general purpose, not the edge cases.
4. Be consistent: you write "dom" and "DOM" in the same sentence.

packages/core/src/Modifiers.ts
```diff
@@ -1,6 +1,12 @@
...
+ *
+ * The order of modifiers are important because it impact the way they will
```
Zinston, line 7: "The order" -> "is"
"It" -> "impacts"

=> "The order of modifiers is important because it may impact their rendering." ?

packages/plugin-inline/src/Format.ts
```diff
@@ -2,16 +2,31 @@ import { Modifier } from '../../core/src/Modifier';
 import { Modifiers } from '../../core/src/Modifiers';
 import { Attributes } from '../../plugin-xml/src/Attributes';
 
+/**
+ * A Format might represent DOM inline nodes (a, b, strong) that were parsed or
+ * the DOM inline node (a, b, strong) that we might render.
```
Zinston, line 7: There is a TODO just below: "remove this reference to DOM."
I would try and rewrite this to not invoke the DOM.
Plus, all the uses of "might" make this very hesitant ^^

Now what this does however is reveal how a `Format` is not fundamentally different from any other `Modifier` apart from its DOM rendering... @dmo-odoo Any input?

packages/plugin-inline/src/Format.ts
```diff
@@ -2,16 +2,31 @@ import { Modifier } from '../../core/src/Modifier';
...
+ */
 export class Format extends Modifier {
     htmlTag: string; // TODO: remove this reference to DOM.
+    /**
+     * The modifiers for a format is `Attribute`.
```
Zinston, line 12: No idea what that sentence is supposed to mean. If you mean that the only modifier a `Format` can have is `Attributes`, that's wrong. `Modifiers` can potentially hold any kind of information to modify a node or a format.

packages/plugin-inline/src/Format.ts
```diff
@@ -31,6 +46,11 @@ export class Format extends Modifier {
     // Public
     //--------------------------------------------------------------------------
 
+    /**
+     * Render the current format into a dom Node.
+     *
+     * Meant to be used with `FormatDomRenderer`.
+     */
```
Zinston, line 53: `// TODO: remove this reference to DOM.`
(my bad)

PR #194, opened 2020-05-29
[ADD] Test: Modifiers, Modifier

Zinston:


packages/core/src/Modifiers.ts
```diff
@@ -268,7 +268,7 @@ export class Modifiers {
      * @param callbackfn
      */
     some(callbackfn: (value: Modifier, index: number, array: Modifier[]) => unknown): boolean {
-        return this._contents?.some(callbackfn);
+        return this._contents?.some(callbackfn) || false;
```
dmo-odoo, line 267: @sebgeelen My bad: `?.` actually returns `undefined` when it fails rather than `false`. Here is a particular case where the difference matter. cc @Zinston

packages/core/src/Modifiers.ts
```diff
@@ -143,10 +143,6 @@ export class Modifiers {
     ): T[] {
         if (!this._contents) {
             return [];
-        } else if (modifier instanceof Modifier) {
-            // `modifier` is an instance of `Modifier` -> find it in the array
-            // and return it in a new array.
-            return this._contents.filter(m => m === modifier) as T[];
```
dmo-odoo, line 149: 🤷

Zinston: Yes that was spectacularly useless :-)

Issue #187, opened 2020-05-27
Link + enter


Goaman:
When hitting enter, the new format should not be a link.
Video https://youtu.be/5Lf9dSy7pZk

dmo-odoo:
Wait. It should not be a link but it should be, say, bold or underline or italic ? All of them are modifiers.

Goaman:
Seems fine for me.

Goaman:
@sebgeelen

sebgeelen:
should be fixed now since #197 is merged

PR #197, opened 2020-06-02
[IMP] Modifier, Link : Add a preserve property , better behaviors of Link format

sebgeelen:
This property is used to determine if a modifier should be kept after a break line or not.

see issue #187

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,13 +1,12 @@
+# dotfiles
+.*
+!.gitignore
+# misc
 node_modules/
 tmp/
 dist/
-.cache/
 yarn.lock
 *.log
-.DS_Store
-.DS_Store?
-.Spotlight-V100
-.Trashes
 ehthumbs.db
 Thumbs.db
 data/
--- a/packages/core/src/Modifier.ts
+++ b/packages/core/src/Modifier.ts
@@ -8,6 +8,7 @@ export interface Modifier {
     constructor: ModifierConstructor & this;
 }
 export class Modifier {
+    preserve = true;
     get name(): string {
         return '';
     }
--- a/packages/plugin-char/src/Char.ts
+++ b/packages/plugin-char/src/Char.ts
@@ -47,7 +47,12 @@ export class Char<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
         const range = params.context.range;
         const text = params.text;
         const inline = this.editor.plugins.get(Inline);
-        const modifiers = inline.getCurrentModifiers(range);
+        let modifiers = inline.getCurrentModifiers(range);
+        // Ony preserved modifiers are applied at the start of a container.
+        const previousSibling = range.start.previousSibling();
+        if (!previousSibling) {
+            modifiers = new Modifiers(...modifiers.filter(mod => mod.preserve));
+        }
         if (params.formats) {
             modifiers.set(...params.formats.map(format => format.clone()));
         }
--- a/packages/plugin-inline/src/Inline.ts
+++ b/packages/plugin-inline/src/Inline.ts
@@ -8,6 +8,7 @@ import { Constructor } from '../../utils/src/utils';
 import { VNode } from '../../core/src/VNodes/VNode';
 import { Modifiers } from '../../core/src/Modifiers';
 import { CssStyle } from '../../plugin-xml/src/CssStyle';
+import { LineBreakNode } from '../../plugin-linebreak/src/LineBreakNode';
 import { Loadables } from '../../core/src/JWEditor';
 import { Renderer } from '../../plugin-renderer/src/Renderer';
 import { InlineFormatDomObjectRenderer } from './InlineFormatDomObjectRenderer';
@@ -147,7 +148,10 @@ export class Inline<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<
 
         let inlineToCopyModifiers: VNode;
         if (range.isCollapsed()) {
-            inlineToCopyModifiers = range.start.previousSibling() || range.start.nextSibling();
+            // TODO: LineBreakNode should have the formats as well.
+            inlineToCopyModifiers =
+                range.start.previousSibling(node => !node.test(LineBreakNode)) ||
+                range.start.nextSibling();
         } else {
             inlineToCopyModifiers = range.start.nextSibling();
         }
--- a/packages/plugin-layout/src/LayoutEngine.ts
+++ b/packages/plugin-layout/src/LayoutEngine.ts
@@ -8,7 +8,7 @@ export type DomZonePosition = 'after' | 'prepend' | 'append' | 'replace';
 
 export interface ComponentDefinition {
     id: ComponentId;
-    render: (editor: JWEditor) => Promise<VNode[]>;
+    render: (editor: JWEditor, props?: {}) => Promise<VNode[]>;
 }
 
 export abstract class LayoutEngine {
@@ -81,17 +81,22 @@ export abstract class LayoutEngine {
      * the default zone.
      * Return every created instance
      *
-     * @param componentDefinition
+     * @param componentId
      * @param zoneId
+     * @param props
      */
-    async add(componentId: ComponentId, zoneId: ZoneIdentifier): Promise<VNode[]> {
+    async add(componentId: ComponentId, zoneId: ZoneIdentifier, props?: {}): Promise<VNode[]> {
         const allZones = [this.root, ...this.root.descendants(ZoneNode)];
         let matchingZones = allZones.filter(node => node.managedZones.includes(zoneId));
         if (!matchingZones.length) {
             matchingZones = allZones.filter(zone => zone.managedZones.includes('default'));
         }
         const componentDefinition = this.componentDefinitions[componentId];
-        const newComponents = await this._instantiateComponent(componentDefinition, matchingZones);
+        const newComponents = await this._instantiateComponent(
+            componentDefinition,
+            matchingZones,
+            props,
+        );
         return this._fillZones(newComponents);
     }
     /**
@@ -219,12 +224,13 @@ export abstract class LayoutEngine {
     private async _instantiateComponent(
         componentDefinition: ComponentDefinition,
         zones: ZoneNode[],
+        props?: {},
     ): Promise<VNode[]> {
         const components = this.components.get(componentDefinition.id) || [];
         // Add into the container.
         const newComponents: VNode[] = [];
         for (const zone of zones) {
-            const nodes = await componentDefinition.render(this.editor);
+            const nodes = await componentDefinition.render(this.editor, props);
             components.push(...nodes);
             newComponents.push(...nodes);
             zone.append(...nodes);
--- a/packages/plugin-link/src/Link.ts
+++ b/packages/plugin-link/src/Link.ts
@@ -37,7 +37,7 @@ export class Link<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
             node = link;
         }
         const format = node.is(InlineNode) && node.modifiers.find(LinkFormat);
-        return link instanceof AbstractNode ? !!format : format === link;
+        return link instanceof AbstractNode ? !!format : link.isSameAs(format);
     }
     static dependencies = [Inline];
     commands = {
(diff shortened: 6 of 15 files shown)


.gitignore
```diff
@@ -1,13 +1,13 @@
+# dotfiles
+.*
```
dmo-odoo, line 2: Smart ! Then I guess you could have removed `.cache` from the list... ;-) That being said, it would be better if this was in a separate commit as it is not related to the `Modifier` improvement here. Same PR is ok but not same commit. I suggest `[IMP] Core: improve gitignore` or something along these lines.

packages/core/src/Modifier.ts
```diff
@@ -8,6 +8,7 @@ export interface Modifier {
     constructor: ModifierConstructor & this;
 }
 export class Modifier {
+    public preserve: boolean = true;
```
dmo-odoo, line 11: `public` is the default, we have chosen to omit it when defining class properties. The only place where we use it is in constructor parameters.

packages/plugin-char/src/Char.ts
```diff
@@ -47,7 +48,12 @@ export class Char<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
         const range = params.context.range;
         const text = params.text;
         const inline = this.editor.plugins.get(Inline);
-        const modifiers = inline.getCurrentModifiers(range);
+        let modifiers = inline.getCurrentModifiers(range);
+
+        const cleanModifier = range.start.previousSibling() === undefined || range.start.previousSibling().is(LineBreakNode);
```
dmo-odoo, line 53: I generally avoid to write `undefined` unless in rare cases where it can be a somewhat valid value.

In this particular case I would have done:
```typescript
const previousSibling = range.start.previouspreviousSibling();
if (!previousSibling || previousSibling.is(LineBreakNode)) {
```

In the version with `undefined`, you have to "translate" in your brain the fact that `previousSibling` being `undefined` means that there is no previous sibling. The suggested code is more idiomatic IMO because it reads exactly what it means: "if there is no previous sibling or it is a LineBreakNode". You don't need to wrap your brain about anything in particular.

sebgeelen: Agree.

packages/core/src/Modifier.ts
sebgeelen: I'll rewrite the commit ;)

packages/plugin-char/src/Char.ts
```diff
@@ -47,7 +48,12 @@ export class Char<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
         const range = params.context.range;
         const text = params.text;
         const inline = this.editor.plugins.get(Inline);
-        const modifiers = inline.getCurrentModifiers(range);
+        let modifiers = inline.getCurrentModifiers(range);
+
+        const previousSibling = range.start.previousSibling();
+        if(!previousSibling || previousSibling.is(LineBreakNode)) {
```
Zinston, line 54: Here's the unlinted culprit :-) `if (`

packages/core/src/Modifier.ts
```diff
@@ -8,6 +8,7 @@ export interface Modifier {
     constructor: ModifierConstructor & this;
 }
 export class Modifier {
+    preserve: boolean = true;
```
Zinston, line 11: Is the typing necessary? Doesn't typescript recognize that setting the property to `true` means its type is boolean? Or does it think the type is `true` then?

dmo-odoo: No it's not necessary. I think our prettier config fixes that kind of things automatically.

packages/plugin-char/src/Char.ts
```diff
@@ -1,5 +1,6 @@
 import { JWPlugin, JWPluginConfig } from '../../core/src/JWPlugin';
 import { CharNode } from './CharNode';
+import { LineBreakNode } from '../../plugin-linebreak/src/LineBreakNode';
```
Zinston, line 3: Be careful when referencing plugins in other plugins. This one should not be mandatory for `Char` to work.

packages/plugin-link/src/Link.ts
```diff
@@ -30,7 +30,7 @@ export class Link<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
             node = link;
         }
         const format = node.is(InlineNode) && node.modifiers.find(LinkFormat);
-        return link instanceof AbstractNode ? !!format : format === link;
+        return link instanceof AbstractNode ? !!format : link.isSameAs(format);
```
Zinston, line 33: :+1:

packages/plugin-dom-layout/src/ui/DomLayoutEngine.ts
```diff
@@ -120,8 +120,12 @@ export class DomLayoutEngine extends LayoutEngine {
...
      * @override
      */
-    async add(componentId: ComponentId, zoneId: ZoneIdentifier): Promise<VNode[]> {
-        const nodes = await super.add(componentId, zoneId);
+    async add(
+        componentId: ComponentId,
+        zoneId: ZoneIdentifier,
+        params?: Record<string, any>,
+    ): Promise<VNode[]> {
+        const nodes = await super.add(componentId, zoneId, params);
```
Zinston, line 128: This whole thing is cool but should be split into its own commit.

packages/plugin-link/src/Link.ts
```diff
@@ -73,32 +73,54 @@ export class Link<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
...
+    }
+
+    //--------------------------------------------------------------------------
+    // Private
+    //--------------------------------------------------------------------------
 
-        const sameLink: Typeguard<InlineNode> = Link.isLink.bind(Link, link);
-        for (const inline of [node, ...node.adjacents(sameLink)]) {
-            inline.modifiers.remove(link);
+    _removeLinkOnNodes(nodes: VNode[]): void {
```
Zinston, line 120: doc please :-)

packages/plugin-link/src/Link.ts
```diff
@@ -73,32 +73,54 @@ export class Link<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
...
+        for (const node of nodes) {
+            const link = node.modifiers.find(LinkFormat);
+            node.modifiers.remove(link);
```
Zinston, line 123: `node.modifiers.remove(LinkFormat)` should suffice :-)

packages/plugin-xml/src/Attributes.ts
```diff
@@ -5,6 +5,7 @@ import { ClassList } from './ClassList';
 export class Attributes extends Modifier {
     private _record: Record<string, string> = {};
     style = new CssStyle();
+    preserve = false;
```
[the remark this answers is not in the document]
sebgeelen: You tell me. 

@dmo-odoo or you ( I don't remember) told me to put the preserve false on attributes. 
I can't say I understand why.

packages/plugin-layout/src/LayoutEngine.ts
```diff
@@ -8,7 +8,7 @@ export type DomZonePosition = 'after' | 'prepend' | 'append' | 'replace';
 
 export interface ComponentDefinition {
     id: ComponentId;
-    render: (editor: JWEditor) => Promise<VNode[]>;
+    render: (editor: JWEditor, params?: Record<string, any>) => Promise<VNode[]>;
```
dmo-odoo, line 11: Your linter let you do that ? That's weird because we're supposed to have a shared configuration and ours does not let us write `any`. We usually write `{}` instead. Like `any` it matches everything, but it will generate an error if you try to use it in any meaningful way.

With any, you can do that:
```ts
var a: Any = { a: 'a' };
console.log(a.b); // no warning
```

With `{}`, you can't do that:
```ts
var a: {} = { a: 'a' };
console.log(a.b); // warning: `{}` does not have a `b` property
```

packages/plugin-link/src/Link.ts
```diff
@@ -65,8 +65,8 @@ export class Link<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
         components: [
             {
                 id: 'link',
-                async render(): Promise<OwlNode[]> {
-                    return [new OwlNode({ Component: LinkComponent, props: {} })];
+                async render(editor: JWEditor, params?: Record<string, any>): Promise<OwlNode[]> {
```
dmo-odoo, line 68: @Zinston @Goaman Shall we name the component params as `props` or do you think it might get confusing if people use them as something else than props in their component's `render` method ?

packages/plugin-link/src/Link.ts
```diff
@@ -132,23 +138,41 @@ export class Link<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
...
+
+    //--------------------------------------------------------------------------
+    // Private
+    //--------------------------------------------------------------------------
 
-        const sameLink: Typeguard<InlineNode> = Link.isLink.bind(Link, link);
-        for (const inline of [node, ...node.adjacents(sameLink)]) {
-            inline.modifiers.remove(link);
+    _removeLinkOnNodes(nodes: VNode[]): void {
+        // remove all link modifiers on the provided nodes
```
dmo-odoo, line 172: Can I ask you to capitalize the first letters of your comments and add a dot `.` at the end so it looks like a proper sentence ? We all do that (except Christophe when he forgets) and it makes the comments look more professional. This obviously applies to all comments.

packages/plugin-link/src/Link.ts
dmo-odoo: This comment should not have been resolved as it is still relevant as of now. I'm unresolving it.

packages/plugin-char/src/Char.ts
```diff
@@ -47,7 +47,12 @@ export class Char<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
         const range = params.context.range;
         const text = params.text;
         const inline = this.editor.plugins.get(Inline);
-        const modifiers = inline.getCurrentModifiers(range);
+        let modifiers = inline.getCurrentModifiers(range);
+
+        const previousSibling = range.start.previousSibling();
+        if (!previousSibling) {
```
dmo-odoo, line 53: I was confused by this for a moment (even though I knew what the feature was about) so I suggest to add a comment on the line before `const previousSibling`:
```// Ony preserved modifiers are applied at the start of a container.```

File: packages/plugin-inline/src/Inline.ts [hunk 16]
```diff
@@ -147,7 +148,9 @@ export class Inline<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<
 
         let inlineToCopyModifiers: VNode;
         if (range.isCollapsed()) {
-            inlineToCopyModifiers = range.start.previousSibling() || range.start.nextSibling();
+            inlineToCopyModifiers =
+                range.start.previousSibling(node => !node.test(LineBreakNode)) ||
+                range.start.nextSibling();
```
dmo-odoo, line 153: I'm not sure I understand this fix. What it does is that it takes the first previous sibling that is *not* a `LineBreakNode`, but why does it matter ? Besides, are there no tests for this fix or are they all put together in the "tests" commit ?

packages/plugin-layout/src/LayoutEngine.ts
sebgeelen: Nice to know, I will change it

packages/plugin-inline/src/Inline.ts
sebgeelen: This allow the `getCurrentModifiers()` method to return the same modifiers whether `this.cache.modifiers` is empty or not.
This fix a difference of behaviour between typing a character just after a `LineBreak` or after changing the cursor position the beginning of the line and typing a character.
In the latter, the modifiers was empty.

If this is not a issue, we can remove this commit.

packages/plugin-link/src/Link.ts
sebgeelen: The comment below is not clear enough ?

packages/plugin-link/src/Link.ts
sebgeelen: will do

packages/plugin-link/src/Link.ts
dmo-odoo: Only proper function doc written above the function in `/**` format is read by automated documentation systems like `tsdoc`. This is why we (try to) document all our functions in that way. You can open other files to see examples of it. Please note that we do not document the types of the parameters since this is redundant with the typing of Typescript which is correctly parsed by `tsdoc`.

packages/plugin-link/src/Link.ts
Goaman: Why would you want to use them as something else than props?
For me, props is good because it's the standard way of naming the parameters that a component receive.

dmo-odoo: cc @sebgeelen for renaming the `params` parameter into `props` up along the entire function call chain.

PR #201, opened 2020-06-05
[FIX] dom-layout: long press bacspace can create an altered

Gorash:
This fix, avoid to have error when we try to locate a VNode where the
rendering is altered and not inside the dom.
Fix also the re-insertion of node removed from alteration.


dmo-odoo:
Pour les prochains fixes, tu veux bien écrire "Dom Layout" au lieu de dom-layout s'il te plaît ? Merci ! ;)

PR #203, opened 2020-06-08
[FIX] OdooSnippet: apply classes to LinkFormat instead of CharNodes

Zinston:


packages/plugin-divider/src/DividerXmlDomParser.ts
```diff
@@ -14,6 +15,11 @@ export class DividerXmlDomParser extends AbstractParser<Node> {
     async parse(item: Element): Promise<DividerNode[]> {
         const divider = new DividerNode();
         divider.modifiers.append(this.engine.parseAttributes(item));
+
+        // `DividerNode` can be configured to be unbreakable.
+        const configuration: DividerConfig = this.engine.editor.plugins.get(Divider).configuration;
+        divider.breakable = 'breakable' in configuration ? !!configuration.breakable : true;
```
dmo-odoo, line 21: I think you should do a `defaultConfig` like you did somewhere else and do something like:
`this.config = { ...this.defaultConfig, config }`
Such that we can simply do `divider.breakable = configuration.breakable` here.

packages/plugin-odoo-snippets/src/OdooSnippet.ts
```diff
@@ -182,8 +182,9 @@ export class OdooSnippet<T extends JWPluginConfig = JWPluginConfig> extends JWPl
     }
     addClassToLink(params: AddClassToLinkParams): void {
         const nodes = params.context.range.targetedNodes(InlineNode);
-        for (const node of nodes) {
-            node.modifiers.get(Attributes).set('class', params.classes);
+        const links = nodes.map(node => node.modifiers.find(LinkFormat)).filter(f => f);
```
dmo-odoo, line 185: What's the point of the `filter(f => f)` ?

packages/plugin-divider/src/DividerXmlDomParser.ts
Zinston: Done.

packages/plugin-odoo-snippets/src/OdooSnippet.ts
Zinston: To not have the ones that are undefined...

dmo-odoo: This is sad. Just because I might regret not asking it: would using `get` be wrong in this case ? (I'm pretty sure it would though because I assume `url` wouldn't be set.) To be fair I'm not sure I understand the point of the filter as the method is called `addClassToLink` so clearly all the given nodes are supposed to be part of a link, aren't they ? What would be the expected behavior of this function for a node that would not yet be inside of a link ?

Zinston: I assumed the same as you so I think it would be wrong. Now in the cleaning of the integration I'll get a chance to make 100% sure the nodes all have a link to begin with so I'll be able to remove that filter.

dmo-odoo: But in the meantime you want it merged, right?

dmo-odoo: To be fair I think you can merge it yourself since it does not target master :)

Zinston: Ok...

PR #213, opened 2020-06-11
[FIX] Devtools: can display: “Beware the Jabberwock, my son!

Gorash:
In the devtools, the special char “ and the rest of the sentence are
displayed out of the box on the right.


dmo-odoo:
cc @Zinston

dmo-odoo:
I'm closing this PR because it was in fact a renderer bug of Chrome. It did not work at least in Chrome 80 but it does in Chrome 83.

PR #220, opened 2020-06-18
[FIX] DomLayout: inline node who use default renderer

Gorash:
The inline node can render child charNode before an inline node and
wrap into a objectFragment. The Inline node have no reference because
the domObject in 'unfilterdOldObjectMap' allready exist.


dmo-odoo:
Merged in https://github.com/odoo-dev/jabberwock/commit/c96999cdf8bd02f702a0a35153dd8889f4d47ada.

PR #226, opened 2020-06-22
Master odoo add prepend nby

Goaman:
I need to prepend to an element in Odoo


packages/plugin-dom-helpers/src/DomHelpers.ts
```diff
@@ -175,6 +175,19 @@ export class DomHelpers<T extends JWPluginConfig = JWPluginConfig> extends JWPlu
         }
         await this.editor.dispatcher.dispatchHooks('@redraw');
     }
+    /**
+     * Prepend a DOM Node inside the child of another.
```
Zinston, line 179: "Prepend a DOM Node to another DOM Node's children." - otherwise it seems like you want to prepend it to that other node's child.

packages/plugin-dom-helpers/src/DomHelpers.ts
```diff
@@ -175,6 +175,19 @@ export class DomHelpers<T extends JWPluginConfig = JWPluginConfig> extends JWPlu
...
+     * @param params
+     */
+    async prepend(fromDomNode: Node, toDomNode: Node): Promise<void> {
+        const toNodes = this.getNodes(toDomNode);
+        const toNode = toNodes[toNodes.length - 1];
+        for (const fromNode of this.getNodes(fromDomNode).reverse()) {
+            fromNode.prepend(toNode);
+        }
+        await this.editor.dispatcher.dispatchHooks('@redraw');
+    }
```
Zinston, line 190: So... the implementation is alright by me, but I have questions regarding the concept. We have `moveBefore`. Can't it already be used for that?
Like `moveBefore(fromDomNode, toDomNode.firstChild)` ? If it's specifically for nodes that don't have children yet, then why specifically `prepend` and not `append` ?
In any case, if you make a `prepend`, please also make a `append` :-)

Goaman: The reason is that we do not want to check if there is a child or not.

packages/plugin-dom-helpers/src/DomHelpers.ts
```diff
@@ -175,6 +175,32 @@ export class DomHelpers<T extends JWPluginConfig = JWPluginConfig> extends JWPlu
         }
         await this.editor.dispatcher.dispatchHooks('@redraw');
     }
+    /**
+     * Prepend a DOM Node to another DOM Node's children.
+     *
+     * @param params
+     */
+    async prepend(fromDomNode: Node, toDomNode: Node): Promise<void> {
```
dmo-odoo, line 183: I don't understand what "prepend from" and "prepend to" mean. If I prepend "to a node" then I prepend inside this node ? Then `toDomNode` is the node in which the node is prepended ? Or I prepend "from" the node ? I have no idea what that means as I think this is not proper english. Obviously, `append` has the same problem. I suggest `node` and `container` for parameters. cc @Zinston if I'm just dumb by not understanding the english of "from" and "to" as prepositions for "prepend".

Goaman: Corrected

packages/plugin-dom-helpers/src/DomHelpers.ts
```diff
@@ -175,6 +175,32 @@ export class DomHelpers<T extends JWPluginConfig = JWPluginConfig> extends JWPlu
         }
         await this.editor.dispatcher.dispatchHooks('@redraw');
     }
+    /**
+     * Prepend a DOM Node to another DOM Node's children.
+     *
+     * @param params
+     */
+    async prepend(container: Node, node: Node): Promise<void> {
```
dmo-odoo, line 183: We discussed with @Zinston and the only proper way in english is to say "prepend X to Y" so we would switch the order of the parameters. Same for `append`, obviously.

Goaman: I disagree with the english argument. As we saw with sge and the command insertHTML, it is confusing to have to think about what english rule is for any command. It is far simpler to always set the node that will be modified as the first argument as 99% of our helpers.

Zinston: Well actually either way `node` has to be first... `insertHTML` has the html to insert as first argument so `prepend` should have the node to insert as first argument as well.

Goaman: No, all other command except `insertHTML` has the "target" as the first argument. It's `insertHTML` that should have it's current argument changed.

PR #231, opened 2020-06-26
[IMP] Core: downgrage build to es2017

Goaman:
Currently Odoo is configured to suppot maximum es2017.

### DIFF
--- a/packages/plugin-devtools/src/components/ShortcutsComponent.ts
+++ b/packages/plugin-devtools/src/components/ShortcutsComponent.ts
@@ -1,16 +1,14 @@
 import { OwlComponent } from '../../../plugin-owl/src/OwlComponent';
 import { Keymap } from '../../../plugin-keymap/src/Keymap';
 import { argsRepr } from '../utils';
+import { flat } from '../../../utils/src/utils';
 
 export class ShortcutsComponent extends OwlComponent<{}> {
-    mappings = this.env.editor.plugins
-        .get(Keymap)
-        .mappings.flat()
-        .sort((a, b) => {
-            if (a.configuredCommand.commandId < b.configuredCommand.commandId) return -1;
-            if (a.configuredCommand.commandId > b.configuredCommand.commandId) return 1;
-            return 0;
-        });
+    mappings = flat(this.env.editor.plugins.get(Keymap).mappings).sort((a, b) => {
+        if (a.configuredCommand.commandId < b.configuredCommand.commandId) return -1;
+        if (a.configuredCommand.commandId > b.configuredCommand.commandId) return 1;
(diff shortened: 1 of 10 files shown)


packages/plugin-list/src/ListDomObjectRenderer.ts
```diff
@@ -28,7 +28,7 @@ export class ListDomObjectRenderer extends AbstractRenderer<DomObject> {
             }
         }
         const renderedChildren = await this.engine.renderChildren(node);
-        domObject.children.push(...renderedChildren.flat());
+        domObject.children.push(...[].concat(...renderedChildren));
```
Goaman, line 31: I did not succeed making the signature of `flat` match the type of `renderedChildren`.

dmo-odoo: Well it's because Typescript is smarter than we all are. The flat on this line is historical and not needed anymore. I expect `domObject.children.push(...renderedChildren);` to work.

packages/plugin-devtools/src/components/ShortcutsComponent.ts
```diff
@@ -1,16 +1,14 @@
...
-        .sort((a, b) => {
-            if (a.configuredCommand.commandId < b.configuredCommand.commandId) return -1;
-            if (a.configuredCommand.commandId > b.configuredCommand.commandId) return 1;
-            return 0;
-        });
+    mappings = flat(this.env.editor.plugins.get(Keymap).mappings).sort((a, b) => {
+        if (a.configuredCommand.commandId < b.configuredCommand.commandId) return -1;
+        if (a.configuredCommand.commandId > b.configuredCommand.commandId) return 1;
+        return 0;
+    });
```
dmo-odoo, line 11: Surprisingly, this new `flat` makes things more readable across the board.

packages/plugin-list/src/ListDomObjectRenderer.ts
Goaman: Ok I trie that

Goaman: Done, all test pass

dmo-odoo:
@Goaman So you managed to get AL's reply on the matter of es2017 ?

Goaman:
Yes, the policy is to keep the version of es2014.

dmo-odoo:
"2014" ? You mean 2017, right ? So no more explanation other than "it's the policy" ?

Goaman:
Yep "es2017". The reason is that like linux distribution (e.g. ubutnu), there is LTS support and version that are froozen to be sure it remain stable across time.

PR #232, opened 2020-06-26
Add modes and integrate for translation

Zinston:
Related Odoo PR: https://github.com/odoo-dev/odoo/pull/295

### DIFF
--- a/packages/bundle-odoo-website-editor/OdooWebsiteEditor.ts
+++ b/packages/bundle-odoo-website-editor/OdooWebsiteEditor.ts
@@ -49,7 +49,7 @@ import { FontAwesome } from '../plugin-fontawesome/src/FontAwesome';
 import './odooLayout.css';
 import { OdooField } from '../plugin-odoo-field/src/OdooField';
 
-interface OdooWebsiteEditorOption {
+interface OdooWebsiteEditorOptions {
     source: HTMLElement;
     location: [Node, DomZonePosition];
     customCommands: Record<CommandIdentifier, CommandImplementation>;
@@ -63,7 +63,7 @@ interface OdooWebsiteEditorOption {
 }
 
 export class OdooWebsiteEditor extends JWEditor {
-    constructor(options: OdooWebsiteEditorOption) {
+    constructor(options: OdooWebsiteEditorOptions) {
         super();
         class CustomPlugin extends JWPlugin {
             commands = options.customCommands;
--- a/packages/bundle-odoo-website-editor/OdooWebsiteEditor.ts
+++ b/packages/bundle-odoo-website-editor/OdooWebsiteEditor.ts
@@ -33,10 +33,10 @@ import { Layout } from '../plugin-layout/src/Layout';
 import { DomLayout } from '../plugin-dom-layout/src/DomLayout';
 import { DomEditable } from '../plugin-dom-editable/src/DomEditable';
 import { VNode } from '../core/src/VNodes/VNode';
-import { Toolbar } from '../plugin-toolbar/src/Toolbar';
+import { Toolbar, ToolbarLayout } from '../plugin-toolbar/src/Toolbar';
 import { HtmlNode } from '../plugin-html/src/HtmlNode';
 import { CommandImplementation, CommandIdentifier } from '../core/src/Dispatcher';
-import { JWPlugin } from '../core/src/JWPlugin';
+import { JWPlugin, JWPluginConfig } from '../core/src/JWPlugin';
 import { OdooVideo } from '../plugin-odoo-video/src/OdooVideo';
 import { DomZonePosition } from '../plugin-layout/src/LayoutEngine';
 import { HtmlDomRenderingEngine } from '../plugin-html/src/HtmlDomRenderingEngine';
@@ -46,7 +46,7 @@ import { parseEditable } from '../utils/src/configuration';
 import { Dialog } from '../plugin-dialog/src/Dialog';
 import { Shadow } from '../plugin-shadow/src/Shadow';
 import { FontAwesome } from '../plugin-fontawesome/src/FontAwesome';
-import { ModeDefinition } from '../plugin-mode/src/Mode';
+import { ModeDefinition } from '../core/src/Mode';
 import './odooLayout.css';
 
 interface OdooWebsiteEditorOptions {
@@ -57,12 +57,34 @@ interface OdooWebsiteEditorOptions {
     snippetMenuElement?: HTMLElement;
     snippetManipulators?: HTMLElement;
     template?: string;
+    toolbarLayout?: ToolbarLayout;
     mode?: ModeDefinition;
     // todo: Remove when configuring the toolbar in another way.
     discardButton?: boolean;
     saveButton?: boolean;
+    plugins?: [typeof JWPlugin, JWPluginConfig?][];
 }
 
+const defaultToolbarLayout = [
+    [
+        [
+            'ParagraphButton',
+            'Heading1Button',
+            'Heading2Button',
+            'Heading3Button',
+            'Heading4Button',
+            'Heading5Button',
+            'Heading6Button',
+            'PreButton',
+        ],
+    ],
+    ['BoldButton', 'ItalicButton', 'UnderlineButton', 'RemoveFormatButton'],
+    ['AlignLeftButton', 'AlignCenterButton', 'AlignRightButton', 'AlignJustifyButton'],
+    ['OrderedListButton', 'UnorderedListButton', 'ChecklistButton'],
+    ['IndentButton', 'OutdentButton'],
+    ['OdooLinkButton'],
+    ['OdooMediaButton'],
+];
 export class OdooWebsiteEditor extends JWEditor {
     constructor(options: OdooWebsiteEditorOptions) {
         super();
@@ -111,32 +133,15 @@ export class OdooWebsiteEditor extends JWEditor {
                 [DomHelpers],
                 [Odoo],
                 [OdooVideo],
-                [OdooField],
                 [CustomPlugin],
+                ...(options.plugins || []),
             ],
         });
         this.configure(Toolbar, {
             layout: [
-                [
-                    [
-                        'ParagraphButton',
-                        'Heading1Button',
-                        'Heading2Button',
-                        'Heading3Button',
-                        'Heading4Button',
-                        'Heading5Button',
-                        'Heading6Button',
-                        'PreButton',
-                    ],
-                ],
-                ['BoldButton', 'ItalicButton', 'UnderlineButton', 'RemoveFormatButton'],
-                ['AlignLeftButton', 'AlignCenterButton', 'AlignRightButton', 'AlignJustifyButton'],
-                ['OrderedListButton', 'UnorderedListButton', 'ChecklistButton'],
-                ['IndentButton', 'OutdentButton'],
-                ['OdooLinkButton'],
-                ['OdooMediaButton'],
-                ...(options.discardButton ? [['OdooDiscardButton']] : []),
+                ...(options.toolbarLayout || defaultToolbarLayout),
                 ...(options.saveButton ? [['OdooSaveButton']] : []),
+                ...(options.discardButton ? [['OdooDiscardButton']] : []),
             ],
         });
 
@@ -209,7 +214,7 @@ export class OdooWebsiteEditor extends JWEditor {
             source: options.source.firstElementChild as HTMLElement,
         });
         if (options.mode) {
-            this.load({
+            this.configure({
                 modes: [options.mode],
             });
             this.configure({ mode: options.mode.id });
--- a/packages/bundle-odoo-website-editor/OdooWebsiteEditor.ts
+++ b/packages/bundle-odoo-website-editor/OdooWebsiteEditor.ts
@@ -46,8 +46,8 @@ import { parseEditable } from '../utils/src/configuration';
 import { Dialog } from '../plugin-dialog/src/Dialog';
 import { Shadow } from '../plugin-shadow/src/Shadow';
 import { FontAwesome } from '../plugin-fontawesome/src/FontAwesome';
+import { ModeDefinition } from '../plugin-mode/src/Mode';
 import './odooLayout.css';
-import { OdooField } from '../plugin-odoo-field/src/OdooField';
 
 interface OdooWebsiteEditorOptions {
     source: HTMLElement;
@@ -57,6 +57,7 @@ interface OdooWebsiteEditorOptions {
     snippetMenuElement?: HTMLElement;
     snippetManipulators?: HTMLElement;
     template?: string;
+    mode?: ModeDefinition;
     // todo: Remove when configuring the toolbar in another way.
     discardButton?: boolean;
     saveButton?: boolean;
@@ -93,7 +94,7 @@ export class OdooWebsiteEditor extends JWEditor {
                 [Underline],
                 [Link],
                 [FontAwesome],
-                [Divider, { breakable: false }],
+                [Divider],
                 [Image],
                 [Subscript],
                 [Superscript],
@@ -207,6 +208,12 @@ export class OdooWebsiteEditor extends JWEditor {
             autoFocus: true,
             source: options.source.firstElementChild as HTMLElement,
         });
+        if (options.mode) {
+            this.load({
+                modes: [options.mode],
+            });
+            this.configure({ mode: options.mode.id });
+        }
     }
 
     /**
--- a/packages/core/src/ContextManager.ts
+++ b/packages/core/src/ContextManager.ts
@@ -18,6 +18,32 @@ export interface Contextual {
     check?: (context: CheckingContext) => boolean;
 }
 
+export interface ContextualEntry<T> extends Contextual {
+    selector: Predicate[];
+    value: T;
+}
+
+export interface RankedResult<T> {
+    lvl1Score: number;
+    lvl2Score: number;
+    index: number;
+    matched: VNode[];
+    entry: ContextualEntry<T>;
+}
+
+/**
+ * Class that rank a hierarchy of vnode through a "specificity" algorithm.
+ *
+ * Specificity level is defined with:
+ * - `lvl2`: if the last predicate of an item selector is deeper in the tree
+ *   than the last predicate of another item selector; the first item have
+ *   more specificity
+ * - `lvl1`: if two or more items have the same `lvl2` specificity; the
+ *   command with the longest selector will have more specificity
+ * - `lvl0`: if the item has no selector (an empty list), there is no
+ *   specificity
+ * ```
+ */
 export class ContextManager {
     editor: JWEditor;
     defaultContext: Context;
@@ -34,138 +60,111 @@ export class ContextManager {
     //--------------------------------------------------------------------------
 
     /**
-     * Match a command from the current context (the `vDocument` selection).
-     *
-     * Search through all command with `commandId` and return the last
-     * `CommandImplementation` that match a particular context.
-     *
-     * Commands with more specificity will have priority.  If there is multiples
-     * commands with the same specificity, the last defined will be retured.
-     *
-     * The context is the `editor.selection.range.start.ancestors()`.
-     *
-     * Specificity is defined with:
-     * - `lvl2`: if the last predicate of a command "a"'s selector is deeper in
-     *   the tree than the last predicate of a command "b"'s selector; command
-     *   "a" have more specificity
-     * - `lvl1`: if two or more commands have the same `lvl2` specificity; the
-     *   command with the longest selector will have more specificity
-     * - `lvl0`: if the command has no selector, there is no specificity
-     *
-     * For example:
-     * ```typescript
-     * const commandUlP: CommandImplementation = {
-     *     selector: [isUl, isP],
-     *     handler: () =>{},
-     * }
-     * const commandP: CommandImplementation = {
-     *     selector: [isP],
-     *     handler: () =>{},
-     * }
-     * const commandUl: CommandImplementation = {
-     *     selector: [isUl],
-     *     handler: () =>{},
-     * }
-     * const commandImage: CommandImplementation = {
-     *     selector: [isImage],
-     *     handler: () =>{},
-     * }
-     * const commandAny: CommandImplementation = {
-     *     selector: [],
-     *     handler: () =>{},
-     * }
-     * dispatcher.registerCommand('command', commandUlP)
-     * dispatcher.registerCommand('command', commandP)
-     * dispatcher.registerCommand('command', commandUl)
-     * dispatcher.registerCommand('command', commandImage)
-     * dispatcher.registerCommand('command', commandAny)
-     * const result = dispatcher._mach('command')
-     * ```
-     *
-     * If the document looks like:
-     * ```html
-     * <ul>
-     *     <li>
-     *         <!-- The char "[]" represent the collapsed selection -->
-     *         <p>[]a</p>
-     *     </li>
-     * </ul>
-     * ```
-     *
-     * The ancestors list is `['ul', 'li', 'p']`.
-     *
-     * The 4 command `commandUlP`, `commandP`, `commandUl`, `commandAny` could
-     * match the `command` identifier but not `commandImage`.
-     *
-     * The priority is calculated with "`lvl2`,`lvl1`".
-     * - `commandUlP` specificity: 2,2 `lvl2`: 2 means the last predicate
-     *   (`isP`) is found at depth 2 in ancestors `lvl1`: 2 means there are 2
-     *   predicates in `commandUlP`'s selector
-     * - `commandP` specificity: 2,1 `lvl2`: 2 means the last predicate (`isP`)
-     *   is found at depth 2 in ancestors `lvl1`: 1 means there is 1 predicate
-     *   in `commandP`'s selector
-     * - `commandUl` specificity: 0,1 `lvl0`: 0 means the last predicate
-     *   (`isUl`) is found at depth 0 in ancestors `lvl1`: 1 means there is 1
-     *   predicate in `commandP`'s selector
-     * - `commandAny` specificity: -1,0 No selector means no specificity.
-     *   `lvl0`: -1 means it has no depth.  `lvl1`: 0 means there is 0
-     *   predicates in `commandAny`'s selector
-     *
-     * The result will be `commandUlP` because it has the highest specificity.
+     * Test all contextuals against a hierarchy of VNodes and return the result
+     * ordered by specificity.
+     */
+    static getRankedMatches<T>(
+        hierarchy: VNode[],
+        contextuals: ContextualEntry<T>[],
+    ): RankedResult<T>[] {
+        const matches: RankedResult<T>[] = [];
+
+        for (let index = 0; index < contextuals.length; index++) {
+            const contextual = contextuals[index];
+            const match = ContextManager._matchNodes(hierarchy, contextual.selector);
+
+            if (match) {
+                matches.push({
+                    lvl1Score: match[0],
+                    lvl2Score: match[1].length,
+                    matched: match[1],
+                    index: index,
+                    entry: contextual,
+                });
+            }
+        }
+
+        // Sort the matches:
+        // - from highest to lowest score
+        // - when the score is the same, from highest to lowest index
+        const rankedMatch = matches.sort(function(a, b) {
+            if (b.lvl1Score === a.lvl1Score && b.lvl2Score === a.lvl2Score) {
+                return b.index - a.index;
+            } else if (b.lvl1Score === a.lvl1Score) {
+                return b.lvl2Score - a.lvl2Score;
+            } else {
+                return b.lvl1Score - a.lvl1Score;
+            }
+        });
+        return rankedMatch;
+    }
+    /**
+     * Match items selector depending on the editor current context range and
+     * return the most specific item.
      *
      * @param items
      * @param paramsContext
      */
     match<T extends Contextual>(items: T[], paramsContext?: Context): [T, Context] {
-        let maxFirstMatchDepth = -1;
-        let maxSelectorLength = 0;
-        let matchingItem: T;
-        let matchingContext: Context;
-
-        for (const item of items) {
-            const context = { ...this.defaultContext, ...item.context, ...paramsContext };
-            let firstMatchDepth = -1;
-            let index = 0;
-            const selector = item.selector || [];
-            const matches: VNode[] = [];
-            const start = context.range.start;
-            const nodes = start.ancestors();
-            const node = start.previousSibling() || start.nextSibling();
-            if (node) {
-                nodes.unshift(node);
-            }
-            const maximumDepth = nodes.length - 1;
-            for (const predicate of [...selector].reverse()) {
-                let matchFound = false;
-                while (!matchFound && index < nodes.length) {
-                    if (nodes[index].test(predicate)) {
-                        matchFound = true;
-                        matches.unshift(nodes[index]);
-                        if (firstMatchDepth === -1) {
-                            // Deeper match has higher specificity. So lower
-                            // index in ancestors, means higher specificity.
-                            firstMatchDepth = maximumDepth - index;
-                        }
-                    }
-                    index++;
-                }
-                // Stop checking the predicates of this particular command
-                // since at least one of them don't match the context.
-                if (!matchFound) break;
-            }
-
-            if (
-                matches.length === selector.length &&
-                maxFirstMatchDepth <= firstMatchDepth &&
-                maxSelectorLength <= selector.length &&
-                (!item.check || item.check({ ...context, selector: matches }))
-            ) {
-                maxFirstMatchDepth = firstMatchDepth;
-                maxSelectorLength = selector.length;
-                matchingItem = item;
-                matchingContext = context;
-            }
+        const context = { ...this.defaultContext, ...paramsContext };
+        const start = context.range.start;
+        const hierarchy = start.ancestors();
+        const node = start.previousSibling() || start.nextSibling();
+        if (node) {
+            hierarchy.unshift(node);
         }
-        return [matchingItem, matchingContext];
+        const entries: ContextualEntry<T>[] = items.map(item => {
+            const entry: ContextualEntry<T> = {
+                selector: item.selector || [],
+                value: item,
+            };
+            return entry;
+        });
+
+        const matches = ContextManager.getRankedMatches(hierarchy, entries);
+        const match = matches.find(match => {
+            return (
+                !match.entry.value.check ||
+                match.entry.value.check({ ...context, selector: match.matched })
+            );
+        });
+        return [match?.entry.value, context];
+    }
+
+    //--------------------------------------------------------------------------
+    // Private
+    //--------------------------------------------------------------------------
+
+    /**
+     * Check whether a hierarchy of `VNode` match with `selector`.
+     *
+     * Return a tuple with the first value being level of specificity lvl2 and
+     * the second value being the VNnode that matched with a selector
+     * (specificity lvl1).
+     */
+    static _matchNodes(hierarchy: VNode[], selector: Predicate[]): [number, VNode[]] | undefined {
+        const matches: VNode[] = [];
+        const maximumDepth = hierarchy.length - 1;
+        let firstMatchDepth = -1;
+        let index = 0;
+        for (const predicate of [...selector].reverse()) {
+            let matchFound = false;
+            while (!matchFound && index < hierarchy.length) {
+                if (hierarchy[index].test(predicate)) {
+                    matchFound = true;
+                    matches.unshift(hierarchy[index]);
+                    if (firstMatchDepth === -1) {
+                        // Deeper match has higher specificity. So lower
+                        // index in ancestors, means higher specificity.
+                        firstMatchDepth = maximumDepth - index;
+                    }
+                }
+                index++;
+            }
+            // Stop checking the predicates of this particular command
+            // since at least one of them don't match the context.
+            if (!matchFound) break;
+        }
+        return matches.length === selector.length && [firstMatchDepth, matches];
     }
 }
--- a/packages/core/src/Core.ts
+++ b/packages/core/src/Core.ts
@@ -77,11 +77,7 @@ export class Core<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
         if (range.isCollapsed()) {
             // Basic case: remove the node directly preceding the range.
             const previousSibling = range.start.previousSibling();
-            if (
-                previousSibling &&
-                range.startContainer.breakable &&
-                range.startContainer.editable
-            ) {
+            if (previousSibling && range.startContainer.editable && previousSibling.editable) {
                 previousSibling.removeBackward();
             } else if (range.startContainer.breakable && range.startContainer.editable) {
                 // Otherwise set range start at previous valid leaf.
@@ -109,7 +105,7 @@ export class Core<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
         if (range.isCollapsed()) {
             // Basic case: remove the node directly following the range.
             const nextSibling = range.end.nextSibling();
-            if (nextSibling && range.endContainer.breakable && range.endContainer.editable) {
+            if (nextSibling && range.endContainer.editable && nextSibling.editable) {
                 nextSibling.removeForward();
             } else if (range.endContainer.breakable && range.endContainer.editable) {
                 // Otherwise set range end at next valid leaf.
--- a/packages/core/src/Core.ts
+++ b/packages/core/src/Core.ts
@@ -1,8 +1,9 @@
-import {JWPlugin, JWPluginConfig} from './JWPlugin';
+import { JWPlugin, JWPluginConfig } from './JWPlugin';
 import JWEditor from './JWEditor';
-import {CommandParams} from './Dispatcher';
-import {VSelectionDescription} from './VSelection';
-import {VNode, RelativePosition} from './VNodes/VNode';
+import { CommandParams } from './Dispatcher';
+import { VSelectionDescription } from './VSelection';
+import { VNode, RelativePosition } from './VNodes/VNode';
+import { RuleProperty } from './Mode';
 
 export type InsertParagraphBreakParams = CommandParams;
 export type DeleteBackwardParams = CommandParams;
@@ -35,6 +36,7 @@ export class Core<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
             handler: this.deleteForward,
         },
     };
+    mode = this.editor.mode;
 
     //--------------------------------------------------------------------------
     // Public
@@ -49,7 +51,7 @@ export class Core<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
         if (!range.isCollapsed()) {
             range.empty();
         }
-        if (range.startContainer.breakable) {
+        if (this.mode.is(range.startContainer, RuleProperty.BREAKABLE)) {
             range.startContainer.splitAt(range.start);
         } else {
             // Use a separator to break paragraphs in an unbreakable.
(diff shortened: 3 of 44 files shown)


packages/plugin-dom-editable/src/EventNormalizer.ts
```diff
@@ -1611,9 +1630,10 @@ export class EventNormalizer {
         // Don't trigger events on the editable if the click was done outside of
         // the editable itself or on something else than an element.
         const target = this._getEventTarget(ev);
-        if (target && this._isInEditable(target)) {
+        const caretPosition = this._getEventCaretPosition(ev);
+        if (target && caretPosition && this._isInEditable(caretPosition.offsetNode)) {
```
Goaman, line 1634: caretPosition will always be an object so no need to check it with `&& caretPosition`

packages/plugin-dom-editable/src/EventNormalizer.ts
```diff
@@ -1644,6 +1664,23 @@ export class EventNormalizer {
...
+                    selection.anchorNode === selection.focusNode &&
+                    selection.anchorOffset === selection.focusOffset;
+                const target = this._getElement(selection.focusNode);
+                if (collapsed && !!target.closest('[contentEditable=true]')) {
+                    document.getSelection().removeAllRanges();
+                }
+                return {
+                    actions: [],
+                };
+            });
```
Goaman, line 1683: Here is a problem as mentioned in a comment above.
The normalizer is not meant to correct the DOM. Only normalizing event of the DOM editable.
So `document.getSelection().removeAllRanges();` is forbidden in the Normalizer.
The Normalizer should send the event, our vDocument correct the change if necessary.

Do you understand why do we use the `Timeout` mechanism?
In this particular case, it is useless to use the `Timeout` class as it reduce to a call to a simple `setTimeout(fn, 0)`.

Even worse, the selection can be removed at the wrong time because of it's "random" call time (dependent of the system performance).
And it will never send the EventBatch as you are not calling `this._triggerEventBatch`.

packages/core/src/JWEditor.ts
```diff
@@ -333,6 +334,24 @@ export class JWEditor {
...
+    }
+
+    /**
+     * Return true if the given node is breakable.
+     *
+     * @param node
+     */
+    isBreakable(node: VNode): boolean {
+        return this.mode ? this.plugins.get(Mode).isBreakable(node) : node.breakable;
+    }
```
Goaman: Verry strange to have that kind of function in the editor itself. It ties the mode to the editor even more.
It's another argument in favor to pass only the mode to the objects that need mode functionnality.

We should avoid to make "proxy" method like that on `JWEditor`.

Also, I believe that `this.mode` should never be undefined. There should always have a mode that is active. Otherwise it will require us to make these kind of branching.

Keep in mind that in the future, the mode could set any kind of properites for the node.

packages/core/src/Core.ts
```diff
@@ -49,7 +49,7 @@ export class Core<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
         if (!range.isCollapsed()) {
             range.empty();
         }
-        if (range.startContainer.breakable) {
+        if (this.editor.isBreakable(range.startContainer)) {
```
Goaman, line 52: `this.mode.isBreakable` :D

packages/core/src/VNodes/AbstractNode.ts
```diff
@@ -13,17 +14,29 @@ export interface AbstractNodeParams {
...
+    get editable(): boolean {
+        const attributes = this.modifiers.find(Attributes);
+        const editable = attributes?.get('contenteditable');
+        return typeof this._editable === 'boolean'
+            ? this._editable
+            : [undefined, 'true', ''].includes(editable);
+    }
+    set editable(state: boolean) {
+        this._editable = state;
+    }
```
Goaman, line 31: I don't think the logic should be in the node itself.
The node in the vdocument should not be aware of a contenteditable attribute that is only present in the DOM.
Another argument: It does not make sense for an `InlineNode`.

I would suggest that the mode `isEditable` would perform that logic.

I also suggest that the `mode.isPropertyX(node)` is always called anywhere in the code (which argue in favor of a mode being always present in the `JWEditor`).
And also make the code more modular, we could imagine a mode that do not care about "contenteditable" because it does know about any DOM concept.

The property of a node should be readonly as it describe how a node is defined but should not be manipulated to avoid confusion.

packages/plugin-align/src/Align.ts
```diff
@@ -103,11 +103,11 @@ export class Align<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T
      */
     align(params: AlignParams): void {
         const nodes = params.context.range.targetedNodes().flatMap(node => {
-            if (node.editable) {
+            if (this.editor.isEditable(node)) {
```
Goaman, line 106: Another example of why it would be beneficial to think in terms of mode, rather than editor that check if something is editable:
We could introduce `params.ctx.mode.isEditable(node)` even though I can't think of any usefull case right now.

packages/plugin-dom-layout/src/DomLayout.ts
```diff
@@ -166,7 +167,12 @@ export class DomLayout<T extends DomLayoutConfig = DomLayoutConfig> extends JWPl
         // We cannot always expect a 'contentEditable' attribute on the main
         // ancestor. So we expect to find the main editor ZoneNode if we are in
         // the editable part of Jabberwock.
-        return node?.ancestor(ZoneNode).managedZones.includes('main') || false;
+        return (
+            (node &&
+                (!this.editor.mode || this.editor.plugins.get(Mode).isEditable(node)) &&
```
Goaman, line 172: This branching would not be require is the editor would always have a mode.

packages/core/src/Core.ts
Zinston: Uuuuh... no :shrug:

packages/core/src/JWEditor.ts
Zinston: Well I'd argue that it's the _editor_'s mode so to me it makes sense.

I removed your default mode because that was the reason you added a `checkEditable` key in the mode, so we could use the `editable` key of the node. That means we'd have to do that for every property that the mode could set...
But come to think of it, that `checkEditable` could maybe be replaced with a rule, like `{ selectors: [ node => node.editable ], editable: true }`.

So in conclusion I guess I could pass the `Mode` plugin wherever it's needed and have a default mode that just has the line above. Then I could remove these two functions.

@dmo-odoo Do you also think that would be better?

packages/core/src/VNodes/AbstractNode.ts
Zinston: I absolutely agree. You actually are the one who wrote this code :-D
But this is the part that I struggled with and finally gave up for the moment, hence the todo in the commit comment.

packages/plugin-align/src/Align.ts
Zinston: lol

packages/plugin-dom-layout/src/DomLayout.ts
Zinston: It would mostly not be required if I had managed to do what I left as a todo. That is precisely the line that is concerned with that todo.

packages/core/src/VRange.ts
```diff
@@ -76,6 +76,10 @@ export class VRange {
         if (boundaryPoints) {
             const [start, end] = boundaryPoints;
             const [startNode, startPosition] = start;
+
+            // Prevent setting a collapsed range into a not-editable node.
+            if (start === end && !startNode.editable) return;
```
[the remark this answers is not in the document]
Goaman: But a range does not necessarily comes from the selection of the editor.
A range can be created from a command, wich does not care about the selection.
And that's another argument to have the range having a `mode` rather than the `editor`.

packages/plugin-dom-editable/src/EventNormalizer.ts
Goaman: You are using the `Timeout` class but not it's usage. You can read the comment above `_pointerSelectionTimeout` to grasp the reason of it's existance.
The mechanism is to:
- make the function timeout to wait for the selection to be correct (you did this properly)
- save the callback that need to be called
- in case performance issue arise, the stack of event can be too big, manually trigger the callback at the right time (otherwise it would be called much later and the selection will be wrong)
- if the callback has been called properly, it will not trigger the timeout

So use the same mechanics, you should do something like this:
```typescript
this._pointerSelectionTimeout = new Timeout<EventBatch>(() => {
                const selection = this._getSelection();
                const collapsed =
                    selection.anchorNode === selection.focusNode &&
                    selection.anchorOffset === selection.focusOffset;
                const target = this._getElement(selection.focusNode);
                if (collapsed && !!target.closest('[contentEditable=true]')) {
                    document.getSelection().removeAllRanges();
                }
                return {
                    actions: [],
                };
});
this._triggerEventBatch(this._pointerSelectionTimeout.promise);
```

Without that the sequence of event when the system lags:
key: callback1
click: callback2
key: callback3

would be interpreted as
key: callback1
key: callback3
click: callback2

Hope it make sense to you.

packages/core/src/JWEditor.ts
Goaman: I forgot to tell in the previous comments, the mode class should not be a plugin, it should be the class provided by the `ModePlugin` (I don't know how to name thoses because of the current convention of plugin naming).

As discussed with @dmo-odoo, you could have multiples range in the dom that are not in the same mode. (e.g. collaboration).

I don't understand your argument about adding a `checkEditable` for all keys as you do it with `isEditable`, `isBreakable`.

packages/core/src/Core.ts
Goaman: Actually it make more sense. You only need the mode, not the editor. the `editor.isBreakable` is calling the `mode.isBreakable`...

packages/core/src/VNodes/AbstractNode.ts
Goaman: Haha, it was a wip

packages/core/src/Core.ts
Zinston: No, `editor.isBreakable` calls `editor.plugins.get(Mode).isBreakable`.
In any case I already mentioned that I was open to having a default mode and therefore handling these through the mode every time.

packages/core/src/JWEditor.ts
Zinston: I changed these things from your first draft for a reason. I renamed `ModePlugin` into `Mode` because of said conventions of plugin naming. Then what you named `Mode` was replaced with `ModeDescription` which is enough to describe a mode. A class is overkill imo as it's purely declarative.

The problem with `checkEditable` is that it's a boolean that informs the plugin of whether it should ignore the mode for a given property of not. `isEditable`, on the other hand, implements a real feature. And as you mentioned above the logic is duplicated between `isEditable` and `isBreakable` so technically I could perfectly do something like `is(node, 'editable')`, `is(node, 'breakable')`, `is(node, 'tangible')` etc. instead.

In any case I countered my own argument just below it when I said I _could_ have a default mode with a `{ selectors: [ node => node.editable ], editable: true }` rule to replace `checkEditable: false` so... :shrug:

packages/core/src/Core.ts
dmo-odoo: This echoes to one of my first comments about either supporting mode at the editor level or accessing it from the plugin rather than "hiding" it thought a dedicated api on the editor.

packages/core/src/JWEditor.ts
dmo-odoo: It's funny because it *seems* we yet again have three different views on this. @Zinston obviously likes the current version since he considers he's done with this PR. @Goaman wants the `Mode` to stay in a plugin and pass the mode rather than the editor (I guess so because you mentionned a `ModePlugin`). And I think I would put the `Mode` in the core of the editor and stop pretending it's a plugin since it is used everywhere and most notably in `VRange`.

packages/core/src/VNodes/AbstractNode.ts
```diff
@@ -13,19 +14,31 @@ export interface AbstractNodeParams {
 let id = 0;
 export abstract class AbstractNode extends EventMixin {
     readonly id = id;
+    _editable: boolean = null;
```
dmo-odoo, line 17: 👀

packages/core/src/VNodes/AbstractNode.ts
dmo-odoo: Is this still used then ? Removing this would make me happy. 🧸

packages/plugin-align/src/Align.ts
dmo-odoo: He's right, actually. Testing whether a node is editable or not through the editor itself (or a plugin, really) considers that only one mode can be active at a time, which might be wrong. Different users might have different modes. A subcommand of a complex command might want to modify the structure in the context of a particular mode.

packages/plugin-align/test/Align.test.ts
```diff
@@ -33,10 +34,13 @@ describePlugin(Align, testEditor => {
                         const domEngine = editor.plugins.get(Layout).engines.dom;
                         const editable = domEngine.components.get('editable')[0];
                         const root = editable;
-                        root.lastChild().editable = false;
+                        root.lastChild()
+                            .modifiers.get(Attributes)
+                            .set('contenteditable', 'false');
```
dmo-odoo: Et quoi on ne sait plus faire autrement alors ? C'est nul de devoir écrire sur un attribut en particulier... :(

packages/plugin-dom-layout/src/DomLayout.ts
dmo-odoo: I see no TODO.

packages/bundle-odoo-website-editor/OdooWebsiteEditor.ts
```diff
@@ -207,6 +208,12 @@ export class OdooWebsiteEditor extends JWEditor {
             autoFocus: true,
             source: options.source.firstElementChild as HTMLElement,
         });
+        if (options.mode) {
+            this.load({
+                modes: [options.mode],
+            });
+            this.configure({ mode: options.mode.id });
+        }
```
dmo-odoo, line 232: Je ne comprends pas ce commit. Je crois que je comprends le principe, c'est d'utiliser le mode comme une "généralisation" de ce qu'on essayait de faire ici avec l'option de Divider, mais je ne comprends pas où est le nouveau code qui dit que les div ne sont pas breakable.

karma.conf.js
```diff
@@ -81,15 +82,15 @@ module.exports = function(config) {
         frameworks: ['mocha', 'chai', 'sinon'],
 
         // list of files / patterns to load in the browser
-        files: config.includeFiles ? config.includeFiles.split(',') : ['packages/**/test/**/*.ts'],
+        files: ['test_index.ts'],
```
dmo-odoo, line 85: If I'm not mistaken this is already merged in master.

packages/plugin-dom-editable/src/EventNormalizer.ts
```diff
@@ -1556,7 +1556,7 @@ export class EventNormalizer {
         const x = ev instanceof MouseEvent ? ev.clientX : ev.touches[0].clientX;
         const y = ev instanceof MouseEvent ? ev.clientY : ev.touches[0].clientY;
         let caretPosition = caretPositionFromPoint(x, y);
-        if (!caretPosition || !this._isInEditable(caretPosition.offsetNode)) {
+        if (!caretPosition) {
```
dmo-odoo, line 1559: *You* are excessive.

packages/core/src/VRange.ts
```diff
@@ -95,6 +97,13 @@ export class VRange {
     // Public
     //--------------------------------------------------------------------------
 
+    get mode(): Mode {
+        if (!this._mode) {
+            this._mode = this.editor.plugins.get(Mode);
+        }
+        return this._mode;
+    }
+
```
dmo-odoo, line 106: Pourquoi lui il a besoin d'un have avec `mode` et `_mode` alors que les autres ça marche très bien de faire `mode = this.editor.plugins.get(Mode)` et c'est tout ?

packages/plugin-dom-layout/src/DomLayout.ts
```diff
@@ -169,7 +169,7 @@ export class DomLayout<T extends DomLayoutConfig = DomLayoutConfig> extends JWPl
         // the editable part of Jabberwock.
         return (
             (node &&
-                (!this.editor.mode || this.editor.plugins.get(Mode).isEditable(node)) &&
+                (node.contentEditable || this.editor.plugins.get(Mode).isEditable(node)) &&
```
dmo-odoo: `node.contentEditable`?

packages/plugin-align/test/Align.test.ts
```diff
@@ -39,8 +39,7 @@ describePlugin(Align, testEditor => {
                             .set('contenteditable', 'false');
                         return align(AlignType.LEFT)(editor);
                     },
-                    contentAfter:
-                        '<p>ab</p><p contenteditable="false"><span style="text-align: left;">c[]d</span></p>',
+                    contentAfter: '<p>ab</p><p contenteditable="false">c[]d</p>',
```
dmo-odoo, line 39: Attends mais les tests d'avant ils étaient ouf si c'était pas éditable et qu'on faisait align et que ça éditait !?

packages/core/src/VNodes/AbstractNode.ts
```diff
@@ -16,21 +16,39 @@ export abstract class AbstractNode extends EventMixin {
...
+        if (typeof this._editable === 'boolean') {
+            return this._editable;
+        } else if (this.modifiers.find(Attributes)?.has('contentEditable')) {
+            // TODO: unbind from `Attributes`.
+            return this.contentEditable;
+        } else if (this.parent) {
+            return this.parent.contentEditable || this.parent.editable || false;
+        } else {
+            return true;
+        }
```
dmo-odoo, line 34: Aie tu construis sur le diff que je voulais supprimer et que tu disais que c'était pourri mais que tu l'avais laissé parce que c'était le truc de nicolas :/

packages/core/src/VNodes/AbstractNode.ts
```diff
@@ -16,21 +16,39 @@ export abstract class AbstractNode extends EventMixin {
...
+     *
+     * TODO: unbind from `Attributes`.
+     */
+    get contentEditable(): boolean {
+        const editable = this.modifiers.find(Attributes)?.get('contentEditable');
+        return editable === '' || editable?.toLowerCase() === 'true' || false;
+    }
+    set contentEditable(state: boolean) {
+        this.modifiers.get(Attributes).set('contenteditable', '' + state);
+    }
```
dmo-odoo, line 51: Attends ça devient ouf là. Déjà y'a `mode.isEditable(node)`, ce qui me semble être le bon truc. Après y'a `node.editable`, que perso je voudrais supprimer parce que si on a le mode bah c'est le mode qui décide si un noeud est éditable, pas le noeud lui-même, autrement franchement qui en dehors de nous va penser à utiliser le mode pour checker qu'un noeud est editable si il voit qu'il existe une propriété qui s'appelle `editable` ? Et alors *en plus* y'a aussi `node.contentEditable` pour rajouter encore une couche de confusion parce que c'est l'attribut "officiel" du DOM qui contrôle ce que nous on a déjà dans les deux autres que j'ai cité avant. Pour moi tout doit disparaitre sauf le mode. Après y'a surement un besoin que j'ai pas compris mais bon, je suis curieux de comprendre... :p

packages/plugin-align/test/Align.test.ts
```diff
@@ -34,9 +33,7 @@ describePlugin(Align, testEditor => {
                         const domEngine = editor.plugins.get(Layout).engines.dom;
                         const editable = domEngine.components.get('editable')[0];
                         const root = editable;
-                        root.lastChild()
-                            .modifiers.get(Attributes)
-                            .set('contenteditable', 'false');
+                        root.lastChild().contentEditable = false;
```
dmo-odoo, line 36: Et franchement si la prop magique `contentEditable` sert juste à faire ça on peut la supprimer sans remords.

packages/plugin-mode/src/Mode.ts
```diff
@@ -49,50 +49,21 @@ export class Mode<T extends ModeConfig = ModeConfig> extends JWPlugin<T> {
...
-            return result[0].entry.data;
-        }
-    }
-    /**
-     * Return true if the given node is breakable.
-     *
-     * @param node
-     */
-    isBreakable(node: VNode): boolean {
+    is(node: VNode, property: 'editable' | 'breakable'): boolean {
```
dmo-odoo, line 52: Au début quand j'ai vu ce truc je me suis dit "oulalalalah il est reparti avec ses paramètres en string". Puis j'ai vu qu'en fait ça se "combine" pas trop mal avec le fait que c'est le nom d'une property sur les rules. Du coup je me demande si on pourrait pas leverage ce concept à fond et du coup retirer la property `data` que je trouve affreuse par du coup le nom de la property et la rule qui du coup serait typée du style :
```ts
export type Rule {
    selector: Predicate[];
    [key: string]: boolean
}
```
Le problème c'est qu'un type comme ça TypeScript le refuse parce qu'il va dire que `selector` ne matche pas l'index signature puisque lui c'est pas un boolean. Je sais plus comment on fait pour faire ça mais je pense qu'il y a moyen. Genre ça p-e (names are wip):
```ts
type Property: string;
type Ruling: Record<Property, boolean>;
type Rule = Omit<Ruling, 'selector'> & {
    selector: Predicate[];
}
```
Et du coup le type du paramètre de `is` ce serait `Property`. Après je suis pas sûr que ça marche mon truc avec `Omit`, mais j'imagine qu'il y a bien une façon de typer un record en disant qu'il y a *une* propriété particulière qui n'est pas du même type.

packages/bundle-odoo-website-editor/OdooWebsiteEditor.ts
```diff
@@ -57,10 +57,12 @@ interface OdooWebsiteEditorOptions {
     snippetMenuElement?: HTMLElement;
     snippetManipulators?: HTMLElement;
     template?: string;
+    toolbar?: boolean;
```
dmo-odoo, line 60: ? Ah c'est parce qu'il n'y a pas de toolbar dans translate c'est ça ? Tu devrais pas faire une classe qui étant `OdooWebsiteEditor` ? Genre `OdooTranslationEditor` qui écraserait simplement la définition de toolbar de `OdooWebsiteEditor` ? Par ce que bon là tu mets `toolbar:false` mais `discardButton: true` et tu end up quand même avec une toolbar alors c'est un peu bizarre cette option `toolbar` tu vois ce que je veux dire ? :p

packages/bundle-odoo-website-editor/OdooWebsiteEditor.ts
```diff
@@ -57,10 +57,12 @@ interface OdooWebsiteEditorOptions {
...
     mode?: ModeDefinition;
     // todo: Remove when configuring the toolbar in another way.
     discardButton?: boolean;
     saveButton?: boolean;
+    plugins?: [typeof JWPlugin, JWPluginConfig?][];
```
dmo-odoo, line 65: ?

packages/plugin-xml/src/Attributes.ts
```diff
@@ -77,7 +77,7 @@ export class Attributes extends Modifier {
      * @param key
      */
     has(key: string): boolean {
-        return this.keys().includes(key);
+        return this.keys().includes(key.toLowerCase());
```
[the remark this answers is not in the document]
Zinston: :shrug:

packages/core/src/ContextManager.ts
```diff
@@ -18,6 +18,33 @@ export interface Contextual {
     check?: (context: CheckingContext) => boolean;
 }
 
+export interface ContextualEntry<T> extends Contextual {
+    selector: Predicate[];
+    // Original item
+    data: T;
```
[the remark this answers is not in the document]
Zinston: image

packages/plugin-dom-editable/src/EventNormalizer.ts
```diff
@@ -610,11 +617,25 @@ export class EventNormalizer {
...
+                        newDomSelection.anchorOffset === newDomSelection.focusOffset;
+
+                    let domSelection;
+                    if (collapsed && !this._isInEditable(newDomSelection.anchorNode)) {
+                        // Prevent a collapsed selection outside the editable
+                        // context: keep the current selection.
+                        domSelection = currentDomSelection;
+                    } else {
+                        domSelection = newDomSelection;
+                    }
```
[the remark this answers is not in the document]
Goaman: I am not satisfied.

If a node is not editable and the normalizer send an action for that node (like setSelection), it is the role of the editor to discard the normalized event, not the normalizer itself.
The normalizer normalize a contenteditable without any knowlede about any other structure.

packages/plugin-dom-editable/src/EventNormalizer.ts
Goaman: I don't get it. If someone click on a target that is in contenteditable:
- the normaliser send "setSelection" to the editor
- the editor process that the selection is not in an editable
- the editor discard the selection
- the editor rerender the selection

packages/core/src/VNodes/AbstractNode.ts
Goaman: I agree that the naming is confusing. But here is the definition of contentEditable by @Zinston 
> contentEditable indique que ses enfants sont éditables, pas lui-même. Par contre il peut triggerer des events.

I also agree with the rest of the comment.

packages/plugin-mode/src/Mode.ts
Goaman: Je suis suis contre un string sur le `Rule`. Le `data` est fait exprès pour être générique.
D’ailleurs la généricité est déjà utilisée 2x avec `getRankedMatches` 

Je suis aussi contre les paramètres en string. C'est nul pour la recherche dans le code.

packages/core/src/ContextManager.ts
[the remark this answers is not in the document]
Goaman: Keep my `data` alive.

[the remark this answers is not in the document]
Goaman: Keep my `data` alive.

I explained in another comment the reason of it's existance. Genericity.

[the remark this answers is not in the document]
Zinston: Can you think of a more descriptive name?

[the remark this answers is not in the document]
Goaman: I've tried and I could not come with a more descriptive name as it is *generic* data.
It's like an ajax request, you receive `data` that could be anything.
Sometimes a property `data` makes sense and I believe it is the best name in the current situation.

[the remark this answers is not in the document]
Zinston: Actually after looking at how it's used, maybe we should just call it `item` instead. It's not much less vague though.

packages/core/src/VRange.ts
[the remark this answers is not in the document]
Zinston: Well this was an extra safety as it _is_ handled in the `EventNormalizer` so no problem.

packages/plugin-color/src/Color.ts
```diff
@@ -54,32 +54,18 @@ export class Color<T extends ColorConfig = ColorConfig> extends JWPlugin<T> {
...
-                // descendants.
-                let parent = node.parent;
-                while (parent && parent.editable && this._isAllColored(parent, color)) {
-                    parent.modifiers.get(Attributes).style.set(this.styleName, color);
-                    // TODO: not remove the children's styles when we have modifiers.
-                    for (const child of parent.children()) {
-                        child.modifiers.find(Attributes)?.style.remove(this.styleName);
-                    }
-                    parent = parent.parent;
-                }
```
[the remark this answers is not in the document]
Zinston: Nope :-)

packages/plugin-dom-editable/src/EventNormalizer.ts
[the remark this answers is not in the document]
Zinston: I merely _corrected_ the current behavior which _did_ already prevent the action right from the `EventNormalizer` through `isInEditable` checks, but did it wrong in some cases.
I remember this is an old debate we had like 6 months ago. I think chm had introduced those checks so that the browser would revert to default outside the editable zones.
If we want to _change_ the current behavior I'm not necessarily opposed to it but it's definitely out of scope, especially in the current rush.

packages/plugin-dom-editable/src/EventNormalizer.ts
Zinston: I find that risky and out of scope. I'm open to discussion though for sure but I don't consider this a priority in the current state of affairs.

packages/plugin-dom-editable/src/EventNormalizer.ts
```diff
@@ -1665,6 +1702,14 @@ export class EventNormalizer {
         }
         return eventBatch;
     }
+    /**
+     * Return a node's parent if it's not an instance of `Element`.
+     *
+     * @param node
+     */
+    _getElement(node: Node): Element {
```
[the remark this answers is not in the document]
Zinston: It is.

packages/plugin-dom-layout/src/DomLayout.ts
```diff
@@ -147,6 +147,12 @@ export class DomLayout<T extends DomLayoutConfig = DomLayoutConfig> extends JWPl
     isInEditable(target: Node): boolean {
         const layout = this.dependencies.get(Layout);
         const domLayoutEngine = layout.engines.dom as DomLayoutEngine;
+        const selection = target.ownerDocument?.getSelection();
+        const anchorNode = selection?.anchorNode;
+        const isAnchorDescendantOfTarget = this._isDescendantOf(anchorNode, target);
+        if (isAnchorDescendantOfTarget) {
+            target = anchorNode;
+        }
```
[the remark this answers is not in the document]
Zinston: That function of `EventNormalizer` depends on the event. We don't have that in this context.

packages/plugin-dom-layout/src/DomLayout.ts
```diff
@@ -174,4 +180,20 @@ export class DomLayout<T extends DomLayoutConfig = DomLayoutConfig> extends JWPl
...
+     * @param ancestor
+     */
+    private _isDescendantOf(node: Node, ancestor: Node): boolean {
+        while (node) {
+            if (node === ancestor) {
+                return true;
+            }
+            node = node.parentElement;
+        }
+        return false;
```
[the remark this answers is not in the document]
Zinston: I wish ^^'
`closest` doesn't exist on `Node` - only on `Element`.

packages/core/src/VNodes/AbstractNode.ts
Zinston: Sadly, no. I has evolved though.

packages/plugin-align/src/Align.ts
Zinston: We now have `params.context.range.mode.isEditable(node)` which is roughly equivalent as you can pass a mode to `withRange`.

packages/plugin-align/test/Align.test.ts
Zinston: Si si, l'autre moyen marche toujours. En fait Nicolas avait adapté ce test, je ne sais pas pourquoi, mais après réflexion je me suis dit que j'allais garder la modification quand même parce que le test avait plus d'intérêt comme ça vu qu'il testait le cas du DOM.

packages/plugin-dom-layout/src/DomLayout.ts
Zinston: Commit message:
> `TODO: if there is no mode, the `editable` property of the node should be used as default. This is not working properly as of yet.`

packages/plugin-dom-editable/src/EventNormalizer.ts
Zinston: image

packages/plugin-dom-layout/src/DomLayout.ts
Zinston:
```ts
/**
 * True if the node has the `contentEditable` attribute. This implies that
 * its children are editable but it is not necessarily itself editable.
 *
 * TODO: unbind from `Attributes`.
 */
get contentEditable(): boolean {
    const editable = this.modifiers.find(Attributes)?.get('contentEditable');
    return editable === '' || editable?.toLowerCase() === 'true' || false;
}
```

packages/plugin-align/test/Align.test.ts
Zinston: Oui c'était n'importe quoi :shrug:

packages/core/src/VNodes/AbstractNode.ts
Zinston: J'ai pas trouvé de meilleure solution, et franchement, c'est pas faute d'avoir cherché, essayé et galéré... Là ça marche enfin correctement... Si tu as mieux, je suis preneur :shrug:

packages/plugin-dom-editable/src/EventNormalizer.ts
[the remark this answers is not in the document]
Goaman: Ok.

packages/plugin-dom-editable/src/EventNormalizer.ts
Goaman: 1) Changing the logic about who revert the selection: can be done in another commit.
2) The timeout: cannot be skipped: use my code to trigger the event at the right moment.

packages/plugin-mode/src/Mode.ts
Zinston: Si c'est pour le grep, tu fais un `enum` et c'est bon...
```ts
export enum ModeProperty {
    EDITABLE = 'editable',
    BREAKABLE = 'breakable',
}
```
Quant à l'argument de l'évolution future, c'est probablement une bonne idée mais c'est très lointain et très abstrait actuellement donc je serais d'avis de ne pas trop s'en préoccuper pour le moment, chaque chose en son temps. Je préfère que ça prenne deux jours de plus dans deux ans qu'un jour de plus maintenant.

Goaman: D'accord mais on ne parle pas d'une journée juste pour une clé (qui existe déjà et qui fait bien le taff) ^^.

packages/bundle-odoo-website-editor/OdooWebsiteEditor.ts
Zinston: oui tu as raison

packages/bundle-odoo-website-editor/OdooWebsiteEditor.ts
Zinston: Côté Odoo, dans le mode.

packages/plugin-dom-editable/src/EventNormalizer.ts
Zinston: Done.

packages/plugin-align/test/Align.test.ts
dmo-odoo: Voilà justement un test similaire à ce qu'on discutait tout à l'heure où on vient mettre un range dans une zone qui est editable puis entre temps elle n'est plus editable puis on fait une commande dedans.

packages/plugin-mode/src/Mode.ts
dmo-odoo: Si j'ai bien compris @Goaman n'aime pas ma suggestion de "généraliser" le paramètre de la méthode `is` à tout le système de mode. Par contre je n'ai pas l'impression qu'il ait donné son avis sur le fait d'avoir un string dans la méthode `is` elle-même. Mais bon en réalité je pense que je connais son avis. cc @Zinston si tu veux défendre ton `is` ? Parce que j'ai pas trop compris votre histoire de deux jours plus tard vs un jour maintenant ou je sais pas quoi... :p

packages/core/src/Mode.ts
```diff
@@ -12,29 +12,57 @@ export interface ModeDefinition {
     rules: ModeRule[];
 }
 
+/**
+ * Properties of a `VNode` that can be modified by a mode.
+ */
+export type ModeProperty = 'editable' | 'breakable';
```
dmo-odoo: ...bah le voilà le truc que je suggérais dans l'autre thread de commentaire... ^^'

packages/core/src/Mode.ts
```diff
@@ -12,29 +12,57 @@ export interface ModeDefinition {
...
+const modeProperties = ['editable', 'breakable'];
```
dmo-odoo, line 22: Il n'existe (à ma connaissance) pas de moyen de faire que TypeScript lise les *valeurs* d'un tableau comme liste de types, du coup t'as dû écrire cette liste deux fois. C'est con hein ? C'est parce qu'ils disent (avec raison) que les valeurs du tableau peuvent changer plus tard même si c'est un const.

packages/core/src/Mode.ts
```diff
@@ -12,29 +12,57 @@ export interface ModeDefinition {
...
+
+type RulesPerProperty = Record<ModeProperty, ContextualEntry<boolean>[]>;
```
dmo-odoo, line 24: Perso j'aurais pas créé un alias pour ça justement pour éviter de devoir le nommer... ^^' Parce qu'en plus c'est pas des "rules" techniquement. Fin bon ce type de "ContextualEntry" aussi... En fait j'ai critiqué `data` mais `ContextualEntry` est louche aussi.

packages/core/src/Mode.ts
```diff
@@ -12,29 +12,57 @@ export interface ModeDefinition {
...
+
 export class Mode {
     id: ModeIdentifier;
     rules: ModeRule[];
+    private _rulesPerProperty: RulesPerProperty;
```
dmo-odoo, line 29: Du coup bah pareil ici, c'est techniquement pas des "rules"... :/

packages/core/src/Mode.ts
```diff
@@ -12,29 +12,57 @@ export interface ModeDefinition {
...
+                    const entry: ContextualEntry<boolean> = {
+                        selector: rule.selector,
+                        data: rule[property],
+                    };
+                    ruleEntries[property].push(entry);
+                }
+            }
+            return accumulator;
+        }, ruleEntries);
+    }
```
dmo-odoo, line 76: Perso j'aurais tapé ça dans le constructeur vu qu'on l'appelle qu'une seule fois avec simplement un commentaire au-dessus de la fonction. Tu sais c'est autorisé de faire plusieurs choses dans la même fonction si c'est bien segmenté et documenté... :p Au moins ça évite de faire trop de jeu de piste quand on multiple les fonctions auxiliaires. Ce qu'il faut éviter c'est quand on fait plusieurs choses en même temps de façon entremêlées et qu'on ne sait pas quelles lignes servent à quel truc parce qu'on fait tout en même temps. (Cela dit même dans ce cas là parfois ça peut se justifier pour des raisons de performances mais bon on n'en est pas là.)

packages/core/src/Mode.ts
```diff
@@ -12,29 +12,57 @@ export interface ModeDefinition {
...
+    //--------------------------------------------------------------------------
+
+    /**
+     * Return an object describing the mode's rules for each mode property.
+     */
+    _getRuleEntries(): RulesPerProperty {
+        const ruleEntries = modeProperties.reduce((accumulator, property) => {
+            accumulator[property] = [];
+            return accumulator;
+        }, {} as RulesPerProperty);
```
dmo-odoo, line 62: Ca marchait pas avec `const ruleEntries: RulesPerProperty` plutot que `as` ?

packages/core/src/Mode.ts
```diff
@@ -15,7 +15,10 @@ export interface ModeDefinition {
 /**
  * Properties of a `VNode` that can be modified by a mode.
  */
-export type ModeProperty = 'editable' | 'breakable';
+export enum ModeProperty {
+    EDITABLE = 'editable',
+    BREAKABLE = 'breakable',
+}
```
dmo-odoo, line 21: Ah bah il l'a fait. Si j'ai bien suivi l'autre conversation, nby disapproves. Ou alors il s'en fout tant que ça vient pas dans la définition même de `Rule` ? (Ouais parce que c'est `RuleProperty` en fait non ?) Comme dit dans l'autre thread de commentaire, j'ai pas trop compris ce qu'il voulait.

packages/core/src/Mode.ts
```diff
@@ -32,6 +32,12 @@ export class Mode {
         this.rules = mode.rules;
         this._rulesPerProperty = this._getRuleEntries();
     }
+    /**
+     * Return true if the given node's given property is true.
```
dmo-odoo, line 36: Je pense que c'est faux de dire que c'est une propriété *du noeud*. Il *se trouve* que les deux propriétés qu'on a se trouvent sur le noeud, mais en soi ce n'est pas *ça* qu'on teste puisque ça dépend du *mode* dans lequel on est plus que de la *propriété* sur le noeud lui-même.

packages/core/src/VNodes/AbstractNode.ts
```diff
@@ -26,29 +26,16 @@ export abstract class AbstractNode extends EventMixin {
...
-     *
-     * TODO: unbind from `Attributes`.
-     */
-    get contentEditable(): boolean {
-        const editable = this.modifiers.find(Attributes)?.get('contentEditable');
-        return editable === '' || editable?.toLowerCase() === 'true' || false;
-    }
-    set contentEditable(state: boolean) {
-        this.modifiers.get(Attributes).set('contenteditable', '' + state);
-    }
```
dmo-odoo: Can I get an *amen* ? 🎤

packages/utils/src/utils.ts
```diff
@@ -9,6 +12,16 @@ export type Constructor<T> = new (...args) => T;
...
 }
+/**
+ * True if the node has the `contentEditable` attribute. This implies that
+ * its children are editable but it is not necessarily itself editable.
+ *
+ * TODO: unbind from `Attributes`.
+ */
+export function isContentEditable(node: AbstractNode): boolean {
+    const editable = node.modifiers.find(Attributes)?.get('contentEditable');
+    return editable === '' || editable?.toLowerCase() === 'true' || false;
```
dmo-odoo, line 23: Si je me trompe pas le `|| false` est redondant.

packages/plugin-dom-editable/src/EventNormalizer.ts
```diff
@@ -1681,6 +1681,7 @@ export class EventNormalizer {
                     actions: [],
                 };
             });
+            this._triggerEventBatch(this._pointerSelectionTimeout.promise);
```
dmo-odoo, line 1684: cc @Goaman

packages/bundle-odoo-website-editor/OdooWebsiteEditor.ts
```diff
@@ -57,14 +57,34 @@ interface OdooWebsiteEditorOptions {
     snippetMenuElement?: HTMLElement;
     snippetManipulators?: HTMLElement;
     template?: string;
-    toolbar?: boolean;
+    layout?: ToolbarLayout;
```
dmo-odoo, line 60: You mean `toolbarLayout`, right ? Or whatever it's called in the `Toolbar` plugin configuration. My point is that `layout` in the context of the editor itself is already something else.

packages/plugin-dom-editable/src/EventNormalizer.ts
Goaman: It should work

packages/core/src/Mode.ts
Goaman: Oui je désaprouve a 100%. `is` est vraiment plus difficile a grepper. En plus on a deja le meme nom de methode sur `AbstractNode`.

En quoi c'est mieux d'avoir `is(ModeProperty.EDITABLE)` que de faire `isEditable`.

On y gagne pas grand chose non plus dans la definition car il faut une ligne par property.
Avec les méthodes, il faut une methode de une ligne (qu'on peut folder) par property.

packages/core/src/ContextManager.ts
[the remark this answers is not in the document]
Zinston: I don't care enough to fight ^^' @Goaman ok for you `value` instead of `data`?

packages/core/src/VNodes/AbstractNode.ts
Zinston: AMEN!

packages/utils/src/utils.ts
Zinston: `editable?` -> si y a pas d'editable, ça renverra `undefined`.

PR #240, opened 2020-07-07
[FIX] DomLayout: allow attribute with empty string

Goaman:


dmo-odoo:
I'm not sure this is right to end up with `target=""` when the user never set any target to begin with.

Goaman:
It is `LinkFormat` that has a ```constructor(url = '#', target = '')```.
Do I change `LinkFormat`?

dmo-odoo:
I would say yes but I'm curious for @Zinston's input here.

dmo-odoo:
@Goaman Considering @Zinston did not reply and won't be working for some days, I would say yes.

PR #249, opened 2020-07-14
[IMP] RendererDomObject: configure BR rendering

Goaman:
Odoo pr: https://github.com/odoo-dev/odoo/pull/317

I'm sure we will discuss about the name ;)

### DIFF
--- a/packages/core/src/Mode.ts
+++ b/packages/core/src/Mode.ts
@@ -18,8 +18,9 @@ export interface ModeDefinition {
 export enum RuleProperty {
     EDITABLE = 'editable',
     BREAKABLE = 'breakable',
+    ALLOW_EMPTY = 'allowEmpty',
 }
-const modeProperties = ['editable', 'breakable'];
+const modeProperties = ['editable', 'breakable', 'allowEmpty'];
 
 type RuleEntries = Partial<Record<RuleProperty, ContextualEntry<boolean>[]>>;
 
--- a/packages/bundle-odoo-website-editor/odoo-integration.ts
+++ b/packages/bundle-odoo-website-editor/odoo-integration.ts
@@ -17,6 +17,7 @@ import { Inline } from '../plugin-inline/src/Inline';
 import { DomHelpers } from '../plugin-dom-helpers/src/DomHelpers';
 import { OdooField } from '../plugin-odoo-field/src/OdooField';
 import { OdooFieldNode } from '../plugin-odoo-field/src/OdooFieldNode';
+import { VElement } from '../core/src/VNodes/VElement';
 
 export {
     OdooWebsiteEditor,
@@ -32,6 +33,7 @@ export {
     LinkFormat,
     Attributes,
     ContainerNode,
+    VElement,
     DividerNode,
     OdooStructureNode,
     OdooField,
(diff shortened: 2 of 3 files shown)


dmo-odoo:
@Zinston Faut trouver un mot pour "le fait de devoir avoir un br dans le rendu quand tu n'as pas d'enfants parce que sinon tu es invisible". On ne sait jamais, y'a pas un terme dans la spec w3c pour ce genre de trucs ? #whoknows

Zinston:
@dmo-odoo `usePlaceholderBr` ?
C'est pas un peu bizarre d'utiliser `Mode` pour configurer quelque chose de si propre au rendering DOM ? J'ai l'impression que ça a plus sa place dans la configuration du plugin qui déclare le renderer DOM.

packages/core/src/Mode.ts
```diff
@@ -18,8 +18,9 @@ export interface ModeDefinition {
 export enum RuleProperty {
     EDITABLE = 'editable',
     BREAKABLE = 'breakable',
+    RENDERBR = 'renderbr',
```
Zinston, line 21: How about `ALLOW_EMPTY` ?

Zinston: Although isn't it a little weird (although it sure _is_ convenient) to use `Mode` to configure something that is so tied to the DOM rendering? It seems like it should rather be done in the configuration of the plugin that declares the DOM renderer.

Goaman: +1 for `ALLOW_EMPTY`

I agree that the property is not likely to change between modes.
And indeed, the cascading expressivity that the ContextManager provide is nice.

We could re-use the ContextManager by creating a different logic for it, but I don't think it is necessary.

With that in mind, I believe we can think of the mode as a CSS sheet. The sheet can change or not but is a way to configure properties in a cascading fashion.

And the properties that can be configured can be any property that a plugin might need. Even a rendering plugin.
In the future, we could have differents rendering depending on the mode for a particular node.

dmo-odoo: I validate ALLOW_EMPTY. I suggest to keep the feature in Mode for now, even though I agree that it is a bit weird.

Goaman:
`ALLOW_EMPTY = 'allow_empty',` ?

packages/core/src/Mode.ts
Goaman: `ALLOW_EMPTY = 'allow_empty',` ?

dmo-odoo: I think `ALLOW_EMPTY = 'allowEmpty',` would be better considering @Zinston coded the `is` in such a way that, when the mode gives no information about a particular property, it looks for it on the node *in case* the node itself might define it. So since it *might* be possible to set this property on a `VNode` (though I don't see why we would), it makes sense to keep camelCasing it.

Goaman:
Done.
Attention, I also changed the condition 
```js
        } else if (!node.is(AtomicNode) && this.editor.mode.is(node, RuleProperty.ALLOW_EMPTY)) {
```
To
```js
        } else if (
            !node.is(AtomicNode) &&
            this.editor.mode.is(node, RuleProperty.ALLOW_EMPTY) !== false
        ) {
```
So, if `is(...)` return `undefined`, it means it's `true`.

dmo-odoo:
If `allowEmpty` is not set either by the mode or the node then it should be considered false and the node should have a BR.

dmo-odoo:
Shouldn't it be `this.editor.mode.is(node, RuleProperty.ALLOW_EMPTY) !== true` ?

Goaman:
No because if we didn't specify, `this.editor.mode.is(node, RuleProperty.ALLOW_EMPTY)` will return `undefined`.
If it is undefined, by default, we want to "allow empty".

You can test with `npm run test` both lines.

Goaman:
Ok, I missunderstood `allow_empty` with `render_br`

PR #258, opened 2020-07-23
[IMP] DomHelpers: remove class on all attributes

Goaman:


packages/plugin-dom-helpers/src/DomHelpers.ts
```diff
@@ -5,6 +5,23 @@ import { Attributes } from '../../plugin-xml/src/Attributes';
...
+        modifiers
+            .filter(modifier => modifier instanceof Attributes || modifier instanceof Format)
+            .map((modifier): Attributes[] => {
+                if (modifier instanceof Format) {
+                    return getAllAttributesModifiers(modifier.modifiers);
+                } else {
+                    return [modifier as Attributes];
+                }
+            }),
+    );
```
dmo-odoo, line 23: Wow, dans quel cas est-ce que tu parviens à avoir plusieurs modifiers `Attributes` sur le même noeud ?

Goaman: Never, why?
I typecast modifier in order to avoid to instanceof.
I could get this:
```
node
- modifier: Attribute
- modifier: Format
  - modifier: Attribute
- modifier: Format
  - modifier: Attribute
```
```

Goaman:
actually it's not working properly yet, I had a false positive

dmo-odoo:
As a general rule, I suggest you don't set me reviewer until the feature in the PR actually works... ;-)

Goaman:
Actually, it was a false - false positive, it's working. ^^'

packages/plugin-dom-helpers/src/DomHelpers.ts
dmo-odoo: Ok I understand, but then I'm surprised you have to do `as Attributes`. Couldn't you type the `modifier` attribute in the `map` as `Attributes | Format` ?

Goaman:
?

Goaman:
@dmo-odoo

dmo-odoo:
This PR is quite old. It was never merged so the issue it is supposed to solve either does not present itself anymore or can still be reproduced. Which one is it ?

Goaman:
Now I forgot which snippet exactly was concerned. Nonetheless, it is an intended behavior.
If you have one of the modifiers that have Attribute modifiers, you want to remove from them as well.

dmo-odoo:
I disagree. Take this html:
`<b class="X"><span class="X">text</span></b>`
The removeClass dom helper, when called with the text node will remove the `X` class from the `span`.
However, in your version, it will remove it from the `b` as well. I think this is wrong, don't you ?

Goaman:
Yes. The question is how can we know which element do we want to remove? Only one or all of them?
Most of the time, the class `X` is only on one of the element, rarely on multiples element. So I prefer the heuristic where we go through all modifiers.
In your version you decide it's only one modifier and it is an arbitrary choice as we could need the alternative. Which I needed.

I checked the code in Odoo and I cannot find where it was required. So it is up to you what we do about this PR.

dmo-odoo:
You say that "Most of the time, the class X is only on one of the element, rarely on multiples element." If this is true then I clearly prefer the current behavior since it means it will work most of the time while your version will only rarely be useful.

As such, I'm closing this PR. If you manage to find the case where you needed this then I'm open to continue this discussion, but right now the current behavior makes more sense to me.

I realize it might be my fault to have let this PR to rot for such a long time that you don't remember in what case it made sense, sorry.

PR #263, opened 2020-07-28
[FIX] Mode: allow cascading and non-cascading rule properties

Zinston:
Up to this commit every rule property was set based on a cascading selector, which means the properties were always propagated to all descendants of the node matched by the selector. This makes sense for `editable` but caused issues with `breakable` and made little sense for `allowEmpty`.
This reorganizes the rule definition by introducing the `cascading` boolean (default: false).

Linked Odoo PR: https://github.com/odoo-dev/odoo/pull/339

### DIFF
--- a/packages/core/src/Core.ts
+++ b/packages/core/src/Core.ts
@@ -36,7 +36,6 @@ export class Core<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
             handler: this.deleteForward,
         },
     };
-    mode = this.editor.mode;
 
     //--------------------------------------------------------------------------
     // Public
@@ -51,7 +50,7 @@ export class Core<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
         if (!range.isCollapsed()) {
             range.empty();
         }
-        if (this.mode.is(range.startContainer, RuleProperty.BREAKABLE)) {
+        if (range.mode.is(range.startContainer, RuleProperty.BREAKABLE)) {
             range.startContainer.splitAt(range.start);
         } else {
             // Use a separator to break paragraphs in an unbreakable.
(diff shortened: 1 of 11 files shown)


packages/core/src/Mode.ts
```diff
@@ -20,9 +23,16 @@ export enum RuleProperty {
...
-const modeProperties = ['editable', 'breakable', 'allowEmpty'];
 
-type RuleEntries = Partial<Record<RuleProperty, ContextualEntry<boolean>[]>>;
+/**
+ * Object matching properties to their definitions, for use in rules.
+ *
+ * If boolean, it is assumed that the value of `cascading` is false and that
+ * what was passed was `PropertyDefinition.value`.
+ */
+export type RuleProperties = { [K in RuleProperty]?: PropertyDefinition | boolean };
```
dmo-odoo, line 33: J'aurais mis ce type là en inline dans `ModeRules` plutot que lui donner un nom particulier. Besides, faut retirer le type `boolean` du coup j'imagine. En fait ça peut être un `Partial<Record` ce type non ?

Zinston: Ca ne me dérange pas.
Partial<Record, j'avais essayé mais ça ne marche pas (une limitation de TS je pense).

Zinston: @dmo-odoo done

packages/core/src/Core.ts
```diff
@@ -119,26 +118,26 @@ export class Core<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
         if (range.isCollapsed()) {
             // Basic case: remove the node directly following the range.
             const nextSibling = range.end.nextSibling();
-            if (nextSibling && this.mode.is(nextSibling, RuleProperty.EDITABLE)) {
+            if (nextSibling && this.editor.mode.is(nextSibling, RuleProperty.EDITABLE)) {
```
dmo-odoo, line 121: Didn't we say that the mode should be taken from the range in context rather than the editor ?

packages/core/src/JWEditor.ts
```diff
@@ -75,7 +75,7 @@ export class JWEditor {
     // each command batch is waited for.
     preventRenders: Set<Function> = new Set();
     enableRender = true;
-    private _modes: Record<ModeIdentifier, Mode> = {
+    _modes: Record<ModeIdentifier, Mode> = {
```
dmo-odoo, line 78: Rhen remove the `_`, same for `_rules`.

packages/plugin-dom-editable/src/DomEditable.ts
```diff
@@ -78,6 +78,8 @@ export class DomEditable<T extends JWPluginConfig = JWPluginConfig> extends JWPl
             case 'selectAll':
                 return ['selectAll', {}];
             case 'setSelection': {
+                const domLayout = this.dependencies.get(DomLayout);
+                if (!domLayout.isInEditable(action.domSelection.anchorNode)) return;
```
dmo-odoo, line 82: I think @Goaman disagrees. I seem to remember he had a discussion with @Gorash about something similar last week. Though it might have been about the actual selection of the dom rather than our internal selection so I'm not sure... I'm waiting for his reaction.

Goaman: The thing that I disagree is the normalizer not sending the `setSelection` in the first place.
And I suggested that after the normalizer send the signal, the editor/layout/whatever discard if it's not in the "editable".

packages/core/src/Core.ts
Zinston: Damn you're absolutely right... :O

packages/plugin-dom-editable/src/DomEditable.ts
Zinston: Actually you're right it shouldn't arrive here at all ^^ The reason is `if (!eventTarget || this._isInEditable(eventTarget)) {` in `EventNormalizer` (line 583). It should be `if (eventTarget && this._isInEditable(eventTarget)) {`.

PR #264, opened 2020-07-28
[IMP] Core: delete an empty container instead of merging into it

Zinston:
If a heading is preceded by an empty paragraph and we press backspace at the beginning of said heading, we want to simply move the heading up, we don't want it to become a paragraph. This commit addresses that issue so the behavior of `deleteBackward` and `deleteForward` follows the GSpec.

### DIFF
--- a/packages/core/test/VDocument.test.ts
+++ b/packages/core/test/VDocument.test.ts
@@ -353,13 +353,20 @@ describe('VDocument', () => {
                         contentAfter: '<h1>ab[]</h1>',
                     });
                 });
+                it('should merge a heading1 with text into an empty paragraph (keeping the heading)', async () => {
+                    await testEditor(BasicEditor, {
+                        contentBefore: '<p><br>[]</p><h1>ab</h1>',
+                        stepFunction: deleteForward,
+                        contentAfter: '<h1>[]ab</h1>',
+                    });
+                });
             });
             describe('With attributes', () => {
                 it('should merge a paragraph without class into an empty paragraph with a class', async () => {
                     await testEditor(BasicEditor, {
                         contentBefore: '<p class="a"><br>[]</p><p>abc</p>',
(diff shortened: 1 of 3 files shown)


packages/core/test/VDocument.test.ts
```diff
@@ -353,13 +353,20 @@ describe('VDocument', () => {
...
+                    });
+                });
             });
             describe('With attributes', () => {
                 it('should merge a paragraph without class into an empty paragraph with a class', async () => {
                     await testEditor(BasicEditor, {
                         contentBefore: '<p class="a"><br>[]</p><p>abc</p>',
                         stepFunction: deleteForward,
-                        contentAfter: '<p class="a">[]abc</p>',
+                        contentAfter: '<p>[]abc</p>',
```
dmo-odoo, line 369: I do not agree with this particular change.

packages/core/test/VDocument.test.ts
```diff
@@ -989,7 +1003,7 @@ describe('VDocument', () => {
                     await testEditor(BasicEditor, {
                         contentBefore: '<p class="a"><br></p><p>[]abc</p>',
                         stepFunction: deleteBackward,
-                        contentAfter: '<p class="a">[]abc</p>',
+                        contentAfter: '<p>[]abc</p>',
```
dmo-odoo, line 1006: Neither this one, obviously.

Zinston: But we just discussed it... Gspec...

dmo-odoo: Well I did void my disagreement when we discussed it though. Discussing things does not mean agreeing on things you know... :p We just ended up changing the subject on this one.

Zinston: ...

PR #267, opened 2020-07-29
[FIX] Normalizer: better check inputEvent.data

Goaman:


dmo-odoo:
Just for the record because I just realized I never wrote it down, I'm not merging this unless I get to understand how this change can have an actual impact. From my understanding, the new code is 100% equivalent to the previous one so I don't understand in which case it can fix anything.

PR #272, opened 2020-08-03
[ADD] Table: buttons to add/remove rows/columns

Zinston:


packages/core/src/VRange.ts
```diff
@@ -116,6 +116,23 @@ export class VRange {
         const startIndex = this.start.parent.childVNodes.indexOf(this.start);
         return this.startContainer.childVNodes[startIndex + 1] === this.end;
     }
+    /**
+     * Return true if the range is contained within the given container.
```
dmo-odoo, line 120: Technically, this sentence is wrong. It returns true if at least the start or the end of the rage is contained within the given container. The *entire* range being contained within a given container is something else.

packages/plugin-table/assets/tableUI.css
```diff
@@ -18,3 +18,18 @@ table.table-picker td {
...
+    width: 1em;
+    color: #00FF00;
+}
+span.table-handler.fa-minus-square {
+    position: absolute;
+    cursor: pointer;
+    background-color: white;
+    width: 1em;
+    color: #FF0000;
+}
```
dmo-odoo, line 35: Taken from Odoo and improved ? :p

packages/plugin-table/src/Table.ts
```diff
@@ -99,7 +99,7 @@ export class Table<T extends TableConfig = TableConfig> extends JWPlugin<T> {
             },
             {
                 id: 'TablePicker',
-                async render(editor: JWEditor): Promise<VNode[]> {
+                async render(editor: JWEditor): Promise<TablePickerNode[]> {
```
dmo-odoo, line 102: Is this really useful ? I'm not complaining, it can stay there I don't mind but I'm wondering if this actually changes something meaningful somewhere.

packages/plugin-table/src/TableCellDomObjectRenderer.ts
```diff
@@ -15,7 +16,7 @@ export class TableCellDomObjectRenderer extends NodeRenderer<DomObject> {
     async render(cell: TableCellNode): Promise<DomObject> {
         // If the cell is not active, do not render it (it means it is
         // represented by its manager cell's colspan or rowspan: it was merged).
-        if (!cell.isActive()) return { children: [] };
+        if (!cell.isActive() || !cell.ancestor(TableNode)) return { children: [] };
```
dmo-odoo, line 19: How can this occur ? Just wondering.

packages/plugin-table/assets/tableUI.css
Zinston: Yup :D

packages/plugin-table/src/Table.ts
Zinston: No but I find it clearer: it makes it clear that what is being rendered is a special node that I can then go inspect.

packages/core/src/VRange.ts
Zinston: _You_'re the end of the rage.

dmo-odoo: https://media.giphy.com/media/3orieZbA0wPRjR0b8Q/giphy.gif

PR #279, opened 2020-08-06
[FIX] followRange : Prevent floating menu to appear if out of the editable zone

sebgeelen:


packages/plugin-dom-follow-range/src/FollowRangeZoneDomObjectRenderer.ts
```diff
@@ -62,11 +62,19 @@ export class FollowRangeZoneDomObjectRenderer extends NodeRenderer<DomObject> {
                 }
             }
         } while (doc);
+        // Check if the selection is inside the the content editable
```
Goaman, line 65: The name of the variable is sufficient

packages/plugin-dom-follow-range/src/FollowRangeZoneDomObjectRenderer.ts
```diff
@@ -62,11 +62,19 @@ export class FollowRangeZoneDomObjectRenderer extends NodeRenderer<DomObject> {
...
+        const selectionIsIneditable =
+            !!selection &&
+            selection.anchorNode &&
+            selection.anchorNode.parentElement.closest('[contenteditable="true"]');
 
-        // If the selection goes into an input inside the jw-follow-range, do nothing.
+        // If the selection goes into an input inside the jw-follow-range,
+        // or the selection is not in editable zone
+        // do nothing.
```
Goaman, line 73: The comment is not necessary, the code is obvious

packages/plugin-dom-follow-range/src/FollowRangeZoneDomObjectRenderer.ts
```diff
@@ -62,11 +62,19 @@ export class FollowRangeZoneDomObjectRenderer extends NodeRenderer<DomObject> {
...
         if (
-            document.activeElement instanceof HTMLInputElement &&
-            document.activeElement.closest('JW-FOLLOW-RANGE')
+            (document.activeElement instanceof HTMLInputElement &&
+                document.activeElement.closest('JW-FOLLOW-RANGE')) ||
+            !selectionIsIneditable
```
Goaman, line 77: What happen if the toolbar is opened and wether:
- the user click outside editable
- the user destroy the editor

I guess that at least for the first case, the toolbar will remain open.

Maybe in the next condition `if (selection.rangeCount && !isCollapsed) {`
You can add `!selectionIsInEditable`

So: `if (selection.rangeCount && !isCollapsed && !selectionIsInEditable) {`

packages/plugin-dom-follow-range/src/FollowRangeZoneDomObjectRenderer.ts
```diff
@@ -62,11 +62,19 @@ export class FollowRangeZoneDomObjectRenderer extends NodeRenderer<DomObject> {
                 }
             }
         } while (doc);
+        // Check if the selection is inside the the content editable
+        const selectionIsIneditable =
```
dmo-odoo, line 66: `selectionIsInEditable`

packages/plugin-dom-follow-range/src/FollowRangeZoneDomObjectRenderer.ts
sebgeelen: Well, don't we want the floating bar to remain open if we fidget outside the editor ?

Goaman: I don't think so. Otherwise to close it, you need to click on the editor, then click on the other gui elements.
Ckeditor is doing what I suggest:
https://ckeditor.com/ckeditor-5/demo/#balloon

sebgeelen: ok, I'll change

PR #291, opened 2020-08-11
deadlock detection

Goaman:


packages/core/src/JWEditor.ts
```diff
@@ -364,6 +364,12 @@ export class JWEditor {
                 );
                 return;
             }
+            const timeout = setTimeout(() => {
+                console.warn(
+                    'An execCommand call take more than 10seconds to finish. It might be caused by a deadlock.\n' +
```
dmo-odoo, line 369: *is taking more than 10 seconds*

Goaman: done

PR #292, opened 2020-08-11
[IMP] DomHelper: introduce ExecutionContext

Goaman:


packages/plugin-dom-helpers/src/DomHelpers.ts
```diff
@@ -5,6 +5,18 @@ import { Attributes } from '../../plugin-xml/src/Attributes';
 import { Layout } from '../../plugin-layout/src/Layout';
 import { DomLayoutEngine } from '../../plugin-dom-layout/src/DomLayoutEngine';
 import { Parser } from '../../plugin-parser/src/Parser';
+import { Format } from '../../core/src/Format';
+import { Modifiers } from '../../core/src/Modifiers';
+import { flat } from '../../utils/src/utils';
+import { CommandParamsType, Commands } from '../../core/src/JWEditor';
+import { Context } from '../../core/src/ContextManager';
+
+export interface ExecutableContext {
```
dmo-odoo, line 14: This morning I realized I used `ExecutionContext` when talking about it rather than `ExecutableContext`. What do you think about `ExecutionContext` ? Yay or nay ?

Goaman:
Odoo pr: https://github.com/odoo-dev/odoo/pull/364

packages/plugin-dom-helpers/src/DomHelpers.ts
Goaman: changed in code and commit message

PR #295, opened 2020-08-11
[FIX] FontAwesome: remove FontAwesome classes on parsing

dmo-odoo:
And restore them afterwards. This required to rewrite how ClassList
worked in order to keep a history of parsed-then-removed classes to
be able to restore them at their previous position on rendering.

### DIFF
--- a/packages/plugin-xml/src/AttributesDomObjectModifierRenderer.ts
+++ b/packages/plugin-xml/src/AttributesDomObjectModifierRenderer.ts
@@ -20,11 +20,15 @@ export class AttributesDomObjectModifierRenderer extends ModifierRenderer<DomObj
      */
     async render(modifier: Attributes, contents: DomObject[]): Promise<DomObject[]> {
         const keys = modifier.keys();
-        if (keys.length) {
+        const classHistory: Record<string, boolean> = modifier.classList.history();
+        if (keys.length || Object.keys(classHistory).length) {
             const attributes: DomObjectAttributes = {};
             for (const name of keys) {
                 if (name === 'class') {
-                    attributes.class = new Set(modifier.classList.items());
+                    // This is going to use the class history feature anyway.
+                    // TODO: This entire file should probably be reorganized to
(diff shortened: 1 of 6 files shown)


packages/plugin-xml/src/ClassList.ts
```diff
@@ -72,13 +71,24 @@ export class ClassList extends VersionableObject {
      * @param name
      */
     has(name: string): boolean {
-        return this._classList?.has(name) || false;
+        return this._classList?.[name] || false;
     }
     /**
      * Return an array containing all the items in the list.
+     *
```
Zinston, line 78: .

packages/plugin-xml/src/AttributesDomObjectModifierRenderer.ts
```diff
@@ -78,20 +82,42 @@ export class AttributesDomObjectModifierRenderer extends ModifierRenderer<DomObj
...
+                this._applyClassHistory(content, classHistory);
             } else if (name === 'style') {
                 attr.style = Object.assign({}, attributes.style, attr.style);
             } else {
                 attr[name] = attributes[name];
             }
         }
     }
+
+    private _applyClassHistory(
```
Zinston, line 103: Doc would have been nice.

packages/plugin-xml/src/AttributesDomObjectModifierRenderer.ts
```diff
@@ -78,20 +82,42 @@ export class AttributesDomObjectModifierRenderer extends ModifierRenderer<DomObj
...
+    ): void {
         if (!content.attributes) content.attributes = {};
         const attr = content.attributes;
         for (const name in attributes) {
             if (name === 'class') {
-                if (!attr.class) attr.class = new Set();
-                for (const className of attributes.class) {
-                    attr.class.add(className);
-                }
+                this._applyClassHistory(content, classHistory);
```
Zinston, line 94: Karl Marx is not going to be happy about this.

PR #297, opened 2020-08-13
[IMP] Link: Keep selecting after command link

Goaman:
In Odoo we need to attach information after calling a link command.
The behavior to keep the selection after the selection make sense.

By having the selection remaining, it is possible to attach information
to characters of the range.

### DIFF
--- a/packages/plugin-link/src/Link.ts
+++ b/packages/plugin-link/src/Link.ts
@@ -6,7 +6,7 @@ import { InlineNode } from '../../plugin-inline/src/InlineNode';
 import { LinkFormat } from './LinkFormat';
 import { Char } from '../../plugin-char/src/Char';
 import { Modifiers } from '../../core/src/Modifiers';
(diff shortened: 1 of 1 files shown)


packages/plugin-link/src/Link.ts
```diff
@@ -145,11 +145,18 @@ export class Link<T extends JWPluginConfig = JWPluginConfig> extends JWPlugin<T>
...
-        return params.context.execCommand<Char>('insertText', {
-            text: label || link.url,
-            formats: new Modifiers(link),
-            context: params.context,
-        });
+        return this.editor.execWithRange<Char>(
+            [
+                [range.start, RelativePosition.AFTER],
+                [range.end, RelativePosition.BEFORE],
+            ],
```
dmo-odoo, line 152: You probably want to use `VRange.clone(range)` in this case. However, this is going to diverge from gspec. Wouldn't it be better to clone the range from Odoo side and call `execWithRange` on the previously cloned range to do the additional stuff you would like to do with the "previous" range ?

PR #311, opened 2020-08-18
[FIX] DomLayout: don't try to redraw item previously removed.

Gorash:


packages/plugin-dom-layout/src/DomReconciliationEngine.ts
```diff
@@ -785,6 +785,7 @@ export class DomReconciliationEngine {
 
         const oldChildren = old?.children || [];
         const newChildren = domObject.children || [];
+        const p = domObject;
```
dmo-odoo, line 788: Pourquoi `p` ?

PR #324, opened 2020-08-25
[FIX] FontAwesome: remove double click handler when detach

sebgeelen:


packages/plugin-odoo/src/OdooFontAwesomeDomObjectRenderer.ts
```diff
@@ -13,16 +13,21 @@ export class OdooFontAwesomeDomObjectRenderer extends FontAwesomeDomObjectRender
         const domObject: DomObject = await super.render(node, worker);
         if (domObject && 'children' in domObject) {
             const fa = domObject.children[1];
+
+            const bdClickCallback = () => {
```
dmo-odoo, line 17: bdClick? ^^'

dmo-odoo (commented):
No test to ensure this bug cannot come back ? :(

packages/plugin-odoo/src/OdooFontAwesomeDomObjectRenderer.ts
sebgeelen: XD

sebgeelen:
@dmo-odoo We don't yet have tests on odoo specific plugin.
To be honest, I don't even know how I should tests this on JW code only ...

So, should i spend time on this now ?

dmo-odoo:
@sebgeelen I didn't realize this was the Odoo version of fontawesome renderer. It's fine sorry.

PR #325, opened 2020-08-25
[FIX] Toolbar: properly display image UI

Zinston:


packages/plugin-dom-layout/src/ActionableGroupDomObjectRenderer.ts
```diff
@@ -36,4 +53,27 @@ export class ActionableGroupDomObjectRenderer extends NodeRenderer<DomObject> {
...
     }
+    /**
+     * Update option rendering after the command if the value of visible
+     * changed.
+     *
+     * @param actionable
+     * @param element
+     */
+    protected _updateActionables(): void {
+        for (const [actionable, element] of this.actionableNodes) {
```
dmo-odoo, line 64: I don't understand. Why do we have two "_updateActionables" functions?

packages/plugin-odoo/src/Odoo.ts
```diff
@@ -60,12 +61,9 @@ const imageClassesLabels = {
  * If more images are within the range, return undefined.
  */
 function getSingleImage(range: VRange): ImageNode | undefined {
-    const next = range.start.nextLeaf();
-    if (next instanceof ImageNode) {
-        const prev = range.end.previousLeaf();
-        if (prev === next) {
-            return next;
-        }
+    const selected = range.selectedNodes(node => node instanceof AtomicNode);
```
dmo-odoo, line 64: I think this function was previously refactored by @Gorash for performance reasons. Further checking is needed to ensure your version is not reverting back to a less performant state.

packages/plugin-dom-layout/src/ActionableGroupDomObjectRenderer.ts
Zinston: We have three :sunglasses:
One for Actionable, one for ActionableGroup, and one for ActionableGroupSelectItem.

packages/plugin-odoo/src/Odoo.ts
Zinston: @Gorash @Gorash @Gorash

Gorash: no!
don't use selectedNodes in getSingleImage, it's called fto display button too.

packages/plugin-dom-layout/src/ActionableGroupDomObjectRenderer.ts
dmo-odoo: But... only the one of Actionable shoulb be called `_updateActionables`, right ? Shouldn't the others be `_updateActionableGroupItems` and `_updateActionableGroupeSelectItems` ?

Zinston:
yes sure

dmo-odoo:
@Zinston This PR has conflicts.

PR #327, opened 2020-08-26
[FIX] Odoo: wrap only inlineNode

Goaman:
In the signature of Odoo, the content is only based of inline nodes
and the Editor does not yet support properly when that case happen.

For instance, when trying to align the content that does not have
a ContainerNode that is editable, the command will do nothing.

This commit fix that by wrapping all the node inside the default
container of the editor.

### DIFF
--- a/packages/plugin-odoo/src/NoteEditableXmlDomParser.ts
+++ b/packages/plugin-odoo/src/NoteEditableXmlDomParser.ts
@@ -1,0 +1,45 @@
+import { XmlDomParsingEngine } from '../../plugin-xml/src/XmlDomParsingEngine';
+import { DividerNode } from '../../plugin-divider/src/DividerNode';
+import { nodeName } from '../../utils/src/utils';
+import { DividerXmlDomParser } from '../../plugin-divider/src/DividerXmlDomParser';
+import { AtomicNode } from '../../core/src/VNodes/AtomicNode';
+import { VNode } from '../../core/src/VNodes/VNode';
+
+export class NoteEditableXmlDomParser extends DividerXmlDomParser {
+    static id = XmlDomParsingEngine.id;
+    engine: XmlDomParsingEngine;
+
+    predicate = (item: Node): boolean => {
+        return (
+            item instanceof Element &&
+            nodeName(item) === 'DIV' &&
+            item.classList.contains('note-editable')
+        );
+    };
+
+    async parse(item: Element): Promise<DividerNode[]> {
+        const divider = (await super.parse(item))[0];
+
+        let looseChildren: VNode[] = [];
+        const newChildren: VNode[] = [];
+        for (const child of divider.childVNodes) {
+            if (child instanceof AtomicNode) {
+                looseChildren.push(child);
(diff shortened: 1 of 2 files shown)


packages/plugin-odoo/src/NoteEditableXmlDomParser.ts
```diff
@@ -0,0 +1,30 @@
...
+    };
+
+    async parse(item: Element): Promise<DividerNode[]> {
+        const divider = (await super.parse(item))[0];
+
+        const children = divider.childVNodes;
+        if (children.every(child => child instanceof AtomicNode)) {
+            const container = new this.engine.editor.configuration.defaults.Container();
+            container.append(...children);
+            divider.append(container);
```
dmo-odoo, line 26: Wait so... every atomic child of the `note-editable` div is put into its own personal container ? Wouldn't it be more logical to have them all in the same container ? Besides, wouldn't your bug still exist if only some of the nodes are atomic but not all of them ? And wouldn't thix fix create additional `div` after rendering ? This fix here basically looks extremely fishy to me.

Goaman: You're right, I forgot to consider the case where there is some node that are not Atomic. I'm adapting the code.
This fix indeed create additional `div` after the rendering. 

That's a trade-off we were discussing with @Zinston. The more appropriate fix is to handle the `InlineNode` inside a container but that would require more work.

dmo-odoo: What is the actual bug (steps to reproduce) this is trying to solve ?

Goaman: Try to edit the signature in Odoo. The content is 
```html
<span data-o-mail-quote="1">--&nbsp;</span>
<br data-o-mail-quote="1">
<span data-o-mail-quote="1">Administrator</span>
```

Goaman: I changed the code to take into consideration nodes that are not inline.

packages/plugin-odoo/src/NoteEditableXmlDomParser.ts
```diff
@@ -0,0 +1,44 @@
...
+            } else {
+                this.addChildren(newChilds, looseChilds);
+                looseChilds = [];
+                newChilds.push(child);
+            }
+        }
+        this.addChildren(newChilds, looseChilds);
+        divider.append(...newChilds);
+        return [divider];
+    }
```
dmo-odoo, line 36: I don't like this construction but the alternatives I came up with were not necessarily better so we'll leave it at that and carry on. However, "childs" does not exist in english. The plural form of "child" is "children".

Goaman: done

dmo-odoo:
This PR has conflicts.

Goaman:
Done. It was an import conflict.

PR #330, opened 2020-08-26
[IMP] DocumentIcon: Add document-icon

Goaman:


dmo-odoo:
I'm closing this PR because @Gorash told me the current implementation of documents in Odoo (with a background image) was not by design but because he had to circumvent a limitation of summernote. He explained to me how he would have liked it to work (by having an `img` inside a `a`) and I asked @Zinston to try it. Since it worked and is way simpler than the current implementation in Odoo, I merged it. https://github.com/odoo-dev/odoo/pull/384 Since adding documents from the media modal is now supported (without even creating a custom node in jw for it), I'm closing this PR.

PR #331, opened 2020-08-27
[IMP] fontAwesome auto select icon when click

sebgeelen:
+ add data for odoo font awesome replacement

link to odoo PR : https://github.com/odoo-dev/odoo/pull/373

### DIFF
--- a/packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
+++ b/packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
@@ -20,10 +20,27 @@ export class FontAwesomeDomObjectRenderer extends NodeRenderer<DomObject> {
         node: FontAwesomeNode,
         worker: RenderingEngineWorker<DomObject>,
     ): Promise<DomObject> {
+        const select = (ev: Event) => {
+            ev.preventDefault();
+            const selectFontAwesome = () => {
+                this.engine.editor.selection.select(node, node);
+            };
+            // The normaliser wait a tick before changing the selection when the user click
+            // in order to override the normalizer default behavior we also need to wait a tick
+            setTimeout(() => {
+                this.engine.editor.execCommand(selectFontAwesome);
+            });
+        };
         const fontawesome: DomObjectElement = {
             tag: node.htmlTag,
             attributes: {
                 class: new Set(node.faClasses),
+            },
+            attach: (el: HTMLElement): void => {
+                el.addEventListener('mouseup', select, true);
+            },
+            detach: (el: HTMLElement): void => {
+                el.removeEventListener('mouseup', select, true);
             },
         };
         // Surround the fontawesome with two invisible characters so the
--- a/packages/plugin-odoo/src/OdooFontAwesomeDomObjectRenderer.ts
+++ b/packages/plugin-odoo/src/OdooFontAwesomeDomObjectRenderer.ts
@@ -26,7 +26,11 @@ export class OdooFontAwesomeDomObjectRenderer extends FontAwesomeDomObjectRender
                     }
                     el.addEventListener('dblclick', dbclickCallback);
                 };
+                const savedDetach = fa.detach;
                 fa.detach = (el: HTMLElement): void => {
+                    if (savedDetach) {
+                        savedDetach(el);
+                    }
                     el.removeEventListener('dblclick', dbclickCallback);
                 };
             }
--- a/packages/plugin-odoo/src/OdooFontAwesomeDomObjectRenderer.ts
+++ b/packages/plugin-odoo/src/OdooFontAwesomeDomObjectRenderer.ts
@@ -2,6 +2,7 @@ import { FontAwesomeNode } from '../../plugin-fontawesome/src/FontAwesomeNode';
 import { DomObject } from '../../plugin-renderer-dom-object/src/DomObjectRenderingEngine';
 import { FontAwesomeDomObjectRenderer } from '../../plugin-fontawesome/src/FontAwesomeDomObjectRenderer';
 import { RenderingEngineWorker } from '../../plugin-renderer/src/RenderingEngineCache';
+import { RelativePosition } from '../../core/src/VNodes/VNode';
 
 export class OdooFontAwesomeDomObjectRenderer extends FontAwesomeDomObjectRenderer {
     predicate = FontAwesomeNode;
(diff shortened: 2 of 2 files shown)


packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
```diff
@@ -20,11 +20,28 @@ export class FontAwesomeDomObjectRenderer extends NodeRenderer<DomObject> {
...
         worker: RenderingEngineWorker<DomObject>,
     ): Promise<DomObject> {
+        const select = (ev: Event) => {
+            ev.preventDefault();
+            const selectFontAwesome = () => {
+                this.engine.editor.selection.select(node, node);
+            };
+            // The normaliser wait a tick before changing the selection when the user click
+            // in order to override the normalizer default behavior we also need to wait a tick
+            setTimeout(() => {
```
dmo-odoo, line 30: Ca ca fait peur quand même...

packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
```diff
@@ -20,11 +20,28 @@ export class FontAwesomeDomObjectRenderer extends NodeRenderer<DomObject> {
...
+                this.engine.editor.execCommand(selectFontAwesome);
+            });
```
Goaman, line 32: This can be wrong in case of concurrency issue.
Imagine that a user do a batch of event like:
[select, keydown, ...] in one tick, then the selection will be wrong as it happens after many event occurred.

packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
```diff
@@ -20,11 +20,28 @@ export class FontAwesomeDomObjectRenderer extends NodeRenderer<DomObject> {
         node: FontAwesomeNode,
         worker: RenderingEngineWorker<DomObject>,
     ): Promise<DomObject> {
+        const select = (ev: Event) => {
+            ev.preventDefault();
+            const selectFontAwesome = () => {
+                this.engine.editor.selection.select(node, node);
+            };
```
Goaman, line 27: Why do you need to create a variable function? You only use it once.

packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
```diff
@@ -20,11 +20,28 @@ export class FontAwesomeDomObjectRenderer extends NodeRenderer<DomObject> {
...
             tag: node.htmlTag,
             attributes: {
                 class: new Set(node.faClasses),
             },
+            attach: (el: HTMLElement): void => {
+                el.addEventListener('mouseup', select, true);
+            },
+            detach: (el: HTMLElement): void => {
+                el.removeEventListener('mouseup', select, true);
+            },
```
Goaman, line 44: Don't we only want that in Odoo?

packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
sebgeelen: To be able to have the custom command named in the JW dev tool (best practice recommended by @Gorash )

packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
sebgeelen: I don't know, I based this on similar behavior for images, that I found in JW.
@Zinston made changes on this in images recently, and suggested to do the same for FontAwesome

packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
sebgeelen: Yeah, I'm not a fan either, But how can I do this differently ?

packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
sebgeelen: ouaip :/

packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
Goaman: Also, it is wrong as it does not prevent the `Normalizer`, it rather tries to **correct** it. Which is wrong.
We want to correctly prevent the `Normalizer` for the `mousedown` to not trigger a `setSelection`.

We need a mechanism in the `Normalizer` for that.
Like when there is a `mousedown`, we need to wait if there was no (click/double click/triple click) that has followed.

packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
Goaman: ah yes I forgot

packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
Goaman: As I changed related code in my PR "concurency", if we want that mechanism, we need to wait the PR to be merged.

packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
Goaman: Maybe it's what we need. cc @dmo-odoo @Zinston 
I believe it belongs to click rather than `mouseup`. But for that, the `Normalizer` need the detection mechanism mention above.

sebgeelen: @Zinston expecitely changed it from clic to mouseUp in his PR. I guess he has a good reason.

sebgeelen: cf : https://github.com/odoo-dev/jabberwock/pull/325

packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
sebgeelen: We can wait if your changes make this better

dmo-odoo: I think this might be a case where it is worth it to wait for the concurrency PR to be merged.

packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
dmo-odoo: @Goaman I just merged @Zinston's PR so if this click/mouseup shenanigans need to be changed with the concurrency changes, I suggest you to rebase your branch on master to check if it works or if it can/needs to be changed.

dmo-odoo: Then you can report to @sebgeelen with your findings.

packages/plugin-fontawesome/src/FontAwesomeDomObjectRenderer.ts
sebgeelen: :+1:

PR #350, opened 2020-09-01
[FIX] Odoo: properly replace an existing link

Zinston:
Linked Odoo PR: https://github.com/odoo-dev/odoo/pull/375

### DIFF
--- a/packages/core/src/VNodes/AbstractNode.ts
+++ b/packages/core/src/VNodes/AbstractNode.ts
@@ -310,6 +310,8 @@ export abstract class AbstractNode extends EventMixin {
     }
     /**
      * Return the nodes adjacent to this VNode that satisfy the given predicate.
+     * Note: include this VNode within the return value, in order of appearance
+     * (if it satisfies the given predicate).
      */
     adjacents<T extends VNode>(predicate?: Predicate<T>): T[];
     adjacents(predicate?: Predicate): VNode[];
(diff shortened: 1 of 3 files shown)


packages/core/src/VNodes/AbstractNode.ts
```diff
@@ -310,16 +310,21 @@ export abstract class AbstractNode extends EventMixin {
...
      * Return the nodes adjacent to this VNode that satisfy the given predicate.
+     * If `includeSelf` is set to true, include this VNode within the return
+     * value, in order of appearance (if it satisfies the given predicate).
      */
-    adjacents<T extends VNode>(predicate?: Predicate<T>): T[];
-    adjacents(predicate?: Predicate): VNode[];
-    adjacents(predicate?: Predicate): VNode[] {
+    adjacents<T extends VNode>(predicate?: Predicate<T>, includeSelf?: boolean): T[];
+    adjacents(predicate?: Predicate, includeSelf?: boolean): VNode[];
+    adjacents(predicate?: Predicate, includeSelf = false): VNode[] {
```
dmo-odoo, line 318: I disagree about this ad-hoc parameter. Either we decide that we always return self or we adapt the code that needs self in there to not require including self in adjacents.

Zinston: Then we should always include it unless stated otherwise because including it in the right place is a pain in the bottocks.

PR #354, opened 2020-09-02
[FIX] FontSize: remove duplicated logic for input

Goaman:
There were two conflicting logic to display the input in it's renderer.

### DIFF
--- a/packages/plugin-font-size/src/FontSizeDomObjectRenderer.ts
+++ b/packages/plugin-font-size/src/FontSizeDomObjectRenderer.ts
@@ -51,7 +51,6 @@ export class FontSizeDomObjectRenderer extends InputDomObjectRenderer {
 
         let fontSize = next?.modifiers.find(Attributes)?.style?.get('font-size');
 
-        input.style.display = 'block';
         if (fontSize) {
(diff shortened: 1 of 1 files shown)


dmo-odoo:
Ca me fait peur là, parce que tu retires deux endroits en disant que c'est de la logique dupliquée. Donc ça veut dire qu'il y a encore un *troisième* endroit où on faisait ça alors ? Il est où ce troisième endroit ?

dmo-odoo:
ping @Goaman

Goaman:
In the code of the function there is 
```ts
        const domVisible = input.style.display !== 'none';
        const visible = isInTextualContext(editor);
        if (visible !== domVisible) {
            if (visible) {
                input.style.display = 'inline-block';
            } else {
                input.style.setProperty('display', 'none', 'important');
            }
        }
```
Which was cancelled with:
```ts
input.style.display = 'block';
```
And the last line:
```ts
        } else {	
            input.style.setProperty('display', 'none', 'important');	
        }
```
It was hidding the input when the range was at the end of chars.

In the meantime I noticed that the fontsize was not changing when editing the input while the range was collapsed.
I added it in our pad with low priority as it is the same in Odoo master

PR #372, opened 2020-09-07
[FIX] Core: applyFormat everything

Goaman:


packages/plugin-mail/test/Mail.test.ts
```diff
@@ -52,10 +52,9 @@ describePlugin(Mail, testEditor => {
                     // with mac, the base hase the same layout but the base64 is different.
                     result.replace(/data:image\/png;base64,[^"]+/gi, 'data:image/png;base64,'),
                 ).to.equal(
-                    testMailResult.replace(
-                        /data:image\/png;base64,[^"]+/gi,
-                        'data:image/png;base64,',
-                    ),
+                    testMailResult
+                        .trim()
```
dmo-odoo, line 56: You added a `trim` that was not there before. Why ?

PR #432, opened 2020-10-06
Master reactive info nby

Goaman:


Goaman:
to merge with https://github.com/odoo-dev/odoo/pull/405

packages/core/src/Modifier.ts
```diff
@@ -19,11 +20,17 @@ export interface Modifier {
...
 export class Modifier extends EventMixin {
+    readonly _isVersionable = true;
     preserveAfterNode = true; // True to preserve modifier after the node that holds it.
     preserveAfterParagraphBreak = true; // True to preserve modifier after a paragraph break.
     preserveAfterLineBreak = true; // True to preserve modifier after a line break.
     level = ModifierLevel.MEDIUM;
 
+    constructor() {
+        super();
+        return makeVersionable(this);
```
dmo-odoo, line 31: That won't work AFAIK.
```js
class A {
    constructor() {
        return 42;
    }
}
const a = new A();
a !== 42 // true
```
I'm not sure how much of this PR is still relevant considering this wont work so I won't review further for the moment. In particular, I don't understand what is the problem of having EventMixin versionable. Basically, I think everything in our editor should be versionable. We should not have `ReactiveValueVersionable`, it should always be versionable (and versioned) IMO.

Goaman: I've look with @Gorash for this pull request.
Your code will not work because you don't return an object but the following yes:
```js
const foo = {'foo': 'bar'};
class A {
    constructor() {
        return foo;
    }
}
const a = new A();
a === foo // true
```
@Gorash told me he already tested this case.

I have at least two argument against versionning a ReactiveValue.

1) The `ReactiveValue` that we create here does not need to be versioned and we need to set it after the memory is closed.

2) Then, not everything in the editor should be versionable. The plugins, the engines, etc ...
But in the future, they all could be **reactive** to change value after start (add plugin/add shortcuts/...) and have automatic cascade throughout the editor. **This should not be versionable**.

Lastly, I've tested it, all the tests pass and it's working perfectly.

PR #442, opened 2020-10-14
master-typo-dom-nby

Goaman:


packages/plugin-renderer-dom-object/src/DomObjectRenderingEngine.ts
```diff
@@ -490,7 +490,7 @@ export class DomObjectRenderingEngine extends RenderingEngine<DomObject> {
             if (parent) {
                 const markers: VNode[] = [];
                 parent.childVNodes.forEach(sibling => {
-                    // Filter and sort the ndoes.
+                    // Filter and sort the nodes.
```
dmo-odoo, line 493: Please write a better commit message. It should be something like:
```[FIX] DomObjectRenderer: fix typo in comment```

Goaman: Done.

PR #446, opened 2020-10-19
[IMP] Resizer: prepare for use of Resizer in Odoo

Zinston:
Odoo Community PR: https://github.com/odoo/odoo/pull/60267
Odoo Enterprise PR: https://github.com/odoo/enterprise/pull/14201

### DIFF
--- a/packages/plugin-resizer/src/ResizerDomObjectRenderer.ts
+++ b/packages/plugin-resizer/src/ResizerDomObjectRenderer.ts
@@ -97,13 +97,14 @@ export class ResizerDomObjectRenderer extends NodeRenderer<DomObject> {
         const mainZone = this.domEngine.root.descendants(
             node => node instanceof ZoneNode && node.managedZones.includes('main'),
         )[0];
-        this.targetToResize = (this.domEngine.getDomNodes(
-            mainZone,
-        )[0] as HTMLElement)?.parentElement;
+        const domMain = this.domEngine.getDomNodes(mainZone)[0] as HTMLElement;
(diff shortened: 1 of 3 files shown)


packages/plugin-resizer/src/ResizerDomObjectRenderer.ts
```diff
@@ -97,13 +97,12 @@ export class ResizerDomObjectRenderer extends NodeRenderer<DomObject> {
...
-        this.targetToResize = (this.domEngine.getDomNodes(
-            mainZone,
-        )[0] as HTMLElement)?.parentElement;
+        const domMain = this.domEngine.getDomNodes(mainZone)[0] as HTMLElement;
+        this.targetToResize = domMain?.parentElement || domMain;
 
         // Force the overflow on the targetElement.
         // Necesary to make the resizer works out of the box.
-        this.targetToResize.style.overflow = 'auto';
+        if (this.targetToResize) this.targetToResize.style.overflow = 'auto';
```
dmo-odoo, line 105: I'm against `if (<condition>) doSomething`. The only time I accept it is when it's only a `return` but otherwise I would advise against it. If you disagree, I can explan you why using `if` without `{}` is a bad idea but I'm not doing it right now to save time in case we agree on the matter and this is just an oversight.

Zinston: I actually just did the same as 9 lines below (`if (this.targetToResize) this.targetToResize.style.height = height + 'px';`) but I agree it's not good practice. Should I correct the one below while I'm at it?

PR #453, opened 2020-10-21
master-modifier-contains-nby

Goaman:


packages/core/src/Modifiers.ts
```diff
@@ -3,7 +3,7 @@ import { Constructor, isConstructor } from '../../utils/src/utils';
 import { EventMixin } from '../../utils/src/EventMixin';
 import { VersionableArray } from './Memory/VersionableArray';
 
-export class Modifiers extends EventMixin {
+export class Modifiers extends EventMixin implements Iterable<Modifier> {
```
Zinston, line 6: C'est bien ça je trouve, ça décrit exactement ce que c'est !

packages/core/src/Modifiers.ts
```diff
@@ -247,6 +247,22 @@ export class Modifiers extends EventMixin {
     toggle(modifier: Modifier | Constructor<Modifier>): void {
         this.remove(modifier) || this.append(modifier);
     }
+    /**
+     * Check that all `otherModifiers` ar contained within this ones.
```
Zinston, line 251: "are", "one".
Sinon, j'aurais préféré une phrase active plutôt que passive (inversée quoi).
"Return true if this array contains all of the given modifiers." ?

packages/core/src/Modifiers.ts
```diff
@@ -247,6 +247,22 @@ export class Modifiers extends EventMixin {
...
+     *
+     * @param otherModifiers
+     */
+    contains(otherModifiers: Modifiers): boolean {
```
Zinston, line 255: Du coup pour rester dans l'esprit des autres méthodes, ce serait bien d'aussi accepter `...modifiers: Array<Modifier | Constructor<Modifier>>`.

packages/core/src/Modifiers.ts
```diff
@@ -247,6 +247,22 @@ export class Modifiers extends EventMixin {
     toggle(modifier: Modifier | Constructor<Modifier>): void {
         this.remove(modifier) || this.append(modifier);
     }
+    /**
+     * Check that all `otherModifiers` ar contained within this ones.
+     *
+     * @param otherModifiers
```
Zinston, line 253: J'appellerais l'argument `modifiers` parce que je vois pas de raison de l'appeler `otherModifiers`. Dans le cas de `areSameAs` je pense que ça a vaguement du sens parce que c'est vraiment fait pour comparer deux instances de `Modifiers` mais ici je trouve que `contains` c'est un truc qu'on aurait probablement envie de pouvoir utiliser de manière plus versatile, comme `filter`.

packages/core/src/Modifiers.ts
```diff
@@ -247,6 +247,22 @@ export class Modifiers extends EventMixin {
...
+     */
+    contains(otherModifiers: Modifiers): boolean {
+        for (const otherModifier of otherModifiers) {
+            const foundModifier = this.find(m => m.isSameAs(otherModifier));
+            // Modifier.isSameAs(undefined) could return true. See
+            // `Attribute.isSameAs`.
```
Zinston, line 259: `Attribute` est un plugin non ? Je ne le mentionnerais pas dans core. Tu peux peut-être plutôt expliquer pourquoi on peut se retrouver dans cette situation ?

packages/core/src/Modifiers.ts
```diff
@@ -255,17 +271,7 @@ export class Modifiers extends EventMixin {
...
-        );
-        const aModifiers = Array.from(modifiersMap.keys());
-        const bModifiers = Array.from(modifiersMap.values());
-
-        const allAinB = aModifiers.every(a => a.isSameAs(modifiersMap.get(a)));
-        const allBinA = otherModifiers.every(
-            b => bModifiers.includes(b) || b.isSameAs(this.find(b)),
-        );
-        return allAinB && allBinA;
+        return this.contains(otherModifiers) && otherModifiers.contains(this);
```
Zinston, line 274: C'est beaucoup plus clair comme ça pour le coup :-)
Ca m'a l'air bon sinon. Je suppose que tu as lancé les tests donc si tout est vert a priori y a pas de raison d'en douter (vu le nombre de comportements qui dépendent de ça et qui sont bien testés).

packages/core/src/Modifiers.ts
```diff
@@ -306,6 +312,20 @@ export class Modifiers extends EventMixin {
     map<T>(callbackfn: (value: Modifier, index: number, array: Modifier[]) => T): T[] {
         return this._contents?.map(callbackfn) || [];
     }
+    /**
+     * Iterate through all modifiers.
+     */
+    [Symbol.iterator](): Iterator<Modifier> {
```
Zinston, line 318: Et du coup avec ça tu peux faire `for (const modifier of modifiers)` avec `modifiers instanceof Modifiers` ? :+1:

packages/core/test/Modifiers.test.ts
```diff
@@ -505,6 +510,56 @@ describe('core', () => {
                 expect(modifiersMap[1] instanceof ExtendedModifier).to.be.true;
             });
         });
+        describe('contains()', () => {
+            it('should contain itself favorably', () => {
```
Zinston, line 514: Ca veut dire quoi "se contenir favorablement" ? :thinking:

packages/core/src/Modifiers.ts
Goaman: I specifically need to check if `Modifiers` are contained, not `Modifier`.

Including `Constructor<Modifier>` would introduce a new logic that I think is useless. `Modifier.isSameAs` does not take `Constructor<Modifier>`.
If we need to check the constructor, then using a `.filter()` would be sufficient as we don't need to check the undefined.

On the other hand I could use `Modifier[]` in the signature. Which could have two forms:
```ts
    contains(modifiers: Modifiers | Modifier[]): boolean {
        //...
```
or:
```ts
    contains(...args: [Modifiers] | Modifier[]): boolean {
        const modifiers = args[0] instanceof Modifiers ? args[0] : (args as Modifier[]);
        //...
```
Which one you prefer @dmo-odoo @Zinston ?

I've push the first version as I found it more elegant but I'll change it if you have a different taste.

packages/core/src/Modifiers.ts
Goaman: It's tricky to explain I have removed Attribute.isSameAs. Even though it does not shock me to make a reference to another plugin within a comment in core if it help to understand. Do you have an explanation that comes to your mind?

packages/core/src/Modifiers.ts
Zinston: The reason I suggested to include the other types was that most methods in `Modifiers` do and that this one is not fundamentally different from the others so from an API perspective I would expect to be able to use this one the same was as the others. Even though it admittedly adds a little bit of logic. As you said though:
> If we need to check the constructor, then using a .filter() would be sufficient as we don't need to check the undefined.
So there's little cost to putting that in `contains`.

packages/core/src/Modifiers.ts
Zinston: "Modifier.isSameAs(undefined) may return true in the case of modifiers that are themselves containers of values. Then we might want to equate the presence of an "empty" modifier with the absence of that modifier."

packages/core/test/Modifiers.test.ts
Zinston: This comment applied to the other `it`s as well.

packages/core/src/Modifiers.ts
Goaman: I find it accurate but difficult to understand for someone not familiar. Moreover, technically a modifier that does not represent a container can return true if undefined is provided. 

The example of `Attribute.isSameAs` would give me more information IMO. That's no t a problem to have an external link to explain or provide an example to a concept.

I've changed for your comment anyway.

Zinston: Fair enough. I'd do both then.

packages/core/src/Modifiers.ts
Goaman: So what do I do?
Do I continue to take `Modifiers` as argument?
Do I add the logic for Contructor<Modifier> ?

If I take it all it means the function would become this:
```ts
    contains(...args: [Modifiers] | (Modifier | Constructor<Modifier>)[]): boolean {
        if (args[0] instanceof Modifiers || args[0] instanceof Modifier) {
            const modifiers =
                args[0] instanceof Modifiers
                    ? args[0]
                    : (args as (Modifier)[]);
            for (const modifier of modifiers) {
                const foundModifier = this.find(m => m.isSameAs(modifier));
                // Modifier.isSameAs(undefined) may return true in the case of
                // modifiers that are themselves containers of values. Then we might
                // want to equate the presence of an "empty" modifier with the
                // absence of that modifier.
                if (!foundModifier && !modifier.isSameAs(undefined)) {
                    return false;
                }
            } else {
                const modifierClasses = [...new Set(args as Constructor<Modifier>[])];
                return modifierClasses.every(this.find.bind(this))
            }
        }
        return true;
    }
```
Which becomes complicated to read for not so much value.
And also if I want consistencies, it means that all other method should have the same signature (i.e. including `Modifiers`).

Goaman: To summarize some of the options:
`...modifiers: (Modifier | Constructor<Modifier>)[]`
- I don't like because it does not take `Modifiers`
- add branching + logic for Contstructor<Modifier>

`modifiers: Modifiers | Modifier[]`
- +simple
- require to change the other signature for consistency
- add branching + logic for Contstructor<Modifier>

`modifiers: Modifiers | (Modifier | Constructor<Modifier>)[]`
- require to change the other signature for consistency
- add branching + logic for Contstructor<Modifier>

`...args: [Modifiers] | (Modifier | Constructor<Modifier>)[]`
- require to change the other signature for consistency
- add branching + logic for Contstructor<Modifier>

Zinston: And my view is that we should look at all those options individually but do the same as the other methods as that is what I would expect from an API.

dmo-odoo: Mmh... Isn't it possible to make the `Modifiers` class compatible with `...` destructuring ?
Such that this would work:
```
function f(...modifiers: Modifier[] || Constructor<Modifier>[]) {}
```
```
const modifiers = new Modifiers();
f(...modifiers)
```
This way the `...` signature would work in all cases. It would work for `Modifier[]`, for `Constructor<Modifier>` *and* for `Modifiers` because `...` would transform it into `Modifier[]` through the iterable interface.

Goaman: I think the following would work:
```ts
const modifiers = new Modifiers();
f(...[modifiers]).
```

dmo-odoo: The would be great because you could then give *any* iterable of modifiers to these functions, including `Modifiers` itself since it would be iterable.

Zinston: :+1:
[...modifiers] would magically call Modifiers.clone ?

Goaman: I've tested I it does not work. You need `f(...[...modifier])`.
Then the code would be:
```ts
    contains(...modifiers: Modifier[] | Constructor<Modifier>[]): boolean {
        if (modifiers[0] instanceof Modifier) {
            for (const modifier of modifiers as Modifier[]) {
                const foundModifier = this.find(m => m.isSameAs(modifier));
                // Modifier.isSameAs(undefined) may return true in the case of
                // modifiers that are themselves containers of values. Then we might
                // want to equate the presence of an "empty" modifier with the
                // absence of that modifier.
                if (!foundModifier && !modifier.isSameAs(undefined)) {
                    return false;
                }
            }
            return true;
        } else {
            const modifierClasses = [...new Set(modifiers as Constructor<Modifier>[])];
            return modifierClasses.every(this.find.bind(this));
        }
    }
    areSameAs(otherModifiers: Modifiers): boolean {
        return this.contains(...[...otherModifiers]) && otherModifiers.contains(...[...this]);
    }
```

Goaman: The first `...` is the syntax to specify that you want an array being considered as the list of arguments.
So you need to `...` over an array.
The second expression with `[...otherModifiers]` say take the iterator OtherModifier and place it in an array through the spreat iterator.
I think the first `...` is not considered as a spread operator although it is written the same way.

dmo-odoo: Mmmh... Then what about this:
```
contains(...modifiers: Iterable<Modifier | Constructor<Modifier>): boolean {
```
Wouldn't `Modifiers` be an `Iterable<Modifier>` ?

Goaman: I already tried. I've tested in javascript and it would work.

The problem is typescript.
If you do `fn(...anything: AnyType) {}`, typescript tell you AnyType must be an array of something.
An iterable is not an Array so `fn(...anything: Iterable<any>) {}` cannot work.

And if you do `contains(...modifiers)` (which would work in js), typescript tells you:
> Argument of type 'unknown[]' is not assignable to parameter of type 'Modifier[] | Constructor<Modifier>[]'.

PR #456, opened 2020-10-26
[FIX] LineBreak: List item empty with br are displayed as twice line

Gorash:
The mobile versions represent the <br> in li as making 2 lines and not
just one (unlike browsers). The implementation of linebreaks needs to
change. The invisible terminal BR must be an invisible character so that
the previous BR is displayed but respecting more the W3 rules.


packages/plugin-linebreak/src/LineBreakXmlDomParser.ts
```diff
@@ -10,7 +10,10 @@ export class LineBreakXmlDomParser extends AbstractParser<Node> {
     engine: XmlDomParsingEngine;
 
     predicate = (item: Node): boolean => {
-        return item instanceof Element && nodeName(item) === 'BR';
+        return (
+            (item instanceof Element && nodeName(item) === 'BR') ||
+            (item instanceof Text && item.textContent === '\u200B')
```
dmo-odoo, line 15: Euh... non ? C'est pas comme si *tous* les text nodes qui avaient U200B étaient des Linebreak hein !

Gorash: ce n'est pas ça qui est écrit, ceux qui son seul.... a oui, et je rajoute que le précédent doit être un br

Zinston:
Quid de <br>\u200B<fontawesome>\u200B alors?

Gorash:
Parsing OK, rendering OK, but It's not good I think.

TODO later: Choice to be discussed because the <br> in the <li> whether it is for editing or not seems incorrect.
Therefore it is necessary to be able to add <p> in the li when there is at least one <br>.
Moreover, if there is a <p> during parsing, it must be kept.
(This is a bug regardless of the publisher and even without an editor)

PR #472, opened 2020-11-13
[FIX] table: table-picker rendered under some buttons

EmilienD:
When inactive, the "unlink" button in web_editor has an opacity below one.
This creates a new stacking context, which resets the z-index.
See [this link](https://coder-coder.com/z-index-isnt-working/#3-youre-setting-opacity-transform-or-other-css-properties-that-put-the-element-in-a-new-stacking-context) for more info.
It made the "unlink" button visible even when the table picker was over it.
It was fixed by giving the table picker a higher z-index, default being zero.

### DIFF
--- a/packages/plugin-table/assets/tableUI.css
+++ b/packages/plugin-table/assets/tableUI.css
@@ -1,5 +1,6 @@
 table.table-picker {
     position: absolute;
+    z-index: 1;
     background-color: white;
     border: 1px solid black;
     -webkit-box-shadow: 2px 2px 5px 0px rgba(0,0,0,.4);


packages/plugin-table/assets/tableUI.css
```diff
@@ -1,5 +1,6 @@
 table.table-picker {
     position: absolute;
+    z-index: 1;
```
Zinston, line 3: I'm surprised that works. Did you make sure it does?

EmilienD: What do you expect this to do? If you don't understand why it's supposed to work, you can read the PR explanation and the link that goes with it. I tested in Chrome and Firefox on my machine, if you think I should try somewhere else, feel free to tell me.

Goaman:
Fine by me.

PR #473, opened 2020-11-16
[FIX] Color: do not color a ZoneNode

Zinston:


dmo-odoo:
`ZoneNode` should be an intangible container to avoid cases like this. cc @Gorash However, since the intangible container feature is still in development, I'm merging this as a "quickfix".

PR #481, opened 2020-11-23
[FIX] Make sure falsy attribute values are note transform into null

sebgeelen:
An empty string use to be transformed into a null attribute value in the Dom Reconciliation engine

+ adapt a test to cover this case


sebgeelen:
Done with the help of master @Goaman.

PR #487, opened 2020-12-01
[IMP] Odoo: add translation initial value map

EmilienD:
Allows checking whether a translation has been modified or not, to avoid saving it for nothing for example.


EmilienD:
No longer needed, logic implemented in `odoo:addons/web_editor/static/src/js/wysiwyg/wysiwyg.js` instead.

PR #490, opened 2020-12-04
[FIX] Color, TextColor: fix background color behavior

EmilienD:
It turns out that the color behavior was not a standard behavior but specific to TextColor (at least in the context of Odoo).
The simpler behavior is now implemented in `Color` and is used by `BackgroundColor` while the specific behavior is implemented in `TextColor` directly.

### DIFF
--- a/packages/plugin-backgroundcolor/test/BackgroundColor.test.ts
+++ b/packages/plugin-backgroundcolor/test/BackgroundColor.test.ts
@@ -62,41 +62,43 @@ describePlugin(BackgroundColor, testEditor => {
                         '<p style="background-color: yellow;">a[<span style="background-color: red;">bc]</span>d</p>',
                 });
             });
-            it('should set the background color of a paragraph to red', async () => {
+            it('should set the background color of the selected text only to red', async () => {
                 await testEditor(BasicEditor, {
                     contentBefore: '<p>[abc]</p><p>def</p>',
                     stepFunction: async editor => {
                         await colorBackground(editor, 'red');
                     },
-                    contentAfter: '<p style="background-color: red;">[abc]</p><p>def</p>',
+                    contentAfter:
+                        '<p><span style="background-color: red;">[abc]</span></p><p>def</p>',
                 });
             });
-            it('should set the background color of everything to red', async () => {
+            it('should not set the background color of the parent to red', async () => {
                 await testEditor(BasicEditor, {
                     contentBefore: '<p>[abc]</p>',
                     stepFunction: async editor => {
                         await colorBackground(editor, 'red');
                     },
-                    contentAfter: '<p style="background-color: red;">[abc]</p>',
+                    contentAfter: '<p><span style="background-color: red;">[abc]</span></p>',
                 });
             });
-            it('should not set the background color of characters that already have that background color through an ancestor', async () => {
+            it('should set the background color of characters that already have that background color through an ancestor', async () => {
                 await testEditor(BasicEditor, {
                     contentBefore: '<p style="background-color: yellow;">a[bc]d</p>',
                     stepFunction: async editor => {
                         await colorBackground(editor, 'yellow');
                     },
-                    contentAfter: '<p style="background-color: yellow;">a[bc]d</p>',
+                    contentAfter:
+                        '<p style="background-color: yellow;">a[<span style="background-color: yellow;">bc]</span>d</p>',
                 });
             });
-            it("should only set the background color of characters that don't already have that background color through an ancestor", async () => {
+            it("should set the background color of both characters that have and don't already have that background color through an ancestor", async () => {
                 await testEditor(BasicEditor, {
                     contentBefore: '<p style="background-color: yellow;">a[bc</p><p>de]f</p>',
                     stepFunction: async editor => {
                         await colorBackground(editor, 'yellow');
                     },
                     contentAfter:
-                        '<p style="background-color: yellow;">a[bc</p><p><span style="background-color: yellow;">de]</span>f</p>',
+                        '<p style="background-color: yellow;">a[<span style="background-color: yellow;">bc</span></p><p><span style="background-color: yellow;">de]</span>f</p>',
                 });
             });
             it('should not set the background color of characters that already have that background color through a format', async () => {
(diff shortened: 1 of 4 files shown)


packages/plugin-color/src/Color.ts
```diff
@@ -169,69 +129,7 @@ export class Color<T extends ColorConfig = ColorConfig> extends JWPlugin<T> {
...
-            return colorAncestor;
-        }
-        for (const format of node.modifiers.filter(Format)) {
-            if (this.hasColor(color, format)) {
-                return format;
-            }
-        }
-    }
-    private _applyColor(node: VNode | Format, color: string): void {
+    protected _applyColor(node: VNode | Format, color: string): void {
```
dmo-odoo, line 132: Tu l'as mises `protected` mais elle n'est pas overridée au final, je me trompe ?

packages/plugin-backgroundcolor/test/BackgroundColor.test.ts
```diff
@@ -62,25 +62,25 @@ describePlugin(BackgroundColor, testEditor => {
...
             });
-            it('should not set the background color of characters that already have that background color through an ancestor', async () => {
+            // Following tests do not match 14.0 behavior, which spec is more
+            // correct probably needs to be discussed.
+            // I'll also mention that if we don't want to set the background
+            // color when it's already the one of the ancestor, perhaps we should
+            // also expect that setting the background color on the ancestor
+            // should also remove the similar background color on the children.
+            // This all pertains to cascading behavior support, which isn't
+            // implemented in a standard way yet.
```
Zinston, line 82: So if I got this right, you mean that in the context of:
```html
<p><span style="background-color: red;">ab</span></p>
```
If I want to set the background color of the paragraph, it should remove the style on the span (or the span altogether).
I guess coloring the ancestor can be done indeed, in the case of a Website snippet.

Now I wonder if this is true in any case. Say I background-color some text green, then I background-color the containing snippet green, then later I background-color the snippet red instead. Thing is, if it just happened I guess I'd expect the text to remain green, but chances are these operations are spaced with weeks or even months... Then I forgot that the text was actually colored individually, so it takes me as a surprise.
So I would tend to agree with you. Do you know what GSpec does? Probably testable with a table cell color.

packages/plugin-backgroundcolor/test/BackgroundColor.test.ts
```diff
@@ -166,53 +166,25 @@ describePlugin(BackgroundColor, testEditor => {
...
-            it('should set the background color of two characters to white, within a paragraph with yellow background', async () => {
-                await testEditor(BasicEditor, {
-                    contentBefore: '<p style="background-color: yellow;">a[bc]d</p>',
-                    stepFunction: async editor => {
-                        await uncolorBackground(editor);
-                    },
-                    contentAfter:
-                        '<p style="background-color: yellow;">a[<span style="background-color: white;">bc]</span>d</p>',
-                });
-            });
```
Zinston, line 178: Why did you choose to remove this test instead of skipping it and adding a comment, like for the others?
If I understand this right, this behavior doesn't exist anymore because we now only treat the background-color change at the character level, so that uncolorBackground would rightly do nothing because the _characters_ have no background color. Is that correct?

packages/plugin-backgroundcolor/test/BackgroundColor.test.ts
```diff
@@ -166,53 +166,25 @@ describePlugin(BackgroundColor, testEditor => {
...
-                    contentAfter: '<p>[abc]</p><p>def</p>',
-                });
-            });
-            it('should unset the background color of everything', async () => {
-                await testEditor(BasicEditor, {
-                    contentBefore: '<p style="background-color: red;">[abc]</p>',
-                    stepFunction: async editor => {
-                        await uncolorBackground(editor);
-                    },
-                    contentAfter: '<p>[abc]</p>',
```
Zinston, line 205: Same question.

Zinston (commented):
@dmo-odoo asked me to review the tests. It seems alright to me but I did have a couple of non-crucial questions (mostly relating to skipping vs removing and unclear comment, then just some confirmation that the behavior is what I think it is ^^).

packages/plugin-color/src/Color.ts
EmilienD: Elle est pas overridée mais elle est utilisée dans Color et dans TextColor.

packages/plugin-backgroundcolor/test/BackgroundColor.test.ts
EmilienD: I think I removed them because the behaviour was not relevant anymore, as opposed to a choice in spec that could be made in the future.
On top of that I also felt like uncolouring should remove the colour that was set on the element, as opposed to colouring it white. As in, if I have a red background on the paragraph, I highlight some text in green, then uncolour that text, it's pretty surprising to find it highlighted in white instead.

Your assessment seems correct.

packages/plugin-backgroundcolor/test/BackgroundColor.test.ts
```diff
@@ -62,25 +62,25 @@ describePlugin(BackgroundColor, testEditor => {
...
-                        await colorBackground(editor, 'red');
-                    },
-                    contentAfter: '<p style="background-color: red;">[abc]</p>',
+                    contentAfter:
+                        '<p><span style="background-color: red;">[abc]</span></p><p>def</p>',
                 });
             });
-            it('should not set the background color of characters that already have that background color through an ancestor', async () => {
+            // Following tests do not match 14.0 behavior, which spec is more
+            // correct probably needs to be discussed.
```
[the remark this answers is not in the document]
Zinston: @dmo-odoo OK for you? Seems alright to me. Though I would still replace this comment with a note explaining why we'd consider that less specific behavior acceptable.

packages/plugin-backgroundcolor/test/BackgroundColor.test.ts
Zinston: > Do you mean "is it an actual use case"? If so, it is a possible scenario in Website snippets indeed.
Yes that is indeed what I meant to say. Have you tested how it behaved before your changes and compared with how it behaves now? I mean this particular use case with Website snippets.

> What's GSpec?
"When in doubt, do as Google Docs does". We're in doubt here so let's do as Google Docs does. Does it do the same as you? You can probably test that by coloring a cell of a table.

packages/plugin-backgroundcolor/test/BackgroundColor.test.ts
Zinston: :+1:

packages/plugin-backgroundcolor/test/BackgroundColor.test.ts
EmilienD: GSpec preserves highlighting, GSpec also lets you highlight text where the container background is already of the highlight colour.
This is also the behaviour that is happening with these changes.

